PR #22, opened 2023-08-11
Layered edits to the home page.

akcano:


docs/index.rst
```diff
@@ -3,67 +3,58 @@
...
-environments, perform code reviews or experiment in a separate light-weight
-container.
-
-It is common to have a non-trivial project setup with dependencies on particular
-Linux distributions, SDKs from multiple publishers, and system and language
-packages. Most such projects can organise setup complexity with Workspace.
-Examples include AI/ML, Robotics, IoT, EdTech and similar domains.
+=========
+
+**Workspace automates configuration and management
```
dmitry-lyfar, line 8: The intention there is to say that Workspace is a tool that organises (or orchestrates) reproducible dev environments. The suggested edit reads (and I could be getting it wrong): Workspace automates configuration and management of _something_ **with** dev environments, i.e.  the meaning is different?

docs/index.rst
```diff
@@ -3,67 +3,58 @@
...
+with reproducible development environments**.
+
+**Define your dev environment in straightforward YAML**.
+The tool consumes it to create a contained workspace,
+installs the SDKs and packages it lists,
```
dmitry-lyfar, line 13: Not the original nor the suggested copy seems to offer a good enough explanation showing that Workspace differentiate between SDKs and packages. SDKs and packages are two separate entities, SDK may or may not install packages. Given that we can only install SDKs for now, I suggest we drop _packages it lists_?

Also, Workspace does not _add_ hooks technically, these are supplied by the SDK authors; Workspace calls them at certain events (e.g. launch, refresh) to let the SDK execute the required logic. Hence, _control_ in the original home page. Any thoughts on a better way of describing this?

docs/index.rst
```diff
@@ -3,67 +3,58 @@
...
+and adds life cycle hooks for run-time control.
+IDEs such as Visual Studio Code or Jupyter Lab
+can discover workspaces and leverage them in their operation;
+when you're done with a workspace,
+its disposal doesn't affect the host system.
```
dmitry-lyfar, line 18: The, perhaps not obvious, message there was not only about affecting the host system, but also about a person using the tool becoming better organised and/or efficient (hence, the choice of _clean_). 

Now: the host is unaffected. ok, so what does it give to me?
Wanted:  the host is unaffected, so no unused packages, conflicting dependencies for different projects that deteriorate the host system with time.

dmitry-lyfar (changes requested):
Thanks @akcano! Only a few comments from me there, let's discuss on a call / here.

PR #28, opened 2023-08-17
Minor README rework

akcano:


docs/README.rst
```diff
@@ -1,98 +1,126 @@
...
 
-Development
------------
 
-Workspace uses a "go test"-compatible `gocheck <https://pkg.go.dev/gopkg.in/check.v1#section-readme>`_:
+Workspace downloads and installs the SDKs your definition lists;
+the project is now ready to use them.
+
+
+Testing
```
dmitry-lyfar, line 108: I used to call this section Testing. Switched to Development so the reader skimming through it could see a familiar OSS project structure: Install -> Quick Start -> Link to a wider documentation -> Information for Developers and Contributors. The latter is planned to include Workspace development documentation (which, for now, is just how to run tests, i.e. rather a Development -> Testing structure). 

I will merge this one now and put this question on our next agenda to discuss as this section will likely be pretty minimal in the foreseeable future.

dmitry-lyfar (approved):

docs/README.rst
akcano: I see; however, it's always best to choose titles most appropriate for what's actually there.
Another idea I had re: this section is merging it into 'Installation', which kinda makes sense (you install Workspace from source, then you run tests to make sure it's not broken). A sub-idea here is to make the tests a part of `go install` (unless they're already there).

PR #29, opened 2023-08-18
Create snap for Workspace

dmitry-lyfar:


snap/snapcraft.yaml
```diff
@@ -0,0 +1,28 @@
+name: workspace
+title: Workspace
+base: core20
+version: 0.1.0
+summary: Workspace
```
dmitry-lyfar: @akcano The summary field has a limitation of 78 symbols that does not fit our headline. Hence, the headline went to the description field. WDYT should be here?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,28 @@
+name: workspace
+title: Workspace
+base: core20
```
jnsgruk:
```suggestion
base: core22
```

snap/snapcraft.yaml
```diff
@@ -0,0 +1,28 @@
...
+  workspaced:
+    environment:
+      WORKSPACE: ${SNAP_COMMON}
+    command: bin/workspaced run --create-dirs
+    daemon: simple
+parts:
+  workspace:
+    plugin: go
+    source: .
+    source-type: local
```
jnsgruk, line 37:
```suggestion
    source-type: local
    build-snaps:
        - go/latest/stable
```

jnsgruk: This will be needed if you switch to `core22`

snap/snapcraft.yaml
akcano: @dmitry-lyfar 78 characters it is, then:
`Workspace automates intricate prerequisite setup for your development projects`

Also, take a look [here](https://docs.google.com/document/d/1NHFShs9zBsymdnLzsOEDpQ1O5H3b7gDYS7OVqkVgMpQ/edit) for an elaborated version.

docs/README.rst
```diff
@@ -50,32 +50,15 @@ for low-level operation:
    sudo snap install lxd
    sudo lxd init --auto
 
-
-Next, install Workspace from source code:
+Build and install the workspace snap package:
```
akcano: A bit of nitpicking; I suggest this:
`Build and install the ``workspace`` snap:`

Here, 'snap package' is a pleonasm; it deteriorates the messaging.
Compare to [this](https://snapcraft.io/docs/getting-started):

```
A snap is a bundle of one or more applications (“apps”) and their dependencies that works without modification across many different Linux distributions. Snaps are discoverable and installable from the [Snap Store](https://snapcraft.io/store), a public app store with an audience of millions.
```

No mention of packages.

docs/README.rst
```diff
@@ -50,32 +50,15 @@ for low-level operation:
...
-.. code-block:: bash
-
-   mkdir ~/workspace
-   export WORKSPACE=~/workspace
-   workspaced run
-
-     2023-08-17T01:37:23.962Z [workspaced] Started daemon.
-     ...
-
+   git clone git@github.com:canonical/workspace.git
```
akcano, line 57: I suggest adding the HTTPS option. Not everyone have their SSH keys at the ready all the time:
```
git clone git@github.com:canonical/workspace.git
# -- or --
git clone https://github.com/canonical/workspace.git
```

akcano (changes requested):
The snap worked like a charm on the host machine, thanks!
However, there were issues with actual workspace creation after installation, will report them privately.

On the VM, the snap failed; some `multipass`-related problems need investigation; will report them privately as well.

snap/snapcraft.yaml
dmitry-lyfar: @jnsgruk thanks, added. With core22 I got [this](https://github.com/canonical/craft-providers/pull/182); although, looks like it was fixed. `snapcraft clean` solves the issue and, hopefully, it was a one-off problem due to snapcraft not disabling auto-refresh in my environment that was used to build core20 snaps.

dmitry-lyfar: @jnsgruk sorry for a bunch of possible GitHub updates, I could not get it applied as a single batch on GitHub for some reason it fails to combine core22 suggestions. Also, snapcraft complains about plugs being in the global space for core22, changed that as well.

docs/README.rst
dmitry-lyfar: Updated, thanks!

akcano (approved):
Also, let's maybe monitor where we use these descriptions:
https://docs.google.com/spreadsheets/d/1KYo0NXiZHWB8YU51vKq4FGIBq9bne_tAW--azEw4c88

jnsgruk (approved):

docs/README.rst
```diff
@@ -50,32 +50,18 @@ for low-level operation:
...
-   go install github.com/canonical/workspace/cmd/workspaced@latest
-   go install github.com/canonical/workspace/cmd/workspace@latest
-
-
-Running Workspace
------------------
-
-To run the daemon,
-create a directory to store Workspace state and data,
-save its pathname in the ``$WORKSPACE`` environment variable,
```
akcano, line 67: Seems we're losing this info with the update, but the snap installation doesn't work without the variable; can you please confirm the intended behaviour, @dmitry-lyfar ?

akcano (changes requested):

docs/README.rst
dmitry-lyfar: @akcano I suspect you still have ``workspace`` installed from sources which takes precedence over the one installed from snap. Try ``which workspace``; it is likely the binary from ``~/go/bin``, whilst it should be ``/snap/bin/workspace``. The ``WORKSPACE`` variable is set and used by snap, so no need to set it up in a normal scenario when you install it from snap.

akcano: @dmitry-lyfar confirmed, that's the case.

akcano (approved):

PR #32, opened 2023-08-21
Rewriting tutorials + ongoing maintenance

akcano:


docs/tutorial/getting-started/install/install.rst
```diff
@@ -1,91 +1,100 @@
...
 
 
-Go
----
-Currently you can install Workspace only from its source code, which requires Go run-time:
+Next, ensure the
+`LXD daemon
+<https://documentation.ubuntu.com/lxd/en/latest/explanation/lxd_lxc/#lxd-daemon>`_
+is enabled and running.
+With a snap-based installation, use ``systemctl``:
```
dmitry-lyfar: With a snap-based installation, I'd recommend this:

```
snap services lxd.daemon # this should be enough and the daemon must be running at this stage
```
Using ``systemctl enable`` for the daemon that is running within a snap led to some quirky bugs before.

If the daemon is stopped for any reason, one can try:
```
snap start lxd.daemon
```

dmitry-lyfar (changes requested):
Looks great! Left one comment regarding the LXD daemon.

docs/tutorial/getting-started/install/install.rst
akcano: @dmitry-lyfar Thanks, good point.
Updated; I reworked the command sequence a bit to make sure it reflects the 'make sure it's enabled and running' phrasing.

dmitry-lyfar (approved):

PR #35, opened 2023-08-29
Minimum viable rewrite of the existing docs

akcano:


docs/tutorials/essentials/refresh.rst
```diff
@@ -0,0 +1,126 @@
...
+
+       "nimble" refreshed
+
+The workspace is rebuilt from the
+:ref:`base <exp_workspace_base>`;
+then the SDKs are updated from their respective channels.
+If your
+:ref:`project <exp_project>`
+has multiple workspaces,
+``workspace refresh`` simultaneously updates all of them.
```
dmitry-lyfar: Currently, it is implemented so that you have to name every workspace we want to update explicitly, ``workspace refresh one two three``. It's worth considering to have a version that updates everything it finds if no arguments provided.

dmitry-lyfar (commented):
Thanks! One note on the refresh without args only.

docs/tutorials/essentials/refresh.rst
akcano: Updated.

dmitry-lyfar (approved):

PR #38, opened 2023-09-06
Docs: Add 'workspace refresh' CLI reference

akcano:


cmd/workspace/refresh.go
```diff
@@ -17,15 +17,46 @@ type CmdRefresh struct {
...
 func (c *CmdRefresh) Command() *cobra.Command {
 	var cmd = &cobra.Command{
-		Use:   "refresh <workspace>...",
+		Use:   "refresh [--abort|--continue|--wait-on-error] <WORKSPACE>...",
 		Args:  cobra.MinimumNArgs(1),
-		Short: "Refresh one or many workspaces",
+		Short: "Update workspaces according to their updated definitions.",
+        Long: `
+This command updates the workspaces listed as arguments according to the
+changes in their definitions.  For each workspace, it:
```
dmitry-lyfar:
```suggestion
changes in their definitions. For each workspace, it:
```

cmd/workspace/refresh.go
```diff
@@ -17,15 +17,46 @@ type CmdRefresh struct {
 
 func (c *CmdRefresh) Command() *cobra.Command {
 	var cmd = &cobra.Command{
-		Use:   "refresh <workspace>...",
+		Use:   "refresh [--abort|--continue|--wait-on-error] <WORKSPACE>...",
 		Args:  cobra.MinimumNArgs(1),
-		Short: "Refresh one or many workspaces",
+		Short: "Update workspaces according to their updated definitions.",
```
dmitry-lyfar: This is true, however, refresh will update a workspace even a definition hasn't changed. For example, if a new SDK available at a given channel (tbd).

cmd/workspace/refresh.go
```diff
@@ -34,24 +65,24 @@ func (c *CmdRefresh) Run(cmd *cobra.Command, av []string) error {
 	var err error
 
 	if c.Abort && c.Continue {
-		return fmt.Errorf("cannot refresh: flags --continue and --abort are incompatible")
+		return fmt.Errorf("can't refresh: '--abort' can't be used with '--continue'")
```
dmitry-lyfar: TBD: "cannot something" is used throughout workspace and some related Canonical products (e.g. pebble, spread).

cmd/workspace/refresh.go
akcano: Updated. I prefer double spaces between sentences in doc markup, but let's stick to single spaces for in-code text such as help strings.

cmd/workspace/refresh.go
akcano: Updated. I changed the wording in these strings slightly to avoid repeating a larger 'cannot' over and over. Also, 'incompatible' works here but should be probably used with caution when (potentially) applied to entities of different types (e.g. a flag incompatible with a file).

cmd/workspace/refresh.go
akcano: Updated here and in the intro sentence (dropped 'according to the changes in their definitions'). I simply dropped 'updated' to avoid overburdening here and used plain words in the second occurrence, but that's only one way to approach this.

dmitry-lyfar (approved):

PR #42, opened 2023-09-20
Add 'workspace info' and 'workspace launch' to CLI reference

akcano:


cmd/workspace/changes.go
```diff
@@ -30,7 +30,7 @@ func (c *CmdChanges) Run(cmd *cobra.Command, av []string) error {
 
 	cli, err := client.New(&ClientConfig)
 	if err != nil {
-		return fmt.Errorf("cannot create client: %v", err)
+		return fmt.Errorf("cannot create the client: %v", err)
```
dmitry-lyfar: The rule across Canonical products (see Pebble, snapd) that is the error messages are intentionally short, lowercase and no articles. Contrary, log messages must be in correct English. I'm a bit confused on the former as the error messages do make their way to the user too.

cmd/workspace/info.go
```diff
@@ -14,9 +14,21 @@ type CmdInfo struct {
...
+		Long: `
+This command outputs the basic settings, current status and individual SDK
+details for a workspace, formatting them as YAML. Specifically, it prints:
+
+- Essential workspace attributes, such as name, base and project directory
+- Current status (e.g. *Ready*, *Pending*, *Off*) and notes for the workspace
+- Individual SDK details, such as name, channel, installation date and revision
+
+Notes:
+- Avoid assumptions based on SDK channels: 'latest/stable' may be neither
```
dmitry-lyfar, line 29: Not sure I understand the note?

cmd/workspace/launch.go
```diff
@@ -14,9 +14,27 @@ type CmdLaunch struct {
 
 func (c *CmdLaunch) Command() *cobra.Command {
 	var cmd = &cobra.Command{
-		Use:   "launch <workspace>...",
+		Use:   "launch <WORKSPACE>...",
 		Args:  cobra.MinimumNArgs(1),
-		Short: "Launch one or many workspaces",
+		Short: "Initialise one or many workspaces using their definitions.",
```
dmitry-lyfar, line 19: the definitions should be in the current directory, it looks like we are not explicit about it?

cmd/workspace/launch.go
```diff
@@ -14,9 +14,27 @@ type CmdLaunch struct {
...
+		Long: `
+This command initialises the workspaces listed as arguments by going over their
```
dmitry-lyfar: To discuss: initialise vs. launch

cmd/workspace/refresh.go
```diff
@@ -39,6 +39,7 @@ the operation is aborted and reverted for *all* of them.
 
 Notes:
 - The workspace must be *Ready* to be refreshed
+- To initialise a newly defined workspace, use 'workspace launch' instead
```
dmitry-lyfar: Same point about initialise vs. launch. If we initialise, we should call the command init (?).

cmd/workspace/launch.go
dmitry-lyfar: decided: initialise -> construct

dmitry-lyfar (approved):

PR #52, opened 2023-10-25
Docs: Add workshop exec, remove, changes, tasks subcommand reference

akcano:


cmd/workshop/remove.go
```diff
@@ -17,6 +17,22 @@ func (c *CmdRemove) Command() *cobra.Command {
 		Use:   "remove <WORKSHOP>...",
 		Args:  cobra.MinimumNArgs(1),
 		Short: "Remove one or many workshops",
+		Long: `
+This command removes the workshops listed as arguments. For each workshop, it:
+
+- Checks that the workshop isn't *Off* or *Pending*
+- Stops the workshop if it's not already *Stopped*
+- Deletes the workshop's files
```
dmitry-lyfar: The remove will not delete the workshop files, it would only remove the workshop which will result it to be shown as 'Off' in the list output (i.e. only existing in a form of a file).

cmd/workshop/remove.go
```diff
@@ -17,6 +17,22 @@ func (c *CmdRemove) Command() *cobra.Command {
...
+
+If multiple workshops are listed and an error occurs,
```
dmitry-lyfar: Here, the operation is not transactional. If a workshop was removed and some other remove failed, the former is gone forever.

cmd/workshop/remove.go
```diff
@@ -17,6 +17,22 @@ func (c *CmdRemove) Command() *cobra.Command {
...
+the operation is aborted for *all* of them.
+
+Notes:
+- The workshop must be *Stopped*, *Ready* or in *Error* to be removed
```
dmitry-lyfar: I would rewrite it in a reverse manner to highlight that if there is a pending operation, remove will be a no-op.

docs/reference/workshop-exec.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Synopsis
+--------
+
+The :program:`exec` subcommand runs an arbitrary command
+in the specified workshop, waiting for it to complete.
+If a timeout elapses before that, it's terminated.
+
+To run an :program:`exec` command, the workshop must be *Ready* or *Pending*,
+and the workshop base must support the command.
```
dmitry-lyfar: "...base must support the command", I think it can be replaced with stating that the command must exists. It's not only base, but an SDK may provide a command and, thus, it will be available even if not supported by the base.

dmitry-lyfar (commented):
Looks pretty clear. I left a few notes, we can discuss/merge in person.

cmd/workshop/remove.go
akcano: This probably needs a live discussion of what's deleted and what constitutes workshop's 'files'.

cmd/workshop/remove.go
akcano: OK.

cmd/workshop/remove.go
akcano: Again, does this also affect the command behaviour with multiple workshops, as suggested previously (i.e. it's non-transactional)?

docs/reference/workshop-exec.rst
akcano: Fair enough, but I think that we can boil 'base + SDKs' down to 'workshop' here. However, I'd like to bring up another issue: perhaps we can call the `exec`-run entity a 'binary' or an 'executable' instead of a simple 'command'?

cmd/workshop/remove.go
dmitry-lyfar: as per the conversation: add clarification on the consequences for the already deleted workshops.

cmd/workshop/remove.go
dmitry-lyfar: all listed workshops must satisfy the required status.

docs/reference/workshop-exec.rst
dmitry-lyfar: no need to clarify

cmd/workshop/exec.go
```diff
@@ -41,42 +41,58 @@ type CmdExec struct {
...
+- Interactively (for shell sessions)
+- Non-interactively (for scripts)
+
+To set the mode explicitly, use '-i' or '-I'. If neither is supplied,
+'exec' deduces the mode based on the nature of its own streams:
+
+- If stdin and stdout are terminals, the mode is interactive
+- Otherwise, it's non-interactive
+
+To separate the 'exec' subcommand options from the command itself,
```
dmitry-lyfar:
```suggestion
To separate the 'exec' subcommand from the command itself,
```

docs/reference/workshop-exec.rst
```diff
@@ -0,0 +1,108 @@
...
+- Non-interactively (for scripts)
+
+To set the mode explicitly, use :option:`!-i` or :option:`!-I`.
+If neither is supplied,
+:program:`exec` deduces the mode based on the nature of its own streams:
+
+- If :samp:`stdin` and :samp:`stdout` are terminals, the mode is interactive
+- Otherwise, it's non-interactive
+
+To separate the :program:`exec` subcommand options from the command itself,
```
dmitry-lyfar:
```suggestion
To separate the :program:`exec` subcommand from the command itself,
```

dmitry-lyfar (approved):

PR #72, opened 2023-12-06
Docs: Improve command sample logic, rework home page with UX findings

akcano:


docs/explanation/essentials/moving-projects.rst
```diff
@@ -105,6 +117,7 @@ before deleting the project directory:
 
 .. code:: console
 
+   $ cd /home/user/project/
```
dmitry-lyfar: One can do `workshop --project=<dir>` remove golang. I've seen cding into the dir in a few examples in our docs and was curious whether we should use the project option instead. Esp. in this case.

docs/tutorial/index.rst
```diff
@@ -417,9 +406,7 @@ You can run an interactive shell as well:
 .. code:: console
 
    $ workshop exec golang bash
-   $ uname -a
-
-       Linux golang-bf3a1040 6.2.0-35-generic #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct  6 10:23:26 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
+   workshop@golang-cd03e2cd:/project$ uname -a
```
dmitry-lyfar, line 409: hmm, I should definitely work on the default promt which is the scary instance name now. But that's a side note, I'll put it on todo.

dmitry-lyfar (approved):
I think this one is good, won't merge as I wanted your comment on the 'cd' bit.

docs/explanation/essentials/moving-projects.rst
akcano: In brief, this is a legitimate consideration. so I'll proceed to implement it; also, at this point, the explanation also needs a testing script.

docs/tutorial/index.rst
akcano: Actually, it's not that bad (all parts make sense and add information, but we simply omit it at the moment), but making it carry some added value won't hurt.

dmitry-lyfar (approved):

PR #76, opened 2023-12-12
Add spread test suite for workshop documentation

dmitry-lyfar:


tests/documentation/tutorial/task.yaml
```diff
@@ -0,0 +1,16 @@
+summary: Run tutorial commands to ensure it's operational
+prepare: |
+  chmod u+x "$TESTSLIB"/documentation/tutorial.sh
```
akcano, line 3:
```suggestion
  chmod u+x "$TESTSLIB"/documentation/tutorial.sh
  chown ubuntu:ubuntu "$TESTSLIB"/documentation/tutorial.sh
```

akcano: Otherwise the script is unavailable to the runner.

tests/lib/utils.sh
```diff
@@ -2,6 +2,31 @@
...
+  snap install lxd --classic
+  cd /remote; GOBIN=/usr/bin go install -buildvcs=false /remote/cmd/workshop/
+  lxd init --auto
+
+  # run fake GCS bucket storage to emulate SDK store
+  publish_test_sdk_content "$SDKCONTENT" "$SDK_STORE_BUCKET_DIR"
+  /bin/sh -c "nohup go run github.com/fsouza/fake-gcs-server@latest -data /data -scheme http -port 8080 -public-host localhost:8080 > ~/fake_sdk_store.log 2>&1 &"
+
+
+  echo "Waiting fake SDK store to launch on 8080..."
```
akcano, line 18:
```suggestion
  echo "Waiting for the fake SDK store to start on port 8080..."
```

akcano (changes requested):
AWESOME but needs minor tweaks.

akcano (approved):

tests/lib/documentation/tutorial.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+  
+set -eux
+whoami
+
+# Check prerequisites (snapcraft doesn't appear here deliberately)
+
+sudo snap install lxd
```
tigarmo, line 8: hi! Sorry for the drive-by comment on a merged PR, but I saw this update on Dmitry's pulse report and thought I'd chip in with how we do this in Rockcraft. The idea is basically the same but the documentation actually sources from the tests, so there's no way for them to go out-of-sync.

For example: [this spread test](https://github.com/canonical/rockcraft/blob/main/docs/tutorials/code/hello-world/task.yaml#L12) is sourced piece-wise by [this tutorial](https://github.com/canonical/rockcraft/blob/main/docs/tutorials/hello-world.rst), using `start-after` and `end-before`. 

Feel free to hit me/us up if you want to talk about this. Happy holidays ;)

akcano: No worries, and thanks for raising this; a good chance to shed some light on what's going on.

TL;DR: making baby steps here, first needed to make sure we can interoperate with Spread at all; now we can expand to other sections and gradually introduce include mechanics as some parts eventually coagulate

Another concern for doing what you do straight away is the author's convenience.  Using `:start-after:` and `:end-before:` helps in getting single-sourcing work without off-by-one issues, but an important drawback here is loss of context when you work on the documentation itself.  You will need to switch between the script and the doc constantly to make sure you have your names right etc.  When the doc is mostly static and updates rarely, it's OK, but we're still in the rapid development phase, so edits occur often.

tigarmo: @akcano that's all valid and at the end of the day the docs are yours of course ;) . Just thought I'd point out the alternative in case you were unaware. I also agree that there's an extra cognitive burden when initially writing the docs, specially if they are expected to evolve rapidly

dmitry-lyfar: @tigarmo thanks for taking the time to comment (somebody reads our pulse reports!). This is an elegant solution.   I spent some time exploring the documentation tests of rockcraft and cannot fully understand how do you make it to run under a non-root user? I see something like this:

``` 
  # [docs:install-rockcraft]
  sudo snap install rockcraft --classic
  # [docs:install-rockcraft-end]

  # [docs:build-rock]
  rockcraft pack
  # [docs:build-rock-end]

  test -f hello_1.0_amd64.rock
```

Given the sudo in one of the commands, I presume the entire test is ran as a user? But spread.yaml uses the default root backend user. 

This is the problem I came across for our e2e and doc tests too as most of them must run as a user. This is solved by wrapping the command or simply putting the tutorial in a shell script and running as a desired user.

How do you solve this?

PR #77, opened 2023-12-12
Docs: Update boilerplate in README, add a11y GitHub workflow

akcano:


docs/README.rst
```diff
@@ -1,30 +1,31 @@
...
-**Workshop is a tool that automates intricate prerequisite setup
-for your projects**.
-
-**Define your dev environment in straightforward YAML**.
-The tool consumes the definition to create a contained workshop,
-installs the dependencies it lists as a number of SDKs,
-and attaches their life cycle hooks for run-time control.
-IDEs such as Visual Studio Code or JupyterLab
-can discover workshops and use them in their operation,
+**Workshop is a tool that fractions development environment prerequisites
```
dmitry-lyfar, line 4: This is interesting. I would test the headline against a developer. I think it is correct from the correctness point of view but a bit too complex to understand without a context. 

This headline should rather be focused around a specific problem the tool solves that can trigger an aha moment from the first sentence, e.g. setting a reproducible development environment or automating a development environment setup (which we had in the previous version and the dev env, as I see now, should not necessarily be 'intricate' -- it would work for a Go or Flutter env. It is fully-automated, easy to hack with and discard when done). We should talk more about this on a call.

dmitry-lyfar (approved):

PR #79, opened 2023-12-15
Content interface explanation, revamped README boilerplate

akcano:


docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Content interface slot
+----------------------
+
+To let SDKs access the host file system,
+Workshop creates a slot per each content interface plug.
+
+.. note::
+
+   Currently, content can only be exposed by Workshop itself,
```
dmitry-lyfar: A workshop cannot expose the content in our plug and slot terms (meaning exposing is declaring a slot), it is only the host system that can expose content to a workshop. Or Workshop (cap. A) means workshopd, the product?

docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,107 @@
...
+   but can't be shared between two workshops directly.
+
+
+At run-time, the plug will be connected to the slot;
+after some metadata matching that Workshop performs internally,
```
dmitry-lyfar: The "some metadata matching" in this process is a validation and policy check (i.e. the plug declaration is correct + plug is permitted to be installed and auto-connected + the target directory must exist). I guess the target directory existence is an important requirement to disclose.

docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,107 @@
...
+
+
+This setup yields the following pathname:
+
+.. code-block:: none
+
+   ~/.local/share/workshop/project/<INTERNAL PROJECT ID>/content/golang_go_mod-cache.sdk
+
+That's where the module cache from the workshop will end up;
+it's quite obvious that the path is not intended for direct manipulation.
```
dmitry-lyfar: Yes, the source location must be manipulated by the workshop remount command. It does not exist yet (only as a spec), so perhaps, no need to mention it now, just triage the spec on the coverage map and come back here when the command materialises.

docs/explanation/essentials/project-workshop-sdks.rst
```diff
@@ -146,10 +146,97 @@ A single workshop can include multiple SDKs from different publishers.
...
+
+Upon :ref:`refresh <ref_workshop_refresh>`,
+existing connections are preserved in the refreshed workshop
+if their plugs were connected before the operation.
+Stale connections are removed,
+but the content remains.
+
+On :ref:`remove <ref_workshop_remove>`,
+both the interface connections and the host directories
+(if any were created, for example,  to accommodate content slots)
```
dmitry-lyfar:
```suggestion
(if any were created, for example, to accommodate content slots)
```

docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,107 @@
+.. _exp_content_interface:
+
+Content interface
+=================
+
+Introduction
+------------
+
+One of the
```
dmitry-lyfar: I'm not sure how you plan this intro for other interfaces, but maybe it is worth dropping "One of the interfaces" and go straight into what it does given we will have a GPU, ssh, etc interface explanation pages next to this one (and, perhaps, have the interfaces link as a see also)?

docs/explanation/advanced/content-interface.rst
dmitry-lyfar: To enable the other two (a plug can be installed and auto-connected), the interface system uses a policy mechanism. Every interface has an installation and connection policy defined by workshopd. The content interface plug can, according to the slot policy, be installed and auto-connected. Other interfaces may have different policies (e.g. enable installation and prohibit auto-connection for ssh-agent, for example). See https://github.com/canonical/workshop/blob/d2a7365b16d6af8f8d4acab7de4b5773492d1719/internal/interfaces/builtin/content.go#L37

docs/explanation/essentials/project-workshop-sdks.rst
```diff
@@ -146,10 +146,97 @@ A single workshop can include multiple SDKs from different publishers.
...
+To make SDKs customisable and extensible,
+Workshop implements a counterpart to
+:program:`snapd`'s
+`interface manager <https://snapcraft.io/docs/interface-management>`__,
+controlling whether an individual SDK can use resources beyond its confinement.
+You can think of specific interfaces as resource *types*:
+file system, hardware, computational and so on.
+
+In order to provide access to these resource types,
+Workshop exposes so-called *interface slots*.
```
dmitry-lyfar: This is a similar issue about who actually exposes resources. Technically, it is the host system exposes the resource slots (internally named as 'workshop' because we consider the host system as a special kind of a workshop, similarly, how snapd exposes some implicit system slots via 'core' system snap) . It is up to you how to leave this bit as detailing it may make things more confusing. I wanted to point this difference out.

docs/explanation/essentials/project-workshop-sdks.rst
```diff
@@ -146,10 +146,97 @@ A single workshop can include multiple SDKs from different publishers.
...
+Upon :ref:`refresh <ref_workshop_refresh>`,
+existing connections are preserved in the refreshed workshop
+if their plugs were connected before the operation.
+Stale connections are removed,
+but the content remains.
+
+On :ref:`remove <ref_workshop_remove>`,
+both the interface connections and the host directories
+(if any were created, for example,  to accommodate content slots)
+are removed for simplicity.
```
dmitry-lyfar: Just a note for a better context: The reason we remove the content in the default locations is a lesson learnt from snapd: we should not store the users data infinitely and some operations will remove them to avoid accumulating data in the invisible locations that is unlikely to be reused. Shall a user change the source via workshop remount, we will not touch the new directory as the explicitly moved somewhere we have no knowledge of.

docs/README.rst
```diff
@@ -1,25 +1,25 @@
...
-installs the dependencies as a set of SDKs
+installs the components as a set of SDKs
 and attaches custom actions for run-time control.
 IDEs such as Visual Studio Code or JupyterLab can discover workshops
 and use them in day-to-day operations,
 tidying up your system and streamlining your work.
 
-**Untangle the know-how woven into your project**.
-An environment that could take hours to set up
+**Focus on your project, not your setup**.
```
dmitry-lyfar, line 14: This one is much better.

docs/README.rst
```diff
@@ -1,25 +1,25 @@
 Workshop
 ========
 
-**Workshop is a tool that fractions development environment prerequisites
-into manageable pieces**.
+**A tool for decomposing development environments
```
dmitry-lyfar: I like the idea of dropping the product name here and freeing up some space.  I find it hard to agree that the tool decomposes development envs as its final purpose. As we discussed, this is a way (approach, tool) to manage your work environments. Decomposition / containerisation is how you do it.

Also, 'parts' do not ring a bell compared to how prerequisites does it (what is a part and what it means for the part to be structured or unstructured?). The 'part' is something I'd probably reserve for the SDK authors since sdkcraft is literally a way do decompose an SDK into parts.

Can you see if you can make one more pass on this one?

dmitry-lyfar:
Good stuff. Left some comments.

docs/explanation/advanced/content-interface.rst
akcano: Exactly, and I find it more and more necessary to use some semantic highlighting to distinguish the two.

docs/explanation/advanced/content-interface.rst
akcano: Fair point.

docs/explanation/essentials/project-workshop-sdks.rst
akcano: I'll add an extra commit to disambiguate once and for all.

docs/README.rst
akcano: Done.

docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,89 @@
...
+:ref:` validation and policy checks <exp_interfaces_validation>`
+that |project_markup| does internally.
+This involves making sure that the plug declaration is correct;
+the plug is allowed to be installed and auto-connected;
+and the destination directory actually exists.
+
+If the content interface plug passes the checks,
+the :samp:`target` directory, as defined by the plug,
+is mounted on the host file system.
+That's where the module cache from the our example will end up;
```
dmitry-lyfar:
```suggestion
That's where the module cache from our example will end up;
```

docs/explanation/advanced/content-interface.rst
```diff
@@ -0,0 +1,89 @@
...
+the best part is that it will be preserved between workshop
```
dmitry-lyfar: It will also be preserved between refreshes which is the main point of having the cache exposed as a content interface so that then next workshop has it already available. Availability between starts and stops does matter too, but it would still be available without a content interface as the container preserves the state anyway.

docs/explanation/essentials/project-workshop-sdks.rst
```diff
@@ -146,10 +146,121 @@ A single workshop can include multiple SDKs from different publishers.
...
 `snap channels <https://snapcraft.io/docs/channels>`_.
 
+
+SDK state
+~~~~~~~~~
+
 An SDK has a state that persists SDK-specific data,
 such as a model training configuration.
-|project| saves the state before applying any changes to the SDK,
+|project_markup| saves the state before applying any changes to the SDK,
```
dmitry-lyfar:
```suggestion
|project_markup| saves the state before applying any changes to the workshop,
```
Technically, a refresh operation creates a new workshop according to the workshop file and the old workshop gets removed when the former is successful. So I think it is correct to say that the change is applied to a workshop, not SDK here.

docs/explanation/essentials/project-workshop-sdks.rst
```diff
@@ -146,10 +146,121 @@ A single workshop can include multiple SDKs from different publishers.
...
+
+
+.. _exp_interfaces_validation:
+
+Interface validation and policies
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Now, to make sure plugs can be installed and auto-connected,
+|project_markup| uses a set of rules called policies,
+with each connection type having its own.
```
dmitry-lyfar:
```suggestion
with each interface having its own.
```

dmitry-lyfar (approved):

PR #88, opened 2024-01-29
Doc: Refactor structure, add initial SDK hooks info

akcano:


docs/reference/sdk-hooks.rst
```diff
@@ -0,0 +1,66 @@
+.. _ref_sdk_hooks:
+
+SDK life cycle hooks
```
dmitry-lyfar, line 3: This is a very clear view of the available hooks. The only minor suggestion I had is to organise the table rows to follow the order of the hooks execution: setup-base -> save-state -> restore-state -> check-health. However, given it is a reference, the existing alphabetical order makes more sense?

docs/explanation/essentials/sdks.rst
```diff
@@ -0,0 +1,185 @@
...
+usually during a :command:`workshop refresh`,
+|project_markup| saves the workshop's SDK states
+by invoking their :ref:`life cycle hooks <exp_sdk_hooks>`.
+After a successful change,
+the states are respectively restored.
+
+
+.. _exp_sdk_def:
+
+Definition
```
dmitry-lyfar, line 39: There is a PR-irrelevant though: can we call it a `definition` here given we put this sort of yaml under meta/ directory in the SDK archive AND this does not include hooks which would also make a part of the SDK's definition?

Also, not every field provided in the original sdkcraft.yaml goes into the final sdk.yaml: https://github.com/canonical/sdkcraft/blob/b4f0e2149d55d5c0bc222836bf5e84b154dd2f47/sdkcraft/services/package.py#L76

We should probably discuss it out of this PR's scope.

dmitry-lyfar (approved):

PR #92, opened 2024-02-09
workshopctl for SDK hooks

dmitry-lyfar:
# Description

This change sets the foundation for workshopctl and the check-health hook. The approach is taken from snapd and comprised of the following steps:

1. Expose an untrusted socket into a workshop. Currently, only the `/v1/workshopctl` endpoint is reachable from the untrusted socket which is mounted into every workshop on launch and refresh.
2. Mount `workshopctl`, a CLI program that interacts with the endpoint, into every workshop on launch and refresh.
3. Allow `workshopctl` to be called from the SDK hooks by passing a context cookie as a hook environment variable which `workshopctl` would capture and use. `workshopctl` can be used by SDK authors to communicate to workshopd about status and configuration. This change enables the first command, `set-health`, that lets an SDK to retry and/or report a failing health check.

See for the reference: https://snapcraft.io/docs/snapd-api#heading--snapctl

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```


dmitry-lyfar:
@akcano please note the tutorial update; I had to change the snap confinement to make it work correctly in a VM environment, hence, the tutorial update.

PR #109, opened 2024-03-20
Doc: Add workshop remount ref, exp

akcano:


cmd/workshop/info.go
```diff
@@ -28,6 +28,8 @@ details for a workshop, formatting them as YAML. Specifically, it prints:
 - Essential workshop attributes, such as name, base and project directory
 - Current status (e.g. *Ready*, *Pending*, *Off*) and notes for the workshop
 - Individual SDK details, such as name, channel, installation date and revision
+- For plugs remounted by 'workshop remount',
+  it prints non-default slots on the host
```
dmitry-lyfar: Remount operates at the "mounts" level, so the `info` will print mounts' hosts and workshop directories (not slots). It also prints the default directories. So maybe "prints mounts created by the content interface"?

docs/reference/cli/workshop-remount.rst
```diff
@@ -0,0 +1,64 @@
...
+
+
+Synopsis
+--------
+
+This command mounts a new source location on the host to the target directory
+of the specified content interface plug, qualified by the SDK name.
+Specifically, it does the following:
+
+- Checks if the new source is either a non-existing directory
```
dmitry-lyfar: It is almost true. Formally, we don't check anything but try to do it atomically right away. Because if you check and then do, there is a moment in between where the directory may become non-empty, for example. If the attempt to make a remount operation atomically fails, we retract to suggesting to stop the workshop first.

dmitry-lyfar (approved):

PR #116, opened 2024-04-09
Doc: connect, disconnect, connections CLI ref

akcano:


cmd/workshop/connect.go
```diff
@@ -14,16 +14,37 @@ type CmdConnect struct {
...
+This command connects a plug to a target slot
+that is specified as the second argument or deduced from the context.
+
+- If the second argument is omitted entirely, the target is assumed to be
+  <WORKSHOP>/agent:<PLUG>; <WORKSHOP> and <PLUG> come from the first argument
+
+- If the second argument only names the slot itself, the target is
+  <WORKSHOP>/agent:<SLOT>; <WORKSHOP> comes from the first argument
+
+- If the second argument only names the workshop and SDK, the target is
```
dmitry-lyfar, line 30: In the third case, the target slot must be the only one compatible with the plug. If there are multiple slots with <INTERFACE>, connect will fail.

cmd/workshop/connections.go
```diff
@@ -17,14 +17,24 @@ type CmdConnections struct {
...
 	var cmd = &cobra.Command{
-		Use:     "connections [<WORKSHOP>] [--all]",
-		Args:    cobra.MaximumNArgs(1),
-		Short:   "List interface connections.",
+		Use:   "connections [<WORKSHOP>]",
+		Args:  cobra.MaximumNArgs(1),
+		Short: "List interface connections.",
+		Long: `
+This command lists connections between interface plugs and slots
+globally or for a single workshop; each line is a connection between a plug
```
dmitry-lyfar: In our context 'globally' means across projects (see workshop list --global which would show all the projects' workshops). Should it be for a project or for a single workshop? The connections command does not have a global option compared to list.

docs/reference/cli/workshop-connections.rst
```diff
@@ -0,0 +1,61 @@
...
+.. code-block:: console
+
+   $ workshop connect <WORKSHOP>/<SDK>:<PLUG> [<WORKSHOP>/<SDK>][:<SLOT>] [OPTIONS]
+
+
+Synopsis
+--------
+
+This command lists connections between interface plugs and slots
+globally or for a single workshop; each line is a connection between a plug
```
dmitry-lyfar: Same note about using 'globally' here.

dmitry-lyfar (commented):
Nice one. I suspect you defer references to the agent SDK's documentation until it exists (e.g. in the commands reference)?

akcano:
> Nice one. I suspect you defer references to the agent SDK's documentation until it exists (e.g. in the commands reference)?

Yes, the agent-related stuff will be added separately.

cmd/workshop/connections.go
akcano: I suggest we follow what `snap connections` does, i.e. list all the snaps/workshops.

docs/reference/cli/workshop-connections.rst
akcano: See above.

dmitry-lyfar (approved):

PR #121, opened 2024-04-25
Explain connections; add warnings to debug how-to; restructure nav

akcano:


docs/explanation/index.rst
```diff
@@ -6,24 +6,84 @@ Explanation
...
+
+First, |project_markup| is transactional in nature;
+you won't have to trace residual files and libraries all across your system
+after you uninstall a package that turned out too unstable to your taste.
+Even if an SDK dumps something unexpected onto the disk,
+it's contained within the workshop.
+|project_markup| aims to encapsulate each part of functionality you may need,
+keeping things clean and tidy.
+
+Next, it's portable;
```
dmitry-lyfar: This bare minimum section makes it nice altogether, but I don't understand the sending a tarball part; should we talk about sharing a workshop definition here, instead? The one that another developer can take and `workshop launch`?

docs/how-to/debug-workshop-issues.rst
```diff
@@ -4,9 +4,10 @@ How to debug issues in workshops
 ================================
 
 To trace the condition of a misbehaving workshop,
```
dmitry-lyfar, line 6: This section matures! :-) Should we make another change here, namely, explain how to use --wait-on-error on refresh? The refresh used in the changes/tasks is transactional. I can extract this as a separate task as it is beyond the scope here.

docs/explanation/index.rst
akcano: I tipped the woring towards definition without mentioning it specifically.

docs/how-to/debug-workshop-issues.rst
akcano: Agreed verbally to leave this out of this PR's scope.

dmitry-lyfar (approved):

PR #125, opened 2024-05-07
Doc: Minor improvements

akcano:


akcano:
@dmitry-lyfar seems OK for a review

docs/reference/sdk-hooks.rst
```diff
@@ -33,7 +33,9 @@ SDK hooks
        after launching the new workshop
        and running the :samp:`setup-base` hook for the SDK.
 
-     - Restores SDK-specific data from a |project_markup|-defined location.
+     - Restores SDK-specific data,
+       collectively referred to as *state*,
```
dmitry-lyfar, line 37: hmm, that a bit of a collision with a workshop state. Just wanted to note that, not necessarily something we should jump at and change.

dmitry-lyfar: Actually, not. We have a workshop *status*

dmitry-lyfar (approved):

PR #127, opened 2024-05-10
Doc: Add GPU, SSH interfaces to 'Explanation/Advanced'

akcano:


docs/explanation/advanced/ssh-agent-interface.rst
```diff
@@ -0,0 +1,120 @@
+.. _exp_ssh_agent_interface:
+
+SSH agent interface
+===================
+
+The SSH agent interface
+enables access to the host system's SSH agent
+from inside the workshop
+to let it use the SSH keys and configuration on the host securely.
```
dmitry-lyfar:
```suggestion
to let it use the SSH keys and configuration of the host securely.
```
?

docs/explanation/advanced/ssh-agent-interface.rst
```diff
@@ -0,0 +1,120 @@
...
+
+
+SSH interface plug
+------------------
+
+An essential element here is the SSH interface plug
```
dmitry-lyfar: Maybe we should extract it as this section would be reproduced across every interfaces?

akcano: Agreed to implement later as an update on top of this.

PR #128, opened 2024-05-12
Doc: Add ROS2 how-to guide

akcano:


docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,204 @@
...
+Let's use the ROS 2 examples to build the workspace.
+First,
+`clone
+<https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Colcon-Tutorial.html#add-some-sources>`_
+the repo in your project directory,
+where you created the :file:`.workshop.ros2.yaml` file earlier:
+
+.. code-block:: console
+
+   $ git clone https://github.com/ros2/examples src/examples -b humble
```
artivis: I would clone directly in the 'project' folder to make it clearer that the build artifacts are somewhere else,

```suggestion
   $ git clone https://github.com/ros2/examples -b humble
```

Some of the following should then be adapted accordingly.

artivis: So the question really is, how are `.workshop.foo.yaml` file supposed to be distributed? Are they independant or should they live inside the project for which they are used? E.g. in this case, should the `.workshop.yaml` live inside the `https://github.com/ros2/examples` project?

akcano: It's probably unreasonable to expect that ROS 2 devs will include these files in their distribution (at least until Workshop's as ubiquitous as `git`). Workshops are defined and used locally by the consumers (in this case, anyone who wants to run ROS 2 examples inside a workshop).

Thus, the workshop definition is created, updated and maintained by the user as they see fit.

akcano: On the other hand, if you publish a ROS 2 project and want your users to run it with Workshop, it makes absolute sense to include at least a basic `.workshop.foo.yaml` in your repo.

dmitry-lyfar: Yes, I would expect the repo to have it like many other tools do. There is also an item on the road map to move it under .workshop to accommodate multiple workshops scenario.

In our examples, there is no file obviously, so we have to create it.

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
+.. _how_create_ros2_workspace:
+
+How to create a ROS 2 workspace
```
dmitry-lyfar: Shall we not call it workspace or should we be brave and not afraid of confusing workspace and workshops?

I suppose this title says: How to create a ROS2 dev environment / work environment?

On another note, there is also some mix of ROS2 (and, possibly, some other in the future) instructions and how-tos on the workshop tool. These are quite different level topics to me. Can we consider structuring them somehow else?

artivis: I think you're right and should change that for `How to create a ROS 2 workshop`.

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+.. code-block:: yaml
+   :caption: .workshop.ros2-humble.yaml
+
+   name: ros2-humble
+   base: ubuntu@22.04
+   sdks:
+     ros2:
+       channel: latest/edge
+
+
```
artivis: It could be useful to show the output of an `ls` command here just to emphasise again the layout, something along the line of,

```bash
~/path/to/dir$ ls -a
.workshop.ros2-humble.yaml examples
```

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+   $ workshop shell ros2-humble
+   workshop@ros2-humble-8584e57d$ cd /project/
+   workshop@ros2-humble-8584e57d$ ls
+
+     examples
+
+
+Here, you can see the :file:`examples/` directory that you've cloned.
+Now
```
artivis: I don't know if an 'how-to' is the place for that but I would promote in a single sentence the fact that the end user can immediately build the project. The SDK took care of installing and setting up the ROS environment, install the project dependencies so that the user only has to shell inside it and build out of the box.

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+the ROS 2 workspace:
+
+.. code-block:: console
+
+   workshop@ros2-humble-8584e57d$ colcon build
+
+
+Instead of the project directory,
+the results of the build end up in the :file:`~/colcon/` directory:
+
```
artivis:
```suggestion
Upon completion, all of the build artefacts can be found in the :file:`~/colcon/` directory:
```

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+`tests
+<https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Colcon-Tutorial.html#run-tests>`_
+and source the workspace:
+
+.. code-block:: console
+
+   $ workshop shell ros2-humble
+   workshop@ros2-humble-8584e57d$ cd ~/colcon/
+   workshop@ros2-humble-8584e57d$ colcon test
+   workshop@ros2-humble-8584e57d$ source install/setup.bash
```
artivis: Not sure why you source that here?

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+so you can proceed with the
+`tests
+<https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Colcon-Tutorial.html#run-tests>`_
+and source the workspace:
+
+.. code-block:: console
+
+   $ workshop shell ros2-humble
+   workshop@ros2-humble-8584e57d$ cd ~/colcon/
+   workshop@ros2-humble-8584e57d$ colcon test
```
artivis: Running `colcon test` literally only runs the test and provide no feedback to the user. To get the actual results, use the `test-result` verb:

```suggestion
   workshop@ros2-humble-8584e57d$ colcon test
   workshop@ros2-humble-8584e57d$ colcon test-result --all
```

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+   workshop@ros2-humble-8584e57d$ source install/setup.bash
+
+
+And so on; the workshop already contains the additional :program:`ros2` commands
+such as :command:`run` and :command:`launch`,
```
artivis: Not sure this is worth mentioning. We're only installing them because users expect them to be there and thus do not even think about it.

docs/how-to/create-ros2-workspace.rst
```diff
@@ -0,0 +1,216 @@
...
+as well as the dependencies specific to your source code
+that the SDK identified at launch time.
```
artivis: This should appear much earlier in the text. How did we build otherwise?

docs/how-to/create-ros2-workspace.rst
artivis: Gotcha. 
Workshop doesn't mount in `/project/src/` as a ROS user might intuitively expect, therefore the build artifact are enforced to live in `~/colcon` by the SDK as to not pollute the sources.

IHMO this is fine as long as it is explained and documented. 

However if the doc shows a project in a subfolder (as initially in this how-to `src/examples`) but then the build artifacts aren't where they should be `/project/{build,install,log}`, that would create confusion if not push-back.

All of that to drive the point that we should be careful with those examples.
And thanks for accommodating  ;)

docs/how-to/create-ros2-workspace.rst
akcano: As I mentioned in today's lightning talk ;), it's really undesirable to override existing terminology, but I definitely can accomodate and play around it to avoid confusion.

docs/how-to/create-ros2-workspace.rst
akcano: That's absolutely the place to do this.

docs/how-to/create-ros2-workspace.rst
akcano: Adjusted slightly and implemented.

PR #156, opened 2024-07-31
Update installation info, improve layout

akcano:


dmitry-lyfar:
@akcano we should consider not highlighting the product name at all and rely on the common sense that one can tell which is which from the context and from the spelling. I cannot help getting distracted by either bold or italic; perhaps, it is worth highlighting on the first occurrence.

akcano:
> @akcano we should consider not highlighting the product name at all and rely on the common sense that one can tell which is which from the context and from the spelling. I cannot help getting distracted by either bold or italic; perhaps, it is worth highlighting on the first occurrence.

Regular font is absolutely OK; if you remember, we started highlighting product names to distinguish Workshop from a workshop better, so if this is getting counterproductive, the highlight needs to go. However, I'm leaving the semantic markup in place to avoid losing context.

dmitry-lyfar (approved):

dmitry-lyfar:
@akcano there are some link checks failing, but overall looks good!

akcano:
that's transient but I want to investigate, please don't merge yet

PR #163, opened 2024-08-19
Replace a Multipass adhoc backend with LXD VMs in GitHub self-hosted runners

dmitry-lyfar:
# Description

Replace a Multipass adhoc backend with LXD VMs to launch spread integration and end-to-end tests with self-hosted GitHub runners. This allows for a customisable number of workers and fixes issues with cleaning up the environment on self-hosted runners (e.g. removing a multipass instance on a spread test failure).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -37,17 +37,14 @@ jobs:
         suite: [main, documentation, integration]
     runs-on: [self-hosted, linux, jammy, x64, xlarge]
     steps:
-      - uses: actions/checkout@v3
-      - name: Setup tmate session
-        if: ${{ failure() }}
-        uses: canonical/action-tmate@main
+      - name: Install LXD
+        uses: canonical/setup-lxd@main
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/spread.yaml
```diff
@@ -55,6 +67,7 @@ jobs:
           cd spread
           go install ./...
       - name: Run spread
-        run: | 
-          lxd init --auto
+        run: |  
+          # Installed docker related packages might interfere with LXD networking
+          sudo iptables -P FORWARD ACCEPT
```
jnsgruk: You can drop this if you use the `canonical/setup-lxd` GitHub Action which does it for you :)

dmitry-lyfar: Thanks!

jnsgruk (approved):

PR #169, opened 2024-09-12
Single-source CLI help for info

akcano:
# Description

This is a PoC implementation of CLI help single-sourcing.

# Pending issues

- Output formatting should be improved; not spending time on this now
- Robust way to include up-to-date options and flags (can be tied to release cadence)
- Placement of `.md` files - need to choose between code and docs directories

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```


dmitry-lyfar (commented):
This looks interesting. The colour, perhaps should go away completely, but you mentioned this in the comment.

The bigger problem is that one have to keep flags and the markdown in sync (vs. CLI help and readthedocs as before). 

I tried the ascii profile which looks clearer, I suppose:
Screenshot from 2024-09-13 15-26-34

There are two more questions:

1. Should we consider having a very brief help + example in the CLI output with everything else on the website?
2. Should we implement a .md generator with cobra instead to use in our readthedocs? That looks relatively easy to do with cobra (we can implement something similar to `docs.GenMarkdown()`). Also, in the future, if we really want a markdown `renderer` in the terminal, we can use the generator instead of embedding. So the `renderer` would get a generated markdown stream as input instead of the embedded file. Thus, no need to maintain flags documentation manually. That seems like the least maintenance approach.

akcano:
TL;DR: Third-party renderers such as Glamour seem too cumbersome, so I'll try Cobra's `doc.GenReST*` for my next iteration instead; with some tweaking, I hink I'll have it render the docs the way I envision them.

As for the flags, I briefly considered a workaround of having a secret toggle to print them the native Cobra way and update the Markdown accordingly, but that kinda defies the purpose.

1. We discussed this option earlier, and the consensus at the time was that it's viable, especially if we provide a documentation link, but other options should be explored first. Well, here's the exploration.
2. This is another option we considered briefly a while ago, but there was no good customisation available (for some reason, I don't remember including Cobra's `doc` in my review). Still, if we go this route, I'd probably start with the GenReST* family instead, since we're doing reST anyway. The other problem here was choosing when to run the generator; now that we do releases, each release point is a very good candidate for an automated CLI doc update.

https://umarcor.github.io/cobra/#generating-restructured-text-docs-for-your-own-cobracommand

PR #170, opened 2024-09-20
Set `workshop shell` cwd to /project

jy5275:
# Description

Currently,`workshop shell` opens a session with $HOME as a work directory. 

This PR changes it to `/project`, in order to be consistent with the default of `workshop exec`.

https://warthogs.atlassian.net/jira/software/c/projects/WSP/boards/1645?selectedIssue=WSP-36

# Self-review quick check

 * [√] Make decisions that cost a lot to reverse explicit in the PR description.
 * [√] Avoid nested conditions.
 * [√] Delete dead code and redundant comments.
 * [√] Normalise symmetries by sticking to doing identical things identically. 
 * [√] Check that coupled code elements, files, and directories are adjacent. For example, test data is stored as close as possible to a test.
 * [√] Put variable declaration and initialisation together.
 * [√] Divide large expressions into digestable and self-explanatory ones. Use multiple variables if required.
 * [√] Put a blank line between two logically different chunks of code.


dmitry-lyfar:
@jy5275 spread tests are failing.

jy5275:
> @jy5275 spread tests are failing.

Yes I deliberately added a fail command to see if spread test works. Will revert it

dmitry-lyfar:
Thanks, you'd need to rebase against main as the github workflows had to change due to some actions being moved across repos and breaking our flows, that was fixed in main.

cmd/workshop/exec.go
```diff
@@ -135,7 +135,7 @@ func (c *CmdShellAlias) Command() *cobra.Command {
 }
 
 func (cmd *CmdShellAlias) Run(c *cobra.Command, av []string) error {
-	return cmd.execCommand.Run(c, []string{av[0], "su", "-l", "workshop"})
```
dmitry-lyfar: @jy5275 what is the reason for dropping "-l"? Also, that does not seem to change cwd to /project?

jy5275:
Have updated to make `workshop shell` to get a login shell

internal/workshop/workshop.go
```diff
@@ -40,6 +41,7 @@ type Workshop struct {
 	Content map[string]sdk.Setup
 	// Workshop devices installed.
 	Profiles map[string]SdkProfile
+	UserData string
```
dmitry-lyfar: This is an internal detail, it should never be exposed publicly. Your integration tests need to use GetInstance LXD API to obtain the value of a user.* configuration attribute.

dmitry-lyfar:
```suggestion
```

dmitry-lyfar: Note how other properties here are agnostic of LXD API data types. These must not leak out to the client code. `user.user-data` key is a LXD specific configuration key, imagine if we want to replace LXD with some other backend.

internal/workshop/workshop.go
```diff
@@ -23,6 +23,7 @@ var (
 	ConfigWorkshopFile      = "user.workshop.file"
 	ConfigWorkshopContent   = "user.workshop.content"
 	ConfigProjectPathDevice = "workshop.project"
+	ConfigUserData          = "user.user-data"
```
dmitry-lyfar:
```suggestion
```

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -70,6 +71,24 @@ func (f *wsOps) TearDownSuite(c *check.C) {
...
+
+	userData := ws.UserData
+
+	// Check YAML format
+	var data interface{}
+	err = yaml.Unmarshal([]byte(userData), &data)
+	c.Assert(err, check.IsNil)
+
+	// Check prefix #cloud-config
+	c.Assert(userData[0:14], check.Equals, "#cloud-config\n")
```
dmitry-lyfar: @jy5275 strings.HasPrefix

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -70,6 +71,24 @@ func (f *wsOps) TearDownSuite(c *check.C) {
...
 }
 
+func (f *wsOps) TestLxdBackendWorkshopUserData(c *check.C) {
+	helper.LaunchTestWorkshop(c, f.ctx, f.bd, f.project.Path)
+	defer f.bd.RemoveWorkshop(f.ctx, "test")
+	ws, err := f.bd.Workshop(f.ctx, "test")
+	c.Assert(err, check.IsNil)
+	c.Assert(ws, check.NotNil)
+
+	userData := ws.UserData
```
dmitry-lyfar: You can use GetInstance LXD API to fetch the cloud init config.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -479,6 +479,7 @@ func (b *Backend) loadWorkshop(conn lxd.InstanceServer, inst *api.Instance, p *w
 		Content:  c,
 		Profiles: profs,
 		File:     f,
+		UserData: inst.Config[workshop.ConfigUserData],
```
```suggestion
```

dmitry-lyfar (changes requested):
@jy5275 it is getting closer, thank you, see my comments.

internal/workshop/workshop.go
jy5275: I added a method `lxdbackend.Backend.GetWorkshopConfig()` to replace `Workshop.UserData` field.

However, this requires Adding a type assertion in `workshop_test.go` (assert `Backend` interface type to `lxdbackend.Backend`). Could you help to check if this is acceptable?

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -70,6 +73,27 @@ func (f *wsOps) TearDownSuite(c *check.C) {
 	c.Check(err, check.IsNil)
 }
 
+func (f *wsOps) TestLxdBackendWorkshopUserData(c *check.C) {
+	helper.LaunchTestWorkshop(c, f.ctx, f.bd, f.project.Path)
+	defer f.bd.RemoveWorkshop(f.ctx, "test")
+	lxdBd, ok := f.bd.(*lxdbackend.Backend)
+	if !ok {
```
dmitry-lyfar: Thanks, what we should do here is to refactor the test suite so it works with a concrete implementation. Please update f.bd to be of `*lxdbackend.Backend` type and that will rid you of this unnecessary type assertion. This is an integration test for the LXD backend anyway.

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -70,6 +73,27 @@ func (f *wsOps) TearDownSuite(c *check.C) {
 	c.Check(err, check.IsNil)
 }
 
+func (f *wsOps) TestLxdBackendWorkshopUserData(c *check.C) {
```
dmitry-lyfar: The test should have a different name, it is not about user-data but about validating the cloud-init config (specifically, the part about changing a user's default directory).

dmitry-lyfar (changes requested):
Thank you, one more pass is needed, please also resolve the unresolved comments/suggestions.

internal/workshop/workshop.go
dmitry-lyfar: Answered in a separate thread.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -714,6 +738,12 @@ users:
     sudo: ALL=(ALL) NOPASSWD:ALL
     groups: adm,cdrom,sudo,dip,plugdev,audio,netdev,lxd,video,render
     shell: /bin/bash
+write_files:
```
dmitry-lyfar: Since there is an integration test, you can validate that it actually works by checking the content of .profile.

internal/workshop/lxd/tests/integration/workshop_test.go
jy5275: Thanks for pointing out. 

However, this function validates the whole config string, rather than just the module to change the default dir. So I think `TestLxdBackendWorkshopCloudInitConfig` may be better (without `ChangeDefaultDir...`)

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -70,6 +73,23 @@ func (f *wsOps) TearDownSuite(c *check.C) {
...
 
+func (f *wsOps) TestLxdBackendWorkshopCloudInitConfig(c *check.C) {
+	helper.LaunchTestWorkshop(c, f.ctx, f.bd, f.project.Path)
+	defer f.bd.RemoveWorkshop(f.ctx, "test")
+	wxCfg, err := f.bd.GetWorkshopConfig(f.ctx, "test")
+	c.Assert(err, check.IsNil)
+	c.Assert(wxCfg, check.NotNil)
+
+	userData := wxCfg["user.user-data"]
+	c.Assert(len(userData) > 0, check.Equals, true)
```
dmitry-lyfar:
```suggestion

```

internal/workshop/lxd/tests/integration/workshop_exec_test.go
```diff
@@ -251,3 +251,43 @@ func (f *wsExec) TestLxdBackendExecTimeout(c *check.C) {
...
+	stdout, stderr, err := f.exec(c, "", "test", f.project.ProjectId, opts)
+
+	// Validate
+	c.Assert(err, check.IsNil)
+	c.Assert(stderr, check.Equals, "")
+	c.Assert(strings.Contains(stdout, "Valid schema user-data"), check.Equals, true)
+	c.Assert(strings.Contains(stdout, "Error"), check.Equals, false)
+}
+
+func (f *wsExec) TestLxdBackendExecCheckProfile(c *check.C) {
```
dmitry-lyfar: @jy5275 I know that we discussed that above, but I think this test knows too much about the implementation detail. We need to shell into the workshop and confirm that CWD is /project. We don't have to check the content of the .profile as this may change, for example, for other shells.

dmitry-lyfar (commented):
Looks pretty good, see my comments.

internal/workshop/lxd/tests/integration/workshop_exec_test.go
jy5275: > We need to shell into the workshop and confirm that CWD is /project

Yes. But as discussed in a previous call, it's hard to check CWD of `shell` subcommand in `tests/main/shell/task.yaml`, as spread uses non-interactive shell. How should we achieve that?

dmitry-lyfar: @jy5275 why do we need interactivity in this integration test? You can execute a command in a login shell that is not interactive in this test, `sudo -i -u workshop -- pwd`

internal/workshop/lxd/tests/integration/workshop_exec_test.go
```diff
@@ -251,3 +251,41 @@ func (f *wsExec) TestLxdBackendExecTimeout(c *check.C) {
...
+		WorkingDir:  "/",
+		UserId:      new(int),
+		GroupId:     new(int),
+		Environment: map[string]string{"WORKSHOP_SHELL": "1"},
+	}
+
+	// Exec
+	stdout, stderr, err := f.exec(c, "", "test", f.project.ProjectId, opts)
+
+	// Verify that login dir is /home/workshop, not /project
```
dmitry-lyfar: @jy5275 the comment is incorrect.

tests/main/exec/task.yaml
```diff
@@ -23,6 +23,8 @@ execute: |
   echo "Check working directory setup"
   workshop_exec exec ws-exec -- pwd | MATCH '/project'
   workshop_exec exec -w /home/workshop ws-exec -- pwd | MATCH '/home/workshop'
+  workshop_exec exec ws-exec --uid=0 -- su -l workshop -w "WORKSHOP_SHELL" -c pwd | MATCH '/home/workshop'
```
dmitry-lyfar: @jy5275 it is ok to do this in the integration test, but what is the point of doing it in the end to end test? The end to end test must verify the user-facing behavior, which is `workshop shell`. Now this test (almost) duplicates the integration test.

internal/workshop/lxd/tests/helper/helper.go
```diff
@@ -21,9 +21,11 @@ base: ubuntu@22.04
 var MinimalImageServer = "simplestreams:https://cloud-images.ubuntu.com/minimal/releases/"
 
 func DefaultTestDevices() map[string]map[string]string {
+	os.MkdirAll("/tmp/workshop/project", 0755)
```
dmitry-lyfar: @jy5275 who is going to clean this directory after the tests have ran? See c.Mkdir() for such purposes.

You can create a wrapper func for DefaultTestDevices which has a project as an argument and returns a `func ()map[string]map[string]string` to accommodate this change in this test suite.

dmitry-lyfar (changes requested):

internal/workshop/lxd/tests/helper/helper.go
jy5275: I've updated code to ensure the temporary project dir gets removed after each time intergraion runs.

In order not to change the signature of `DefaultTestDevices`, I created new functions `getDefaultDevicesWithProject` and `FakeDefaultDevicesWithProject` - that makes the invokation chain a little more complicated. Could you help to see if it's acceptable?

dmitry-lyfar: @jy5275 I don't see it in the latest commit, there is only the e2e test removal.

What I mean is:

```
func execTestDevices(projectDir string) func () map[string]map[string]string {
    conf := DefaultTestDevices()
    conf["workshop.project"] = projectDir
    return func () map[string]map[string]string {
                return conf
    }
}
```

which then can be called:

```
	f.restoreDevices = lxdbackend.FakeDefaultDevices(execTestDevices(c.Mkdir()))
```

so, you don't have to delete it manually, it is done by the test framework already.

internal/workshop/lxd/tests/integration/workshop_exec_test.go
```diff
@@ -93,7 +93,8 @@ func (f *wsExec) SetUpSuite(c *check.C) {
 		return u, nil
 	}, &daemon.LookupUserId)
 
-	f.restoreDevices = lxdbackend.FakeDefaultDevices(helper.DefaultTestDevices)
+	tmpProjectPath := "/tmp/workshop/project"
+	f.restoreDevices = lxdbackend.FakeDefaultDevicesWithProject(helper.DefaultTestDevices, tmpProjectPath)
```
dmitry-lyfar: @jy5275 we don't need Fake function for every variation of the DefaultTestDevices, that's why the adapter I suggested should be used, so you don't have to create another Fake* function

internal/workshop/lxd/lxd_backend.go
```diff
@@ -780,3 +780,22 @@ func FakeDefaultDevices(f func() map[string]map[string]string) func() {
 	defaultDevices = f
 	return func() { defaultDevices = oldDefault }
 }
+
+func getDefaultDevicesWithProject(oldFunc func() map[string]map[string]string,
+	projectPath string) func() map[string]map[string]string {
+	return func() map[string]map[string]string {
+		devices := oldFunc()
+		os.MkdirAll(projectPath, 0755)
```
dmitry-lyfar: @jy5275 please check my example with more attention to how it should be done. It must use c.Mkdir. See how it is done elsewhere in the codebase, do not do it inconsistently. It is a common problem that has been solved, no need to invent another solution.

jy5275: Thanks for your suggestion. I just noticed that the refered function should be `c.MkDir` - I misunderstood it as `os.Mkdir` before.
Things get a lot easier with the test framework automatically removing the dir for us.

lachypjones:
Hey @jy5275 
The '-D' option (and the associated Chdir_Spec) are only compatible with sudo versions 1.9.3 or higher (see https://www.sudo.ws/docs/man/sudoers.man/#Chdir_Spec). 
Ubuntu 20.04 uses sudo versions 1.8.x (see https://launchpad.net/ubuntu/focal/+source/sudo). 

This change would unfortunately mean that shell wouldn't work on a workshop using 20.04.

Can you see if there are any workarounds?

jy5275:
> Hey @jy5275 The '-D' option (and the associated Chdir_Spec) are only compatible with sudo versions 1.9.3 or higher (see https://www.sudo.ws/docs/man/sudoers.man/#Chdir_Spec). Ubuntu 20.04 uses sudo versions 1.8.x (see https://launchpad.net/ubuntu/focal/+source/sudo).
> 
> This change would unfortunately mean that shell wouldn't work on a workshop using 20.04.
> 
> Can you see if there are any workarounds?

Thanks for pointing that out @lachypjones .

I just found sudo versions < 1.9.3 doesn't have an similar option to change working dir, and Ubuntu version < 22.04 can't upgrade sudo package to 1.9.3 

@dmitry-lyfar I suggest to adopt the previous way (update `/etc/profile.d/xxx.sh` to `cd /project` when starting a login shell) to achieve the goal. 

Specifically, revert [this commit](https://github.com/canonical/workshop/pull/170/commits/b905954c2fc0a9429ee9a55b994dd2d6954874d9). Is that OK?

dmitry-lyfar:
@jy5275 we will have to revert to support 20.04, but I would suggest a slightly different solution to not make the workshop containers depended on the default shell and keep it in the client for now (so no changes to the cloud-init):

```
sudo -i -u workshop bash -c "cd /project; exec bash -l"
```

jy5275:
> @jy5275 we will have to revert to support 20.04, but I would suggest a slightly different solution to not make the workshop containers depended on the default shell and keep it in the client for now (so no changes to the cloud-init):
> 
> ```
> sudo -i -u workshop bash -c "cd /project; exec bash -l"
> ```

I found that `sudo -u workshop bash -l` already satisfies our requirements, because `bash -l` doesn't change working dir. I've updated test cases accordingly.

dmitry-lyfar (approved):

PR #176, opened 2024-09-27
Move postRunWarnings to root command

jonathan-conder:
# Description

- Removed some old `struct` tags which don't seem to be used.
- Replaced `PostRun` on most commands with `PersistentPostRun` on the root command. This needs access to the client, which is now stored at the `CmdRoot`.
- All subcommands now store a `*CmdRoot` and use `CmdRoot.client()` to access the client object.
- The `Project` global is now stored with the `CmdRoot`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/cmd/workshop/changes.go
+++ b/cmd/workshop/changes.go
@@ -11,7 +11,7 @@ import (
 )
 
 type CmdChanges struct {
-	clientMixin
+	root *CmdRoot
 }
 
 func (c *CmdChanges) Command() *cobra.Command {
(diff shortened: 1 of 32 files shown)


dmitry-lyfar:
@jonathan-conder that looks all good, thanks! Can you put a brief note in the PR Description to summarise changes?

dmitry-lyfar (approved):

jonathan-conder:
sure, updated now

PR #178, opened 2024-10-01
Warn if project directory or workshop definition does not exist

jonathan-conder:
# Description

Adds warnings in addition to status for `missing-file` and `missing-project` checks.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/internal/overlord/workshopstate/manager.go
+++ b/internal/overlord/workshopstate/manager.go
@@ -125,6 +125,8 @@ func (w *WorkshopManager) WorkshopHealth(ws *workshop.Workshop) healthstate.Heal
 
 	// Check the project directory exists.
 	if !ws.Project.Exists() {
+		w.state.Warnf("%q project directory %q does not exist", ws.Name, ws.Project.Path)
+
 		healthState.Status = healthstate.ErrorStatus
 		healthState.Code = "missing-project"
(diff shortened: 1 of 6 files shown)


jonathan-conder:
The documentation checks failed initially. Think it was a transient network error but you might want to double-check

akcano:
@jonathan-conder I can't see anything on the logs; the PR scope also doesn't affect anything these checks, well, check.

jonathan-conder:
I reran it already, the original logs are here: https://github.com/canonical/workshop/actions/runs/11138115693/job/30952613787?pr=178

dmitry-lyfar (approved):

dmitry-lyfar:
@jonathan-conder looks pretty good, merging it. @akcano it fails with `https://discourse.canonical.com/c/engineering/sdk/34`, not sure how it accesses a private link, but that seems due to the Discourse outage which happens at times, so transient.

akcano:
@dmitry-lyfar the link check normally bumps into a login gateway redirect, which is legit in its opinion. If Discourse is down, though, no redirects occur, hence the transient failure.

PR #179, opened 2024-10-01
Allow exec to inherit environment variables via `--env NAME`

jonathan-conder:
# Description

This can save some typing: instead of `--env "FOO=$FOO"` you can write `--env FOO`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/docs/reference/cli/workshop-exec.rst
+++ b/docs/reference/cli/workshop-exec.rst
@@ -89,9 +89,10 @@ Options
 
   Set the working directory in the workshop (default: :file:`/project/`).
 
---env <KEY=VALUE>
+--env <KEY>[=<VALUE>]
 
-  Set an environment variable.
+  Set an environment variable, e.g. 'FOO=bar';
+  if only the name is provided, the value is inherited from the CLI environment.
 
 --uid <USER ID>
 
--- a/tests/main/exec/task.yaml
+++ b/tests/main/exec/task.yaml
@@ -9,7 +9,12 @@ execute: |
   . "$TESTSLIB"/utils.sh
 
   echo "Check an environment variable can be set"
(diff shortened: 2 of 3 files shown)


dmitry-lyfar (approved):

docs/reference/cli/workshop-exec.rst
```diff
@@ -89,9 +89,10 @@ Options
 
   Set the working directory in the workshop (default: :file:`/project/`).
 
---env <KEY=VALUE>
+--env <KEY>[=<VALUE>]
 
-  Set an environment variable.
+  Set an environment variable,
+  inheriting unspecified values from the CLI environment.
```
akcano:
```suggestion
Set an environment variable;
if only the name is provided, the value is inherited from the CLI environment.
```

akcano: @jonathan-conder the same text should occur in `exec.go` at line 111, give or take the `FOO=bar` example.

tests/main/exec/task.yaml
```diff
@@ -9,7 +9,12 @@ execute: |
   . "$TESTSLIB"/utils.sh
 
   echo "Check an environment variable can be set"
-  workshop_exec exec --env 'Foo=bar'  ws-exec -- bash -c 'echo -n $Foo' | MATCH bar
+  workshop_exec exec --env 'Foo=bar'  ws-exec -- bash -c '[ "$Foo" = bar ]'
+
+  echo "Check environment variables are inherited"
```
akcano:
```suggestion
  echo "Check that environment variables are inherited"
```

akcano (changes requested):

akcano (approved):

PR #183, opened 2024-10-03
Add shellcheck and rectify existing linting issues

lachypjones:
# Description
Shellcheck is a well known shell script static analysis tool. 
- Add shellcheck actions covering standard shell scripts and scripts contained in spread files
- Add snapd-testing-tools as a submodule to facilitate spread shell testing
- Rectify existing shellcheck warnings

# Self-review quick check
 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/tests/lib/utils.sh
+++ b/tests/lib/utils.sh
@@ -4,10 +4,10 @@ function setup_lxd() {
   snap install --classic lxd
   snap refresh --channel=5.21/stable lxd
   lxd waitready --timeout=180
-  
+
   # can already be initialised if reused
   # https://discuss.linuxcontainers.org/t/how-do-i-know-if-lxd-is-initialized/15473/3
-  if [ $(lxc storage list -f compact | grep -c default) -eq 0  ]; then   
+  if [ "$(lxc storage list -f compact | grep -c default)" -eq 0  ]; then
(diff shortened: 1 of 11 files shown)


lachypjones:
Need to determine relative locations within the github runner to fix shellcheck lint.

tests/lib/utils.sh
```diff
@@ -70,6 +70,8 @@ function publish_test_sdk_content() {
       SDK_PATH=$(readlink -f "$i")
       STORE_PATH="$2"/"$SDK_NAME"/latest/stable/
 
+      # quoting the following results in incorrect behavior
```
dmitry-lyfar: @lachypjones we need a specific reason, not just "incorrect behaviour". Otherwise, the explanation does not make sense for the reader.

dmitry-lyfar (changes requested):
@lachypjones thanks! As an exception, we should rebase this PR to ensure clearer commit history. Generally, before pushing a PR for a review, make sure that all tests are passing locally, that'll help to reduce round trips and extra commits.

dmitry-lyfar (approved):
Thank you! Once updated the description, please merge. Good stuff.

PR #187, opened 2024-10-06
Move workshop definitions into .workshop directory

jonathan-conder:
# Description

Workshop definitions will now look like `.workshop/workshop.name.yaml`. The old format is no longer supported, but can be re-enabled temporarily by reverting 54e239eeb01ea756d4abe25038f255d3f2b222f6.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```


akcano (approved):
@jonathan-conder awesome work, LGTM. I'll double-check some sections this pulse, but everything runs smoothly and seems to be in order at first glance.

internal/overlord/workshopstate/manager.go
```diff
@@ -133,9 +133,9 @@ func (w *WorkshopManager) WorkshopHealth(ws *workshop.Workshop) healthstate.Heal
 	}
 
 	// Check if the associated workshop file exists. We only check if that file
-	// exists in the project directory here; its state (e.g. if it is in sync
+	// exists in the .workshop directory here; its state (e.g. if it is in sync
 	// with the workshop instance or has any errors) is not checked.
-	path := filepath.Join(ws.Project.Path, workshop.Filename(ws.Name))
+	path := filepath.Join(ws.Project.Path, workshop.Directory, workshop.Filename(ws.Name))
```
dmitry-lyfar: That feels like deserving a method for the Workshop which would return its full filepath so you don't have to join in the client code.

internal/overlord/workshopstate/manager_test.go
```diff
@@ -130,10 +134,10 @@ func (s *managerSuite) TestWorkshopHealthMissingFile(c *check.C) {
 	s.state.Lock()
 	defer s.state.Unlock()
 
-	workshop := s.launchWorkshopWithSDKs(c, "test", nil)
-	path := filepath.Join(s.project.Path, ".workshop.test.yaml")
+	testWorkshop := s.launchWorkshopWithSDKs(c, "test", nil)
+	path := filepath.Join(s.project.Path, workshop.Directory, workshop.Filename(testWorkshop.Name))
```
dmitry-lyfar: this would also be easier if the workshop had a filepath method:
```
testWorkshop.Filepath()
```

dmitry-lyfar (approved):
Great stuff. Thank you for updating the documentation as well. It feels so much cleaner to have workshop definitions under .workshop/.

dmitry-lyfar (approved):

jonathan-conder:
I can reproduce the spread failure locally (but it was working a few hours ago...), rerunning with `-debug` now

PR #188, opened 2024-10-07
Use cobra's doc package for CLI reference doc generation

akcano:
# Description

This moves CLI subcommands and the root command to a new `cli` package that is used by both `workshop` and the new `gendocs` utility that generates reST reference docs for the CLI under `./docs-auto` using the [GenReSTTreeCustom()](https://umarcor.github.io/cobra/#customize-the-output-1) function.

This is a draft; a complete implementation needs to override `cobra/doc`, at least to to:

- Drop [See Also](https://github.com/spf13/cobra/blob/78bfc837fe358c750faa7e7f0a8016b300044d58/doc/rest_docs.go#L97) logic
- Adjust section naming and heading style to our docs
- Possibly provide better templating (this may find use upstream and at Canonical)

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```


dmitry-lyfar:
@akcano including @jonathan-conder as the one who has recently refactored the client.

cmd/workshop/main.go
```diff
@@ -2,90 +2,12 @@ package main
...
-	Stdin  io.Reader = os.Stdin
-	Stdout io.Writer = os.Stdout
-	Stderr io.Writer = os.Stderr
-)
-
-// ClientConfig is the configuration of the Client used by all commands.
-var ClientConfig = client.Config{
-	// we need the powerful socket
-	Socket: dirs.SocketPath,
-}
```
jonathan-conder, line 88: I think these globals should stay in `main.go`. Actually I'm not sure how `main_test` is passing now because it seems to modify `Stdin`/`Stdout`/`Stderr`.

cmd/workshop/main.go
```diff
@@ -98,10 +20,10 @@ func main() {
 
 	logger.SetLogger(l)
 
-	rootCmd := (&CmdRoot{}).Command(cwd)
+	rootCmd := (&cli.CmdRoot{}).Command(cwd)
 
 	if err = rootCmd.Execute(); err != nil {
-		fmt.Fprintf(Stderr, "error: %v\n", err)
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
jonathan-conder:
```suggestion
		fmt.Fprintf(cli.Stderr, "error: %v\n", err)
```
(if you leave `Stderr` in `root.go`)

cmd/gendocs/gendocs.go
```diff
@@ -0,0 +1,39 @@
...
+
+func main() {
+	cwd, err := os.Getwd()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "could not get current directory: %v\n", err)
+	}
+
+	rootCmd := (&cli.CmdRoot{}).Command(cwd)
+
+	docDir := "./docs-auto"
```
jonathan-conder:
```suggestion
	docDir := "docs-auto"
```
This is just a slight personal preference

cmd/gendocs/gendocs.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/canonical/workshop/cmd/cli"
+)
+
+func filePrepender(filename string) string {
+	return ""
+}
+
+func linkHandler(name, ref string) string {
+	return fmt.Sprintf(":ref:`%s <%s>`", name, ref)
```
jonathan-conder: This worries me a little bit, do we need to think about escaping/urlencoding?

cmd/workshop/main.go
akcano: I added the stream variables back to `main.go`. The file `main_test.go` is now `root_test.go`, so it stays with the subcommands in `cli` and modifies the streams there.

Did you mean to include `ClientConfig` here? I don't think `main.go` needs it .

cmd/workshop/main.go
akcano: Well spotted. I reverted these lines to use the variables inmstead.

cmd/gendocs/gendocs.go
akcano: No strong preference here. OK.

cmd/gendocs/gendocs.go
akcano: Basically, the default implementation [supplies the command name and its underscore-encoded variant](https://github.com/spf13/cobra/blob/78bfc837fe358c750faa7e7f0a8016b300044d58/doc/rest_docs.go#L104) here, so I'd start worrying only if we overload it later to generate fancy URLs. No such plans exist now.

cmd/workshop/main.go
jonathan-conder: I just think it makes sense to leave globals at the top level, but it's probably fine to leave it in `root.go`. At some point I guess we could move it into `CmdRoot` anyway.

akcano: OK, I'll get them to `root.go` and use `cli.` in `main.go`.

akcano:
If this PoC is fine with us, my next step will be to get it ready for our purposes during this pulse.

I outlined some ideas in the PR description, but the ultimate goal is to avoid any post-generation customisations, e.g. creative `cat`-ting and `sed`-ding, but still get a nice shiny one-pager with arbitrary cross-references that wouldn't look out of place in our docs.

cmd/workshop/main.go
jonathan-conder: oh I didn't notice that `root_test.go` existed originally. Makes more sense now, sorry about the confusion. I would approve but I don't think I can since it's a draft

akcano:
This round required me to borrow some code from `cobra/doc`, but it will have to go in the future; in lieu of that, I'm going to use templating in some form because this 'generation' is abysmal. The point was to slap together an end-to-end code-based generation solution.

The other issue not addressed here is that I'm going to move our existing examples to `cmd.Long` for the time being, but it probably would be best to come up with a way to involve `cmd.Example`, which in theory should provide the proper way to display usage examples at runtime and in docs.

dmitry-lyfar:
@akcano I think this is coming out great. I restructured the packages and rebased against the latest main changes to get rid of the conflicts.

Since I cannot force-push to a PR from a forked repository, there is a new branch: https://github.com/canonical/workshop/tree/docs/generate-ref-docs. Please use this for the work remaining (if any) and raise a PR when ready. I'm closing this one now.

PR #193, opened 2024-10-14
Add basic camera interface

jonathan-conder:
# Description

This is a first attempt, we plan on using `unix-hotplug` devices after https://github.com/canonical/lxd/issues/14266 is resolved (ideally we could ask LXD to expose all devices in the `media` and `video4linux` subsystems).

For now we just expose `/dev/video[0-9]` as `unix-char` devices. This also supports hotplugging but is limited to 10 devices and requires manually setting the device user and group to `1000`.

So far we have a 1-1 correspondence between plugs and devices. This implementation creates a `none` device with the plug name, to help reconstruct an `SdkProfile` from a LXD profile (similar to the mount interface). The `unix-char` devices are named like `<PLUG>/video<N>`, which shouldn't clash with other plugs because their names cannot contain `/`. The other [suitable separators](https://github.com/canonical/lxd/blob/7eff9d101e3a7c3407ab8dee8a3249ec6778da50/shared/validate/validate.go#L750) are `._:`. All of these [can also be used](https://github.com/canonical/lxd/blob/7eff9d101e3a7c3407ab8dee8a3249ec6778da50/lxd/config/map.go#L200) in `user.workshop` config values, but `.` might be confusing. I think `:` is also a reasonable choice, but for the hotplug devices I was using the convention `<PLUG>/<VENDORID>:<PRODUCTID>` so I slightly prefer `/` in that case.

Another detail is that spread now uses 24.04 VM images. This is to enable fake camera devices using `v4l2loopback`. The 22.04 images use a custom kernel which doesn't include the `videodev` module (even as a separate package, as far as I could tell). I ran into an issue with spread choosing container images instead of VM images, which I have a fix for in [my fork](https://github.com/jonathan-conder/spread).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/.spread.yaml
+++ b/.spread.yaml
@@ -13,7 +13,7 @@ repack: |
 backends:
   lxd:
     systems:
-      - ubuntu-22.04:
+      - ubuntu-24.04:
           workers: 2
 suites:
   tests/main/:
--- a/docs/explanation/interfaces/camera-interface.rst
+++ b/docs/explanation/interfaces/camera-interface.rst
@@ -26,8 +26,8 @@ can be invoked manually after the workshop has started:
 Establishing a connection means
 that all currently connected USB-based cameras
 will be made available inside the workshop.
-New cameras can be added by disconnecting and then reconnecting
-the camera interface.
+New cameras can be added
+by disconnecting and then reconnecting the camera interface.
 
 To check if the interface is connected:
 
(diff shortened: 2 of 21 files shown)


jonathan-conder:
First run encountered network issues: https://github.com/canonical/workshop/actions/runs/11320216655/job/31477570678
Seems to be a one off

dmitry-lyfar (approved):

PR #194, opened 2024-10-14
Enable linger for workshop user

lachypjones:
# Description
## Background
`/run/user/<uid>` gets created by pam_systemd when a user has an _active_ login shell. 
This creates a problem when an interface (ie. Desktop) looks to use this directory to proxy sockets into. 

### Linger
This PR enables lingering (`loginctl enable-linger`) for the workshop user as part of the container start script. Linger starts the user manager for `workshop` on boot, enabling interfaces that rely on resources in `$XDG_RUNTIME_DIR` to work.  

Linger introduces no modification to the system environment beyond what a normal login session does, namely:

- `/run/user/$UID` is either created or mounted
- `$XDG_SESSION_ID` is initialised
-  A new systemd scope unit is created for the session (user@1000.service, user-1000.slice)

https://www.freedesktop.org/software/systemd/man/252/pam_systemd.html

### Start Command
This PR also moves the start command to an external shell script (`start_command.sh`) to enable automated linting with CI runners. The script is embedded into workshopd at compile time.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

### DIFF
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -2,6 +2,7 @@ package lxdbackend
 
 import (
 	"context"
+	_ "embed"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -39,6 +40,9 @@ var (
 	imageLock        sync.Mutex
 	currentDownloads map[string]*downloadOp
 )
+
+//go:embed start_command.sh
+var startCommand string
 
 func init() {
 	imageLock.Lock()
@@ -176,17 +180,12 @@ func (s *Backend) StartWorkshop(ctx context.Context, name string) error {
 		return err
 	}
 
-	// Wait until system is up an running before returning
-	// see: https://blog.simos.info/how-to-know-when-a-lxd-container-has-finished-starting-up/
 	args := workshop.Execution{
 		ExecArgs: workshop.ExecArgs{
 			UserId:  0,
 			GroupId: 0,
 			Command: []string{
-				"bash", "-euc", fmt.Sprintf(`while
-[ "$(systemctl is-system-running 2>/dev/null)" != running ] &&
-[ "$(systemctl is-system-running 2>/dev/null)" != degraded ]; do :; done &&
-loginctl enable-linger workshop`),
+				"bash", "-euc", startCommand,
 			},
 			WorkDir: "/",
 		},
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -183,9 +183,10 @@ func (s *Backend) StartWorkshop(ctx context.Context, name string) error {
 			UserId:  0,
 			GroupId: 0,
 			Command: []string{
-				"bash", "-eu", "-c", "while " +
-					"[ \"$(systemctl is-system-running 2>/dev/null)\" != \"running\" ] && " +
-					"[ \"$(systemctl is-system-running 2>/dev/null)\" != \"degraded\" ]; do :; done",
+				"bash", "-euc", fmt.Sprintf(`while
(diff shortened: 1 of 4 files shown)


internal/workshop/lxd/lxd_backend.go
```diff
@@ -183,9 +187,7 @@ func (s *Backend) StartWorkshop(ctx context.Context, name string) error {
 			UserId:  0,
 			GroupId: 0,
 			Command: []string{
-				"bash", "-eu", "-c", "while " +
-					"[ \"$(systemctl is-system-running 2>/dev/null)\" != \"running\" ] && " +
-					"[ \"$(systemctl is-system-running 2>/dev/null)\" != \"degraded\" ]; do :; done",
+				"bash", "-euc", string(startCommand),
```
dmitry-lyfar: @lachypjones startCommand is already a string, why typecast here?

dmitry-lyfar:
@jonathan-conder can you have a look into this change as well, it is important? We need side-effects of having a login session to enable something like a Desktop interface.

jonathan-conder:
Sure, I think something like this might be useful for the camera interface too but I'll try to see if there are any reasons not to do something like this

internal/workshop/lxd/start_command.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+while [ "$(systemctl is-system-running 2>/dev/null)" != running ] \
```
dmitry-lyfar, line 4: @lachypjones please move the comments about the way we query systemd here (+link to the article).

dmitry-lyfar (approved):
looks good to me. Please note, we don't force push if comments were received for a PR: https://canonical-workshop.readthedocs-hosted.com/en/latest/contributing/#coding

jonathan-conder (approved):
Seems fine as far as I can tell. I think we should test the properties we want directly, e.g. does /run/user/1000 exist. But that can be covered in the interface-desktop tests.

Only relevant thing I found from searching is this TODO item which I don't really understand, but also don't think it will break anything for us:  https://github.com/systemd/systemd/commit/648b122045182a1ac19d1622817e2a350beb0354

dmitry-lyfar:
@lachypjones this is a good idea, can we extend the existing e2e test to check that the runtime dir is set? Are you expecting `$XDG_RUNTIME_DIR` to be set for the user login session (and /run dir exists as a result) compared to when not enabled, the directory won't exist:

```
workshop exec test -- test -d /run/user/1000
```

jonathan-conder:
One thing I noticed is linger doesn't actually persist the user session, only the slice/services, e.g.:

```console
$ workshop exec ws -- loginctl enable-linger workshop
$ workshop exec ws -- loginctl list-sessions
No sessions.
$ workshop exec ws -- ls /run/user
1000
```

dmitry-lyfar:
Ah, yes, true, it starts the [user manager](https://www.freedesktop.org/software/systemd/man/latest/loginctl.html#:~:text=Enable%2Fdisable%20user%20lingering%20for,or%20numeric%20UIDs%20as%20argument.) which results in the dir creation: 

> Enable/disable user lingering for one or more users. If enabled for a specific user, 
> a user manager is spawned for the user at boot and kept around after logouts.

lachypjones:
Thanks all,
I'll modify the test to check for the _resultant_ behavior of linger rather than checking whether linger is enabled (good call @jonathan-conder).

As for the user-manager commentary, do we feel my the comments on the PR don't reflect this enough?

jonathan-conder:
> As for the user-manager commentary, do we feel my the comments on the PR don't reflect this enough?

Na I think it's fine

PR #196, opened 2024-10-15
Refactor Agent into generic Proxy interface

lachypjones:
# Description
The ssh-agent currently exists as a singleton proxy resource. This PR refactors ssh-agent and the handling of proxy-type devices to enable having multiple, similar to how mount currently works.

Some decisions (such as the method by which the new interface modifies the environment) have potential ramifications should variables stored in /etc/profile.d not be suitable for future proxy type devices. This however is not a 'one-way' decision and can be easily reworked should this be a requirement in future, I feel the complexity introduced is not worth it for a _potential_ future use-case. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/builtin/ssh_agent.go
+++ b/internal/interfaces/builtin/ssh_agent.go
@@ -116,7 +116,11 @@ func (iface *sshAgentInterface) MountConnectedPlug(spec *lxd_device.Specificatio
 	fromSocket := sock
 	toSocket := filepath.Join(dirs.WorkshopRunDir, name+".ssh")
 
-	return spec.SetSshAgent(workshop.SshAgent{Name: name, Connect: fromSocket, Listen: toSocket})
+	return spec.AddProxyEntry(workshop.Proxy{
+		Name:    name,
+		Connect: fromSocket,
+		Listen:  toSocket,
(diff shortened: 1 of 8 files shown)


internal/interfaces/builtin/ssh_agent.go
```diff
@@ -116,7 +116,11 @@ func (iface *sshAgentInterface) MountConnectedPlug(spec *lxd_device.Specificatio
 	fromSocket := sock
 	toSocket := filepath.Join(dirs.WorkshopRunDir, name+".ssh")
 
-	return spec.SetSshAgent(workshop.SshAgent{Name: name, Connect: fromSocket, Listen: toSocket})
+	return spec.AddProxyEntry(workshop.Proxy{
+		Name:    name,
+		Connect: fromSocket,
+		Listen:  toSocket,
+		Env:     []string{"SSH_AUTH_SOCK=" + strings.TrimPrefix(toSocket, "unix:")}})
```
dmitry-lyfar, line 123: The Env should not be part of the Entry. The entry should contain a type so you can parse it back into a profile. The install/remove logic will know what variables need to exist.

PR #197, opened 2024-10-15
Fix indefinite retries if the store cannot be reached

lachypjones:
# Description
Simple, small quality of life fix for the current fake store implementation. This will time out a failed connection to the store after 10 seconds.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/fakestore/store.go
+++ b/internal/fakestore/store.go
@@ -219,6 +219,10 @@ func storeSdkInfoImpl(ctx context.Context, name, channel string) (storeSdk, erro
 	}
 	track, risk := sa[0], sa[1]
 	obj := bkt.Object(fmt.Sprintf("%s/%s/%s/%s.sdk", name, track, risk, name))
+
+	// Attrs transparently retries forever if it cannot connect to the store
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
 	atr, err := obj.Attrs(ctx)
 	if err != nil {
 		if errors.Is(err, storage.ErrObjectNotExist) {
--- a/internal/fakestore/store.go
+++ b/internal/fakestore/store.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


internal/fakestore/store.go
```diff
@@ -219,6 +219,10 @@ func storeSdkInfoImpl(ctx context.Context, name, channel string) (storeSdk, erro
 	}
 	track, risk := sa[0], sa[1]
 	obj := bkt.Object(fmt.Sprintf("%s/%s/%s/%s.sdk", name, track, risk, name))
+
+	// Attrs transparently retries forever if it cannot connect to the store
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
dmitry-lyfar: @lachypjones we need a unit test for this, as per the comment, is it going to retry forever on every timeout?

akcano:
The issue appears transient and unrelated to the latest commit:
image

internal/fakestore/store.go
```diff
@@ -219,11 +220,21 @@ func storeSdkInfoImpl(ctx context.Context, name, channel string) (storeSdk, erro
...
+	// passed on to SDK download behavior
+	obj = obj.Retryer(
+		storage.WithMaxAttempts(3),
+	)
 	atr, err := obj.Attrs(ctx)
 	if err != nil {
 		if errors.Is(err, storage.ErrObjectNotExist) {
 			return sSdk, fmt.Errorf("SDK not found in %q", channel)
 		}
+		urlError := &url.Error{}
```
dmitry-lyfar: It's not enough to test for url.Error, we should return "failed to connect" on the network errors:
```
	if urlErr, ok := err.(*url.Error); ok {
	    opErr, ok := urlErr.Err.(*net.OpError)
            if ok {
		// cannot connect
	     }
	}
```

lachypjones:
@dmitry-lyfar 
For the error message, do you prefer it with the network error included:
```
error: SDK store action failed:
- test: Cannot connect to store: "dial tcp 127.0.0.1:8080: connect: connection refused"
```
Or without:
```
error: SDK store action failed:
- test: Cannot connect to store
```

dmitry-lyfar:
the network error would be great to include, yes. also, lower case.

dmitry-lyfar (approved):

PR #199, opened 2024-10-16
Workshop hack

dmitry-lyfar:
# Description

## workshop hack

```
workshop hack [--drop|--restore] <WORKSHOP> [<HOOK>]
```
Opens a default editor for the workshop's hack SDK meta file or hook script. Runs partial refresh (i.e. `workshop refresh <WORKSHOP>/hack`) when editing is done, thus, installing the latest update for the hack SDK. 

`--drop` drops the hack SDK and refreshes to remove it from a workshop.

`--restore` restores the stored hack SDK and installs it into a workshop. Restore swaps the current version of the hack SDK with the stored if both existed. 

## hack SDK

The hack SDK embodies the idea of quickly implementing a few changes to a workshop, or, perhaps, sketching a new SDK altogether:

- Hack will never be carrying the data as opposed to a regular SDK.
- Hack can have hooks (setup-base, save-state, restore-state, check-health).
- Hack can use interfaces.
- “hack” is a reserved name.
- Installed last.

## Changes to other commands

- `workshop remove` will execute `workshop hack --drop` for all the removed workshops so that the hack SDK content can be restored if relaunched. The stored hack SDK must eventually be removed if unused within 90 days, however, this is planned as a separate PR where we address all the project life-cycle related issues (e.g. when Workshop decides that a project should not be tracked anymore and how this should affect stored hack SDK content).
- `workshop refresh` supports partial refresh for the hack SDK only (invoked as `workshop refresh <WORKSHOP>/hack`). It will be extended to support an arbitrary SDK with respect to an installation order in the following PRs.

## Breaking change
Workshops that were created with previous versions have to be recreated as this PR introduces sdk.Revision that will be associated with every installed SDK to track installed revisions. This change does not impose any limits on the number of revisions to be kept for a particular SDK. The plan is to introduce the limit in the following PRs together with refresh and revert that support any SDK (currently, partial refresh is available for the hack SDK only).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


cmd/workshop/hack.go
```diff
@@ -0,0 +1,228 @@
...
+	if c.drop {
+		storedir := sdk.WorkshopHackSdkStored(user.HomeDir, project.Id, workshop.Name)
+		reverter, err := dropHack(hackdir, storedir)
+		if err != nil {
+			return err
+		}
+		defer reverter.Fail()
+
+		cmdrefresh := &CmdRefresh{root: c.root}
+		if err = cmdrefresh.Run(cmd, []string{fmt.Sprintf("%s", av[0])}); err != nil {
```
jonathan-conder:
```suggestion
		if err = cmdrefresh.Run(cmd, []string{fmt.Sprintf("%s/hack", av[0])}); err != nil {
```

internal/sdk/revision_test.go
```diff
@@ -0,0 +1,196 @@
...
+		c.Assert(got, Equals, r)
+
+		got = sdk.Revision{}
+		err = json.Unmarshal([]byte(strconv.Itoa(r.N)), &got)
+		c.Assert(err, IsNil)
+		c.Assert(got, Equals, r)
+	}
+}
+
+func (s revisionSuite) ParseRevision(c *C) {
```
jonathan-conder:
```suggestion
func (s revisionSuite) TestParseRevision(c *C) {
```

internal/overlord/sdkstate/handlers.go
```diff
@@ -313,5 +327,9 @@ func (m *SdkManager) undoLinkSdk(task *state.Task, tomb *tomb.Tomb) error {
 		return err
 	}
 
-	return wp.UnlinkSdk(ctx, sdkSetup.Name)
+	if err = wp.UnlinkSdk(ctx, sdkSetup.Name); err != nil {
+		return err
+	}
+
+	return nil
```
jonathan-conder: Is there a reason to change this?

internal/interfaces/lxd_device/tests/integration/backend_test.go
```diff
@@ -323,3 +323,68 @@ exit 0
 	_, err = fs.Stat("/etc/profile.d/consumer-ssh-agent.sh")
 	c.Assert(osutil.IsDirNotExist(err), check.Equals, true)
 }
+
+func (f *backendDeviceSuite) TestRemoveSshAgent(c *check.C) {
```
jonathan-conder: This doesn't seem related, just checking you intended to include it here

internal/daemon/api_workshops_test.go
```diff
@@ -617,33 +618,48 @@ func (s *apiSuite) mockDoInstallSdk(c *check.C, ws string, sdks map[string]strin
...
-		c.Check(err, check.IsNil)
 
-		file, err := fs.OpenFile(filepath.Join(metadir, "sdk.yaml"), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
+		// Install meta.
+		metadir := filepath.Join(sdk.SdkRootPath(sdkname), "1", "meta")
+		err = fs.MkdirAll(metadir, 0755)
+		c.Check(err, check.IsNil)
+		err = fs.Symlink(filepath.Join(sdk.SdkRootPath(sdkname), "1"), filepath.Join(sdk.SdkRootPath(sdkname), "current"))
+		c.Check(err, check.IsNil)
+		file, err := fs.OpenFile(filepath.Join(metadir, "sdk.yaml"), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0744)
```
```suggestion
		file, err := fs.OpenFile(filepath.Join(metadir, "sdk.yaml"), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0644)
```
just to match what was there previously

jonathan-conder (changes requested):

cmd/workshop/hack.go
dmitry-lyfar: When `--drop` is used, the hack must be removed from the workshop completely. That means full refresh with no hack SDK installed.

jonathan-conder: Ah ok. In that case you can avoid `Sprintf` and just put `av[0:1]`

cmd/workshop/hack.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (c *CmdHack) Command() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "hack [--drop|--restore] <WORKSHOP> [hook-name]",
+		Args:  cobra.RangeArgs(1, 2),
+		Short: "Edit hack SDK",
+		RunE:  c.Run,
+	}
+
+	cmd.Flags().BoolVar(&c.drop, "drop", false, "Put hack SDK away and refresh.")
```
lachypjones: Conversation point, is 'drop' the best term for this?
It feels very similar to 'stash' and 'apply' in git. 

I don't have strong feelings either way.

lachypjones: Suggest changing "Put hack SDK away and refresh" to "Detatch hack SDK from workshop and refresh"
If so, change "Restore hack SDK and refresh" to "Attach hack SDK to workshop and refresh".

It's not clear from a first pass what "Put away" means, but this could be pedantic of me. Restore also potentially suggests that the _state_ of the hack is put back, not just the SDK definition

cmd/workshop/hack.go
```diff
@@ -0,0 +1,228 @@
...
+
+type CmdHack struct {
+	root    *CmdRoot
+	drop    bool
+	restore bool
+}
+
+func (c *CmdHack) Command() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "hack [--drop|--restore] <WORKSHOP> [hook-name]",
```
lachypjones: It would be nice to expand on the available hooks. 
That said it seems from the cobra docs that it's non-trivial to add help for arguments.

cmd/workshop/hack.go
```diff
@@ -0,0 +1,228 @@
...
+		res, err := runTextEditor("", []byte(boilerplate))
+		if err != nil {
+			return err
+		}
+
+		if err = writeSdkFile(sdkfile, res); err != nil {
+			return err
+		}
+	}
+
```
lachypjones, line 202: Suggest parsing the keys in the resulting yaml for validity?
Just flagging now as a future potential task, no impact for this PR

internal/overlord/sdkstate/handlers.go
```diff
@@ -144,31 +167,23 @@ func (m *SdkManager) doInstallSdk(task *state.Task, tomb *tomb.Tomb) error {
 	defer fl.Close()
 
 	target := filepath.Join("/root", filepath.Base(sdkSetup.Filename()))
-	sdkMount := workshop.Mount{Name: sdkSetup.Name, What: sdkSetup.Filename(), Where: target}
-	if err = m.backend.AddWorkshopMount(ctx, w, sdkMount); err != nil {
+	sdkmnt := workshop.Mount{Name: sdkSetup.Name, What: sdkSetup.Filename(), Where: target}
```
lachypjones: Personally I find `sdkmnt` harder to read/parse than `sdkMount`
It's a nit though and it's not a hill I'd die on. More than happy to not comment on variable name style in future. Only commenting because it was explicitly changed.

internal/overlord/sdkstate/handlers.go
```diff
@@ -144,31 +167,23 @@ func (m *SdkManager) doInstallSdk(task *state.Task, tomb *tomb.Tomb) error {
...
 		}
 	}
-
-	defer cleanup()
+	defer umount()
 
 	// example: /var/lib/workshop/sdk/cuda/712/
-	sdkPath := filepath.Join(dirs.WorkshopSdksDir, sdkSetup.Name,
-		strconv.Itoa(int(sdkSetup.Revision)))
+	sdkdir := filepath.Join(dirs.WorkshopSdksDir, sdkSetup.Name, sdkSetup.Revision.String())
```
lachypjones: As above, `sdkDir` might be nicer.

cmd/workshop/hack.go
```diff
@@ -0,0 +1,228 @@
...
+	if osutil.FileExists(sdkfile) {
+		old, err := os.ReadFile(sdkfile)
+		if err != nil {
+			return err
+		}
+
+		new, err := runTextEditor(sdkfile, []byte{})
```
lachypjones, line 184: Instead of always opening a text editor, it might be nicer to be able to redirect input directly into hack ie: `workshop hack <workshop> setup-hook < "apt update"`
or 
`workshop hack <workshop> setup-hook < mycomplexhook.yaml`

Then if an input isn't given we open an editor?

This would also neaten the tests up somewhat

lachypjones (approved):
No 'code-related' comments from me save for the variable name question. Again, happy with whatever there.
Remainder of the commentary can be addressed in future PRs if desired.

cmd/workshop/hack.go
dmitry-lyfar: @lachypjones we'll do that on the server when attempt to run a change with the updated hack SDK which will follow the editor closure.  We can potentially validate yaml here as well to be a bit more swift. In any way, I think it is beyond this PR for now.

cmd/workshop/hack.go
dmitry-lyfar: Really good point. We'll see how hack is being used, if requested, we'll support stdin. Will keep it at this level for now.

cmd/workshop/hack.go
dmitry-lyfar: As discussed, 'drop' will be discussed for options separately.

jonathan-conder (approved):

dmitry-lyfar:
@akcano this PR has implications for the documentation, but I'm not inviting you to review this one as you're working on the hack documentation separately. Please refer the PR description for better context on the scope of this change.

PR #203, opened 2024-10-18
Add Desktop interface

lachypjones:
# Description
The desktop interface provides a method for a workshop to access the Wayland and/or X11 sockets of the host, enabling GUI-based applications and development.

Additionally, the desktop interface can be paired with the GPU interface to enable hardware-accelerated graphical applications

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/explanation/interfaces/desktop-interface.rst
+++ b/docs/explanation/interfaces/desktop-interface.rst
@@ -1,0 +1,77 @@
+.. _exp_desktop_interface:
+
+Desktop interface
+
+The desktop interface
+provides access to the host system's Wayland socket
+from inside the workshop, allowing it to securely
+execute Wayland GUI applications.
+
+By using the interface,
+the SDK publisher allows the workshop to access the host's Wayland socket,
+which can be useful for various SDK-specific tasks such as
+building graphical applications or using editors without remote support.
+
+The interface isn't connected automatically at launch and refresh
+for security reasons.
+The :command:`workshop connect` and :command:`workshop disconnect` commands
+can be invoked manually after the workshop has started:
+
+.. code-block:: console
+
+   $ workshop connect ws/desktop-sdk:desktop
+   $ workshop disconnect ws/desktop-sdk:desktop
+
+
+Establishing a connection means
+a proxy Unix domain socket has been created
+and the following environment variables have been set:
+
+- :envvar:`WAYLAND_DISPLAY`
+  Identifies the name of the Wayland socket
+- :envvar:`XDG_SESSION_TYPE`
+  Specifies the current display server type
+- :envvar:`QT_QPA_PLATFORM`
+  Specifies the Qt platform plugin to be used for Qt-based applications
+- :envvar:`ELECTRON_OZONE_PLATFORM_HINT`
+  Specifies the preferred platform for electron applications
+
+To check if the interface is connected:
+
+.. code-block:: console
+
+   $ workshop connections --all
+
+     Interface  Plug                   Slot       Notes
+     ...
+     desktop    ws/desktop-sdk:desktop :desktop   manual
+
+This means the host's Wayland socket is available inside the workshop
+
+.. code-block:: console
+
+   $ workshop shell ws
+   workshop@ws-8584e571$ ls $XDG_RUNTIME_DIR | grep wayland
+
+     wayland-1
+
+
+See also
+--------
+
+Explanation:
+
+- :ref:`exp_interfaces`
+- :ref:`exp_plugs_slots`
+- :ref:`exp_sdk_definition`
+- :ref:`exp_workshop_def`
+
+
+Reference:
+
+- :ref:`ref_workshop_connect`
+- :ref:`ref_workshop_connections`
+- :ref:`ref_workshop_disconnect`
+- :ref:`ref_workshop_launch`
+- :ref:`ref_workshop_refresh`
+- :ref:`ref_workshop_shell`
--- a/docs/explanation/interfaces/desktop-interface.rst
+++ b/docs/explanation/interfaces/desktop-interface.rst
@@ -4,13 +4,13 @@ Desktop interface
 
 The desktop interface
 provides access to the host system's Wayland socket
-from inside the workshop, allowing it to securely
-execute Wayland GUI applications.
+from inside the workshop,
+allowing it to securely run Wayland GUI applications.
 
 By using the interface,
 the SDK publisher allows the workshop to access the host's Wayland socket,
-which can be useful for various SDK-specific tasks such as
-building graphical applications or using editors without remote support.
+which can be useful for various SDK-specific tasks
+such as building graphical applications or using editors without remote support.
 
 The interface isn't connected automatically at launch and refresh
 for security reasons.
@@ -27,14 +27,18 @@ Establishing a connection means
 a proxy Unix domain socket has been created
 and the following environment variables have been set:
 
-- :envvar:`WAYLAND_DISPLAY`
-  Identifies the name of the Wayland socket
-- :envvar:`XDG_SESSION_TYPE`
-  Specifies the current display server type
-- :envvar:`QT_QPA_PLATFORM`
-  Specifies the Qt platform plugin to be used for Qt-based applications
-- :envvar:`ELECTRON_OZONE_PLATFORM_HINT`
-  Specifies the preferred platform for electron applications
+- :envvar:`$WAYLAND_DISPLAY`
+  Sets the Wayland socket name
+  
+- :envvar:`$XDG_SESSION_TYPE`
+  Sets the current display server type
+
+- :envvar:`$QT_QPA_PLATFORM`
+  Sets the Qt platform plugin to be used for Qt applications
+
+- :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
+  Sets the preferred platform for Electron applications
+
 
 To check if the interface is connected:
 
@@ -46,7 +50,8 @@ To check if the interface is connected:
      ...
      desktop    ws/desktop-sdk:desktop :desktop   manual
 
-This means the host's Wayland socket is available inside the workshop
+
+This means the host's Wayland socket is available inside the workshop:
 
 .. code-block:: console
 
--- a/docs/explanation/interfaces/desktop-interface.rst
+++ b/docs/explanation/interfaces/desktop-interface.rst
@@ -70,7 +70,7 @@ To check if the interface is connected:
      desktop    ws/desktop-sdk:desktop :desktop   manual
 
 
-This means the host's Wayland/X11 socket is available inside the workshop:
+This means the host's display socket (Wayland, X11 or both) is available inside the workshop:
 
 .. code-block:: console
 
--- a/docs/explanation/interfaces/desktop-interface.rst
+++ b/docs/explanation/interfaces/desktop-interface.rst
@@ -1,14 +1,15 @@
 .. _exp_desktop_interface:
 
 Desktop interface
+=================
 
 The desktop interface
-provides access to the host system's Wayland socket
+provides access to the host system's display (Wayland/X11) socket(s)
 from inside the workshop,
-allowing it to securely run Wayland GUI applications.
+allowing it to securely run GUI applications.
 
 By using the interface,
-the SDK publisher allows the workshop to access the host's Wayland socket,
+the SDK publisher allows the workshop to utilise the host's display
 which can be useful for various SDK-specific tasks
 such as building graphical applications or using editors without remote support.
 
@@ -27,17 +28,35 @@ Establishing a connection means
 a proxy Unix domain socket has been created
 and the following environment variables have been set:
 
-- :envvar:`$WAYLAND_DISPLAY`
-  Sets the Wayland socket name
-  
-- :envvar:`$XDG_SESSION_TYPE`
-  Sets the current display server type
+.. list-table::
+   :header-rows: 1
+   :width: 95
+   :widths: 30 30 30
 
-- :envvar:`$QT_QPA_PLATFORM`
-  Sets the Qt platform plugin to be used for Qt applications
+   * - Wayland
+     - X11
+     - Both
 
-- :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
-  Sets the preferred platform for Electron applications
+   * - | :envvar:`$WAYLAND_DISPLAY`
+       | :envvar:`$XDG_SESSION_TYPE`
+       | :envvar:`$QT_QPA_PLATFORM`
+       | :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
+       | :envvar:`$XDG_BACKEND`
+     - | :envvar:`$DISPLAY`
+       | :envvar:`$XDG_SESSION_TYPE`
+       | :envvar:`$QT_QPA_PLATFORM`
+       | :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
+       | :envvar:`$XDG_BACKEND`
+       | :envvar:`$XAUTHORITY`:sup:`*`
+     - | :envvar:`$WAYLAND_DISPLAY`
+       | :envvar:`$DISPLAY`
+       | :envvar:`$XDG_SESSION_TYPE`
+       | :envvar:`$QT_QPA_PLATFORM`
+       | :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
+       | :envvar:`$XDG_BACKEND`
+       | :envvar:`$XAUTHORITY`:sup:`*`
+
+*\*only set if present on the host*
 
 
 To check if the interface is connected:
@@ -51,7 +70,7 @@ To check if the interface is connected:
      desktop    ws/desktop-sdk:desktop :desktop   manual
 
 
-This means the host's Wayland socket is available inside the workshop:
+This means the host's Wayland/X11 socket is available inside the workshop:
 
 .. code-block:: console
 
@@ -60,6 +79,12 @@ This means the host's Wayland socket is available inside the workshop:
 
      wayland-1
 
+.. code-block:: console
+
+   $ workshop shell ws
+   workshop@ws-8584e571$ ls /tmp/.X11-unix
+
+     X0
 
 See also
 --------
--- a/docs/explanation/interfaces/index.rst
+++ b/docs/explanation/interfaces/index.rst
@@ -51,8 +51,9 @@ This behaviour varies by interface,
 but the overall aim is to conduct reasonably in each case:
 the :ref:`mount <exp_mount_interface>`
 and the :ref:`GPU <exp_gpu_interface>` interfaces are auto-connected,
-whereas the :ref:`camera <exp_camera_interface>`
-and :ref:`SSH <exp_ssh_interface>` interfaces require manual connection.
+whereas the :ref:`camera <exp_camera_interface>`,
+:ref:`desktop <exp_desktop_interface>` and ref:`SSH <exp_ssh_interface>`
+interfaces require manual connection.
 
 
 See also
--- a/docs/explanation/interfaces/index.rst
+++ b/docs/explanation/interfaces/index.rst
@@ -52,7 +52,7 @@ but the overall aim is to conduct reasonably in each case:
 the :ref:`mount <exp_mount_interface>`
 and the :ref:`GPU <exp_gpu_interface>` interfaces are auto-connected,
 whereas the :ref:`camera <exp_camera_interface>`,
-:ref:`desktop <exp_desktop_interface>` and ref:`SSH <exp_ssh_interface>`
+:ref:`desktop <exp_desktop_interface>` and :ref:`SSH <exp_ssh_interface>`
 interfaces require manual connection.
 
 
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -76,19 +76,14 @@ func (iface *desktopInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInf
 }
 
 func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	usr, err := workshop.LookupUsername(spec.User())
-	if err != nil {
-		return err
-	}
-
-	env, err := systemd.UserEnvironment(usr.Username)
+	env, err := systemd.UserEnvironment(spec.User)
 	if err != nil {
 		return err
 	}
 
 	xdg := env["XDG_RUNTIME_DIR"]
 	if xdg == "" {
-		return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", spec.User())
+		return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", spec.User.Username)
 	}
 
 	desktop := workshop.Desktop{}
@@ -97,7 +92,7 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 	display := env["DISPLAY"]
 
 	if wayland == "" && display == "" {
-		return fmt.Errorf("neither DISPLAY nor WAYLAND_DISPLAY are set for user %q", spec.User())
+		return fmt.Errorf("neither DISPLAY nor WAYLAND_DISPLAY are set for user %q", spec.User.Username)
 	}
 
 	if wayland != "" {
@@ -119,7 +114,7 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 		x.Listen = x.Connect
 	}
 
-	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, usr.Uid, ".Xauthority")
+	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, ".Xauthority")
 	xauth := env["XAUTHORITY"]
 	if xauth != "" {
 		m := workshop.Mount{}
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -55,7 +55,7 @@ const desktopDeclarationPlugs = `
       plug-names:
         - $INTERFACE
     allow-connection: true
-    allow-auto-connection: false
+    deny-auto-connection: true
 `
 
 type desktopInterface struct{}
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -1,0 +1,141 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"fmt"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/canonical/workshop/internal/dirs"
+	"github.com/canonical/workshop/internal/interfaces"
+	"github.com/canonical/workshop/internal/interfaces/lxd_device"
+	"github.com/canonical/workshop/internal/osutil"
+	"github.com/canonical/workshop/internal/sdk"
+	"github.com/canonical/workshop/internal/workshop"
+)
+
+const desktopSummary = `allows SDKs to use the host's wayland compositor`
+
+const desktopBaseDeclarationSlots = `
+  desktop:
+    allow-installation:
+      slot-sdk-type:
+        - system
+      slot-names:
+        - $INTERFACE
+    allow-connection: true
+    deny-auto-connection: true
+`
+
+const desktopDeclarationPlugs = `
+  desktop:
+    allow-installation:
+      plug-sdk-type:
+        - regular
+      plug-names:
+        - $INTERFACE
+    allow-connection: true
+    allow-auto-connection: false
+`
+
+type desktopInterface struct{}
+
+func (iface *desktopInterface) Name() string {
+	return "desktop"
+}
+
+func (iface *desktopInterface) StaticInfo() interfaces.StaticInfo {
+	return interfaces.StaticInfo{
+		Summary:              desktopSummary,
+		BaseDeclarationPlugs: desktopDeclarationPlugs,
+		BaseDeclarationSlots: desktopBaseDeclarationSlots,
+		AffectsPlugOnRefresh: true,
+	}
+}
+
+// Why do we return autoconnect=true when allow-auto-connection is false?
+// Is this a case of two things named similary?
+// ie. allow is policy and the below is technical?
+func (iface *desktopInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInfo) bool {
+	return true
+}
+
+func (iface *desktopInterface) MountConnectedPlug(
+	spec *lxd_device.Specification,
+	plug *interfaces.ConnectedPlug,
+	slot *interfaces.ConnectedSlot,
+) error {
+
+	user, err := workshop.LookupUsername(spec.User())
+	if err != nil {
+		return err
+	}
+
+	uid, _, err := osutil.UidGid(user)
+	if err != nil {
+		return err
+	}
+
+	// systemd is responsible for generating the "WAYLAND_DISPLAY" environment variable
+	// because of this we must parse the environment as it's set by systemd
+	cmd := exec.Command("sudo", "-E", "-u", spec.User(), "systemctl", "--user", "show-environment")
+	// XDG_RUNTIME_DIR may not be set if a command invoked by sudo or
+	// systemd-run; set it here to the default location. It is required for the
+	// systemctl to work with --user. See:
+	// https://unix.stackexchange.com/questions/346841/why-does-sudo-i-not-set-xdg-runtime-dir-for-the-target-user
+	defaultXdg := filepath.Join(dirs.XdgRuntimeDirBase, strconv.FormatUint(uint64(uid), 10))
+	cmd.Env = append(cmd.Env, "XDG_RUNTIME_DIR="+defaultXdg)
+	out, errOut, err := osutil.RunCmd(cmd)
+	if err != nil {
+		return fmt.Errorf(string(errOut))
+	}
+
+	rawEnv := strings.FieldsFunc(string(out), func(r rune) bool { return r == '\n' })
+	env, err := osutil.ParseEnvironment(rawEnv)
+	if err != nil {
+		return err
+	}
+
+	wayland, ok := env["WAYLAND_DISPLAY"]
+	if !ok || wayland == "" {
+		return fmt.Errorf("WAYLAND_DISPLAY is either empty or unset for user %q. Is this a wayland session?", user.Username)
+	}
+
+	xdg, ok := env["XDG_RUNTIME_DIR"]
+	if !ok || xdg == "" {
+		return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", user.Username)
+	}
+
+	name := plug.Sdk().Name + "-" + plug.Name()
+
+	fromSocket := xdg + "/" + wayland
+	// The container XDG_RUNTIME_DIR is always /run/user/1000 for the workshop user
+	// Use the same WAYLAND_DISPLAY identifier as the host
+	toSocket := "/run/user/1000/" + wayland
+
+	return spec.SetDesktop(workshop.Desktop{Name: name, Connect: fromSocket, Listen: toSocket})
+}
+
+func init() {
+	registerIface(&desktopInterface{})
+}
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -111,7 +111,7 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 		x := &desktop.X11
 		x.Name = plug.Sdk().Name + "-" + "x11"
 		x.Connect = filepath.Join("/tmp/.X11-unix", "X"+strings.TrimPrefix(display, ":"))
-		x.Listen = x.Connect
+		x.Listen = filepath.Join(dirs.WorkshopRunDir, "X"+strings.TrimPrefix(display, ":"))
 	}
 
 	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, "Xauthority")
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -112,6 +112,9 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 		desktop.X11.Listen = desktop.X11.Connect
 	}
 
+	// We mount the Xauthority inside a parent folder to ensure that the mounted
+	// cookie is updated when the host cookie changes (ie. reboot).
+	// https://discuss.linuxcontainers.org/t/mount-single-file/17975
 	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, "Xauthority")
 	xauth := env["XAUTHORITY"]
 	if xauth != "" {
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -96,22 +96,20 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 	}
 
 	if wayland != "" {
-		// Add wayland to the profile string
-		w := &desktop.Wayland
-		w.Name = plug.Sdk().Name + "-" + "wayland"
-		w.Connect = filepath.Join(xdg, wayland)
-		w.Listen = filepath.Join("/run/user/1000/", wayland)
+		desktop.Wayland = &workshop.ProxyEntry{}
+		desktop.Wayland.Name = plug.Sdk().Name + "-" + "wayland"
+		desktop.Wayland.Connect = filepath.Join(xdg, wayland)
+		desktop.Wayland.Listen = filepath.Join("/run/user/1000/", wayland)
 	}
 
 	// We pass through the X11 socket regardless of whether XAUTHORITY is present
 	// on the host. This then gives users the option to modify their xhost
 	// settings to allow connections from the container and container user.
 	if display != "" {
-		// Add X11 to the profile string
-		x := &desktop.X11
-		x.Name = plug.Sdk().Name + "-" + "x11"
-		x.Connect = filepath.Join("/tmp/.X11-unix", "X"+strings.TrimPrefix(display, ":"))
-		x.Listen = x.Connect
+		desktop.X11 = &workshop.ProxyEntry{}
+		desktop.X11.Name = plug.Sdk().Name + "-" + "x11"
+		desktop.X11.Connect = filepath.Join("/tmp/.X11-unix", "X"+strings.TrimPrefix(display, ":"))
+		desktop.X11.Listen = desktop.X11.Connect
 	}
 
 	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, "Xauthority")
@@ -125,7 +123,7 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 		spec.AddMountEntry(m)
 	}
 
-	return spec.SetDesktop(&desktop)
+	return spec.SetDesktop(desktop)
 }
 
 func init() {
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -114,14 +114,14 @@ func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification
 		x.Listen = x.Connect
 	}
 
-	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, ".Xauthority")
+	workshopdXauth := filepath.Join(dirs.WorkshopdRunDir, spec.User.Uid, "Xauthority")
 	xauth := env["XAUTHORITY"]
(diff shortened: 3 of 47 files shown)


docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,77 @@
+.. _exp_desktop_interface:
+
+Desktop interface
+
+The desktop interface
+provides access to the host system's Wayland socket
+from inside the workshop, allowing it to securely
+execute Wayland GUI applications.
```
akcano:
```suggestion
from inside the workshop,
allowing it to securely run Wayland GUI applications.
```

akcano: https://sembr.org

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,77 @@
...
+
+By using the interface,
+the SDK publisher allows the workshop to access the host's Wayland socket,
+which can be useful for various SDK-specific tasks such as
+building graphical applications or using editors without remote support.
```
akcano:
```suggestion
which can be useful for various SDK-specific tasks
such as building graphical applications or using editors without remote support.
```

akcano: same

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,77 @@
...
+- :envvar:`WAYLAND_DISPLAY`
+  Identifies the name of the Wayland socket
+- :envvar:`XDG_SESSION_TYPE`
+  Specifies the current display server type
+- :envvar:`QT_QPA_PLATFORM`
+  Specifies the Qt platform plugin to be used for Qt-based applications
+- :envvar:`ELECTRON_OZONE_PLATFORM_HINT`
+  Specifies the preferred platform for electron applications
+
+To check if the interface is connected:
```
akcano:
```suggestion
- :envvar:`$WAYLAND_DISPLAY`
  Sets the Wayland socket name
  
- :envvar:`$XDG_SESSION_TYPE`
  Sets the current display server type

- :envvar:`$QT_QPA_PLATFORM`
  Sets the Qt platform plugin to be used for Qt applications

- :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
  Sets the preferred platform for Electron applications


To check if the interface is connected:
```

akcano: envvar visibility, formatting convention, wording preferences

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,77 @@
...
+
+.. code-block:: console
+
+   $ workshop connections --all
+
+     Interface  Plug                   Slot       Notes
+     ...
+     desktop    ws/desktop-sdk:desktop :desktop   manual
+
+This means the host's Wayland socket is available inside the workshop
```
akcano:
```suggestion
     desktop    ws/desktop-sdk:desktop :desktop   manual


This means the host's Wayland socket is available inside the workshop:
```

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,141 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
```
akcano: is the year ok?

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+
+	rawEnv := strings.FieldsFunc(string(out), func(r rune) bool { return r == '\n' })
+	env, err := osutil.ParseEnvironment(rawEnv)
+	if err != nil {
+		return err
+	}
+
+	wayland, ok := env["WAYLAND_DISPLAY"]
+	if !ok || wayland == "" {
+		return fmt.Errorf("WAYLAND_DISPLAY is either empty or unset for user %q. Is this a wayland session?", user.Username)
```
akcano:
```suggestion
		return fmt.Errorf("WAYLAND_DISPLAY is either empty or unset for user %q. Is this a Wayland session?", user.Username)
```

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +226,34 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	envVars := []string{
+		"WAYLAND_DISPLAY=" + strings.TrimPrefix(dev.Listen, "/run/user/1000/"),
+		"QT_QPA_PLATFORM=wayland-egl",
+		"XDG_SESSION_TYPE=wayland",
+		"ELECTRON_OZONE_PLATFORM_HINT=auto",
+	}
+
+	env, err := fs.Create(filepath.Join("/etc/profile.d", dev.Name+".sh"))
+	if err != nil {
+		return fmt.Errorf("failed to setup the environment for desktop in profile.d for %q: %w", workshop, err)
```
akcano:
```suggestion
		return fmt.Errorf("failed to set up the desktop environment in profile.d for %q: %w", workshop, err)
```

akcano: verb form is mandatory, desktop change is optional

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +226,34 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	}
+	defer env.Close()
+
+	for _, envVar := range envVars {
+		_, err = env.WriteString("export " + envVar + "\n")
+		if err != nil {
+			return fmt.Errorf("failed to setup the environment for desktop in profile.d for %q: %w", workshop, err)
```
akcano: same as above

internal/interfaces/lxd_device/spec.go
```diff
@@ -92,16 +92,39 @@ func (s *Specification) AddMountEntry(dev workshop.Mount) error {
 	return nil
 }
 
+// Ssh Agent and Desktop are both of the lxc 'proxy' type
+// These are network protocol proxy devices and open a port on the host or in a workhop.
```
akcano:
```suggestion
// These are network protocol proxy devices that open a port on the host or in a workhop.
```

internal/workshop/lxd/lxd_backend_profile.go
```diff
@@ -48,12 +48,15 @@ func lxdToSdkProfile(profile string, devs map[string]map[string]string, config m
 			pr.Gpu = &workshop.Gpu{Name: name}
 		case "proxy":
 			devtype := config[DeviceTypeConfigKey(profile, name)]
-			if devtype == "ssh-agent" {
+
+			switch devtype {
```
akcano, line 51: maybe it's time for a map here

akcano (changes requested):
LGTM with minor fixes.

docs/explanation/interfaces/desktop-interface.rst
lachypjones: Good link, thankyou. I'd been trying to determine what the threshold for line breaks were.

internal/workshop/lxd/lxd_backend_profile.go
lachypjones: Originally that's what I had refactored it to be, see:
https://github.com/canonical/workshop/pull/196

Dmitry suggested that it was best to refrain from making them generic (required to use a map where you suggested) as we don't yet know what else 'proxy' type devices may be used for.

internal/interfaces/lxd_device/backend.go
lachypjones: I agree the language was confusing.
The issue is that 'desktop' is the name of the interface and 'desktop environment' has it's own specific meaning. Suggest instead making the error:
`failed to configure the environment for interface: desktop in workshop: <workshop name> (<error>)`
Thoughts?
We then change the error string in other interfaces that configure the environment to match this format?

akcano: fair enough, let's go with your suggestion for now

internal/workshop/lxd/lxd_backend_profile.go
akcano: ok, the refactoring threshold is up to you guys

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return err
+	}
+
+	uid, _, err := osutil.UidGid(user)
+	if err != nil {
+		return err
+	}
+
+	// systemd is responsible for generating the "WAYLAND_DISPLAY" environment variable
```
dmitry-lyfar: The format of the code comments that I'm trying to get at: starts with a capital letter, plain English, sentences are separated with dots. Comments should be hard-wrapped with a width length 80 (for vscode, see the plugin named "Rewrap").

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+func (iface *desktopInterface) StaticInfo() interfaces.StaticInfo {
+	return interfaces.StaticInfo{
+		Summary:              desktopSummary,
+		BaseDeclarationPlugs: desktopDeclarationPlugs,
+		BaseDeclarationSlots: desktopBaseDeclarationSlots,
+		AffectsPlugOnRefresh: true,
+	}
+}
+
+// Why do we return autoconnect=true when allow-auto-connection is false?
```
dmitry-lyfar: These questions should not be here. It can go to the daily agenda for discussion.

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+
+	xdg, ok := env["XDG_RUNTIME_DIR"]
+	if !ok || xdg == "" {
+		return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", user.Username)
+	}
+
+	name := plug.Sdk().Name + "-" + plug.Name()
+
+	fromSocket := xdg + "/" + wayland
+	// The container XDG_RUNTIME_DIR is always /run/user/1000 for the workshop user
```
dmitry-lyfar: Same note about comments applies here.

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +226,34 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	envVars := []string{
+		"WAYLAND_DISPLAY=" + strings.TrimPrefix(dev.Listen, "/run/user/1000/"),
+		"QT_QPA_PLATFORM=wayland-egl",
+		"XDG_SESSION_TYPE=wayland",
+		"ELECTRON_OZONE_PLATFORM_HINT=auto",
+	}
+
+	env, err := fs.Create(filepath.Join("/etc/profile.d", dev.Name+".sh"))
+	if err != nil {
+		return fmt.Errorf("failed to configure the environment for interface: desktop in workshop: %q (%w)", workshop, err)
```
dmitry-lyfar: The error messaging in a similar ssh-agent interface for the env. setup failure starts with "cannot". So it should either be "failed" there or "cannot" here for the similar errors.

Also, that error message does not include the interface name, just the workshop name and an error string.

Let's bring it all to consistency.

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +226,34 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	}
+	defer env.Close()
+
+	for _, envVar := range envVars {
+		_, err = env.WriteString("export " + envVar + "\n")
+		if err != nil {
+			return fmt.Errorf("failed to configure the environment for interface: desktop in workshop: %q (%w)", workshop, err)
```
dmitry-lyfar: Same note about error messaging applies here.

tests/main/interface-desktop/task.yaml
```diff
@@ -0,0 +1,37 @@
+summary: Check desktop-agent interface scenarios
```
dmitry-lyfar: desktop-agent?

tests/main/interface-desktop/task.yaml
```diff
@@ -0,0 +1,37 @@
...
+  workshop_exec launch ws-desktop-fail > launch.log || true
+  MATCH ".*installation not allowed by \"user-desktop\" slot rule of interface \"desktop\"" < launch.log
+
+  echo "Connect the desktop interface"
+  sudo -u ubuntu /bin/bash -c 'XDG_RUNTIME_DIR=/run/user/1000 systemctl --user set-environment WAYLAND_DISPLAY=wayland-1'
+  sudo -u ubuntu /bin/bash -c 'workshop connect ws-desktop/test-sdk-desktop:desktop :desktop'
+
+  echo "Check that the expected environment is configured inside the workshop"
+  workshop_exec exec ws-desktop -- sudo -H -i -u workshop \
+    echo '$WAYLAND_DISPLAY, $QT_QPA_PLATFORM, $XDG_SESSION_TYPE, $ELECTRON_OZONE_PLATFORM_HINT' | \
```
dmitry-lyfar: `test -v` should be used for checking if a variable is set / unset.

dmitry-lyfar (changes requested):
It looks good, see the comments. Also, commit history: use -> Use.

internal/interfaces/lxd_device/backend.go
lachypjones: @dmitry-lyfar 
Thoughts?
I can change ssh-agent to match or revert to what I had originally (which matched ssh-agent)

dmitry-lyfar: @lachypjones I don't think we refer things like "interface: desktop" elsewhere. If you refer a workshop, that would be _"name" workshop_, see other error or log messages. Hence, _"desktop" interface_. For example,

```
return statusBadRequest("remount requires a content interface plug (provided plug is of %q interface)", conn.Plug.Interface())
```

So, I would prefer the desktop interface to comply with the rest.

dmitry-lyfar:
As per the internal discussion: support XWAYLAND too. The test apps:

```
sudo apt-get update
sudo apt install -y kitty
sudo apt install -y mesa-utils
sudo apt install -y x11-apps
```

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,82 @@
...
+allowing it to securely run Wayland GUI applications.
+
+By using the interface,
+the SDK publisher allows the workshop to access the host's Wayland socket,
+which can be useful for various SDK-specific tasks
+such as building graphical applications or using editors without remote support.
+
+The interface isn't connected automatically at launch and refresh
+for security reasons.
+The :command:`workshop connect` and :command:`workshop disconnect` commands
```
akcano, line 18:
```suggestion
The :command:`workshop connect` and :command:`workshop disconnect` commands
```

akcano: Non-breaking spaces.

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,82 @@
...
+  Sets the Wayland socket name
+  
+- :envvar:`$XDG_SESSION_TYPE`
+  Sets the current display server type
+
+- :envvar:`$QT_QPA_PLATFORM`
+  Sets the Qt platform plugin to be used for Qt applications
+
+- :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`
+  Sets the preferred platform for Electron applications
```
akcano:
```suggestion
- :envvar:`$WAYLAND_DISPLAY`:
  Wayland socket name
  
- :envvar:`$XDG_SESSION_TYPE`:
  Current display server type

- :envvar:`$QT_QPA_PLATFORM`:
  Qt platform plugin to be used for Qt applications

- :envvar:`$ELECTRON_OZONE_PLATFORM_HINT`:
  Preferred platform for Electron applications
```

akcano (changes requested):
A few more improvements.

internal/systemd/user.go
```diff
@@ -0,0 +1,42 @@
...
+	"os/user"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/canonical/workshop/internal/dirs"
+	"github.com/canonical/workshop/internal/osutil"
+)
+
+func UserEnvironment(user *user.User) (map[string]string, error) {
```
dmitry-lyfar, line 16: This should accept a user name, see how it is done in a similar function called `func NormalizeUidGid(uid, gid *int, username, group string) (*int, *int, error)`

internal/systemd/user.go
```diff
@@ -0,0 +1,42 @@
...
+	// https://unix.stackexchange.com/questions/346841/why-does-sudo-i-not-set-xdg-runtime-dir-for-the-target-user
+	defaultXdg := filepath.Join(dirs.XdgRuntimeDirBase, strconv.FormatUint(uint64(uid), 10))
+	cmd.Env = append(cmd.Env, "XDG_RUNTIME_DIR="+defaultXdg)
+	out, errOut, err := osutil.RunCmd(cmd)
+	if err != nil {
+		return nil, fmt.Errorf(string(errOut))
+	}
+
+	rawEnv := strings.FieldsFunc(string(out), func(r rune) bool { return r == '\n' })
+	env, err := osutil.ParseEnvironment(rawEnv)
```
dmitry-lyfar: `return osutil.ParseEnvironment(rawEnv)`

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,143 @@
...
+		BaseDeclarationSlots: desktopBaseDeclarationSlots,
+		AffectsPlugOnRefresh: true,
+	}
+}
+
+func (iface *desktopInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInfo) bool {
+	return true
+}
+
+func (iface *desktopInterface) MountConnectedPlug(
```
dmitry-lyfar: This reads well. But we don't use this formatting and go full length in other places, so I suggest to leave it formatted as elsewhere (no new lines).

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if wayland != "" {
+		// Add wayland to the profile string
+		w := &desktop.Wayland
+		w.Name = plug.Sdk().Name + "-" + "wayland"
+		w.Connect = filepath.Join(xdg, wayland)
+		w.Listen = filepath.Join("/run/user/1000/", wayland)
+	}
+
+	// We pass through the X11 socket regardless of whether XAUTHORITY is present
```
dmitry-lyfar, line 105: This comment is not formatted correctly:

```
	// We pass through the X11 socket regardless of whether XAUTHORITY is
	// present on the host. This then gives users the option to modify their
	// xhost settings to allow connections from the container and container
	// user.
```

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,143 @@
...
+	// connections from the container and container user.
+	if display != "" {
+		// Add X11 to the profile string
+		x := &desktop.X11
+		x.Name = plug.Sdk().Name + "-" + "x11"
+		x.Connect = filepath.Join("/tmp/.X11-unix", "X"+strings.TrimPrefix(display, ":"))
+		x.Listen = x.Connect
+	}
+
+	if wayland == "" && display == "" {
```
dmitry-lyfar: This can move up to the point right after `env` is read for those variables.

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +228,75 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	if dev.X11.Name != "" {
+		envVars["DISPLAY"] = ":" + strings.TrimPrefix(dev.X11.Listen, "/tmp/.X11-unix/X")
+	}
+
+	xauth := env["XAUTHORITY"]
+	if xauth != "" {
+		envVars["XAUTHORITY"] = filepath.Join(dirs.WorkshopRunDir, ".Xauthority")
+		err := wsutil.CopyXauthority(usr)
+		if err != nil {
+			logger.Noticef("cannot copy Xauthority file for user %s, X11 applications may not work, %v", user, err)
```
dmitry-lyfar: `user` is a pointer that is attempted to be printed as `%s`

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
+package wsutil
```
dmitry-lyfar: wsutil is `web socket util`. I don't think it aligns with x11.go well. It should be a separate package, it's ok to be a small package.

internal/interfaces/lxd_device/backend.go
```diff
@@ -226,6 +228,75 @@ func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
...
+	}
+
+	if dev.X11.Name != "" {
+		envVars["DISPLAY"] = ":" + strings.TrimPrefix(dev.X11.Listen, "/tmp/.X11-unix/X")
+	}
+
+	xauth := env["XAUTHORITY"]
+	if xauth != "" {
+		envVars["XAUTHORITY"] = filepath.Join(dirs.WorkshopRunDir, ".Xauthority")
+		err := wsutil.CopyXauthority(usr)
```
dmitry-lyfar: I think it makes sense to explain why it is being copied on every install call here.

dmitry-lyfar:
```suggestion
		if err := wsutil.CopyXauthority(usr); err != nil {
```

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	xauth, _ := env["XAUTHORITY"]
+	if xauth == "" {
+		return err
+	}
+
+	destDir := filepath.Join(dirs.WorkshopdRunDir, usr.Uid)
+	if !osutil.IsDir(destDir) {
+		os.MkdirAll(destDir, 0744)
```
dmitry-lyfar: error handling? Also, 755?

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return err
+	}
+
+	env, err := systemd.UserEnvironment(usr)
+	if err != nil {
+		return err
+	}
+
+	xauth, _ := env["XAUTHORITY"]
```
dmitry-lyfar:
```suggestion
	xauth := env["XAUTHORITY"]
```

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
...
+	//
+	// There's no advantage to comparing the files, destroy and copy every
+	// time
+	destFile := filepath.Join(destDir, ".Xauthority")
+	os.RemoveAll(destFile)
+
+	// Make sure the user owns the file.
+	// Important to make sure that the user cannot steal another user's
+	// Xauthority file
+	// See migrateXauthority function inside of snapd for more info
```
dmitry-lyfar: I would copy/adapt the comment for the function here to not force the reader to go and look into the snapd's code base.

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
...
+	// Make sure the target file doesn't exist as a directory.
+	// We do this because if the file doesn't exist and the desktop interface is
+	// mounted, workshopd will create an empty directory as part of the mount
+	// interface.
+	// This then breaks any 'normal' file operations
+	//
+	// There's no advantage to comparing the files, destroy and copy every
+	// time
+	destFile := filepath.Join(destDir, ".Xauthority")
+	os.RemoveAll(destFile)
```
dmitry-lyfar: error handling?

dmitry-lyfar (changes requested):

internal/wsutil/x11.go
lachypjones: Ah that makes sense. Apologies I just instinctively read it as 'workshoputil'

tests/main/interface-desktop/task.yaml
```diff
@@ -0,0 +1,89 @@
...
+      [ -z "$WAYLAND_DISPLAY" ]
+      [ -z "$QT_QPA_PLATFORM" ]
+      [ "$XDG_SESSION_TYPE" == "unspecified" ]
+      [ -z "$ELECTRON_OZONE_PLATFORM_HINT" ]
+  EOF
+
+  echo "Connect the desktop interface for an x11 session"
+  sudo -u ubuntu /bin/bash -c 'touch /tmp/.Xauthority'
+  sudo -u ubuntu /bin/bash -c 'XDG_RUNTIME_DIR=/run/user/1000 systemctl --user set-environment DISPLAY=:0 XAUTHORITY=/tmp/.Xauthority'
+  sudo -u ubuntu /bin/bash -c 'workshop connect ws-desktop/test-sdk-desktop:desktop :desktop'
```
dmitry-lyfar: interesting, why workshop_exec above and sudo -u here?

internal/wsutil/x11.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/canonical/workshop/internal/dirs"
+	"github.com/canonical/workshop/internal/osutil"
+	"github.com/canonical/workshop/internal/systemd"
+	"github.com/canonical/workshop/internal/workshop"
+)
+
+// Copies the user's $XAUTHORITY file to the Workshop run directory.
+// This is used by interfaces that require an X11 socket.
+func CopyXauthority(user string) error {
```
dmitry-lyfar: Tests for CopyXAuthority?

tests/main/interface-desktop/task.yaml
lachypjones: These are on the host, not in the workshop.

dmitry-lyfar:
I mean workshop connect call.

Also, bash -c doesn’t seem to be needed there?

internal/systemd/user.go
lachypjones: The reason I made the function accept the user.User struct was because on 3/4 of occasions where the UserEnvironment function is used, we also require the struct elsewhere in the same scope.

What do you think? (happy with either, I just wanted to explain my rationale)

internal/x11/x11.go
```diff
@@ -9,30 +10,17 @@ import (
...
+// Copies the user's $XAUTHORITY file to the Workshopd run directory.
+func MigrateXauthority(user *user.User, xauth string) (err error) {
 	if xauth == "" {
-		return err
+		return errors.New("xauth cannot be empty")
 	}
 
-	destDir := filepath.Join(dirs.WorkshopdRunDir, usr.Uid)
+	destDir := filepath.Join(dirs.WorkshopdRunDir, user.Uid)
 	if !osutil.IsDir(destDir) {
```
dmitry-lyfar: No need for this check, just run MkdirAll.

internal/interfaces/lxd_device/spec.go
```diff
@@ -26,14 +27,10 @@ type Specification struct {
 	devices map[string]map[string]string
 	config  map[string]string
 
-	user string
+	User *user.User
 	pid  string
```
dmitry-lyfar, line 30: I think it does not need `pid`, looks like a rudimentary property.

internal/x11/x11_test.go
```diff
@@ -43,27 +43,24 @@ func (x *X11TestSuit) TestCopyXAuthority(c *C) {
...
 	c.Assert(err, IsNil)
 	_, err = os.Stat(filepath.Join(dirs.WorkshopdRunDir, user.Uid, ".Xauthority"))
 	c.Assert(err, IsNil)
 }
 
 func (x *X11TestSuit) TestCopyXAuthorityOwnershipFail(c *C) {
+	user, err := userLookup("root")
+	c.Assert(err, IsNil)
+
 	defer restoreWorkshopdRunDir(dirs.WorkshopdRunDir)
```
dmitry-lyfar: See how other values are mocked. The form is this:

```
func FakeWorkshopRunDir(dir string) func() {
  old := dirs.WorkshopRunDir
  dirs.WorkshopRunDir = dir
  return func() {
    dirs.WorkshopRunDir = old
  }
}
```

and then

```
restore := FakeWorkshopRunDir("something")
defer restore()
```
or

```
// less readable
defer FakeWorkshopRunDir(path)()
```

internal/x11/x11_test.go
```diff
@@ -43,27 +43,24 @@ func (x *X11TestSuit) TestCopyXAuthority(c *C) {
...
 	defer xf.Close()
 	c.Assert(err, IsNil)
 
-	fake := testutil.FakeCommand(c, "sudo", fmt.Sprintf("echo XDG_RUNTIME_DIR=\"/tmp\"\necho XAUTHORITY=\"%s/.workshop-Xauthority\"\nexit 0", dirs.WorkshopdRunDir))
-	defer fake.Restore()
-
-	err = workshoputil.CopyXauthority("root")
+	err = x11.MigrateXauthority(user, filepath.Join(dirs.WorkshopdRunDir, ".workshop-Xauthority"))
 	c.Assert(err, NotNil)
 	c.Assert(err.Error(), testutil.Contains, "Xauthority file isn't owned")
```
dmitry-lyfar: I would use ErrorMatches where possible with a full error message. It makes checking for consistency easier. Sometimes it is not practical and a regex can be used.

internal/x11/x11_test.go
```diff
@@ -43,27 +43,24 @@ func (x *X11TestSuit) TestCopyXAuthority(c *C) {
...
-	defer fake.Restore()
-
-	err = workshoputil.CopyXauthority(user.Username)
+	err = x11.MigrateXauthority(user, filepath.Join(dirs.WorkshopdRunDir, ".workshop-Xauthority"))
 	c.Assert(err, IsNil)
 	_, err = os.Stat(filepath.Join(dirs.WorkshopdRunDir, user.Uid, ".Xauthority"))
 	c.Assert(err, IsNil)
 }
 
 func (x *X11TestSuit) TestCopyXAuthorityOwnershipFail(c *C) {
```
dmitry-lyfar: I cannot comment on the above line for some reason, but `TestCopyXAuthority` is not informative. Also, please do not import all names from the `check` package. I know it has been used in some packages, but this is not accurate and all new packages import only names they use.

dmitry-lyfar: Note, there is no need for empty Setup / Teardown.

internal/x11/x11_test.go
```diff
@@ -11,10 +10,11 @@ import (
 
 	"github.com/canonical/workshop/internal/dirs"
 	"github.com/canonical/workshop/internal/testutil"
-	"github.com/canonical/workshop/internal/workshoputil"
+	"github.com/canonical/workshop/internal/x11"
 )
 
 var userCurrent = user.Current
```
dmitry-lyfar: userCurrent is used once to get the current user, it is not mocked, so maybe we don't need it here?

dmitry-lyfar (changes requested):
it gets better, I only commented on the latest commit's changes.

internal/x11/x11_test.go
lachypjones: I understand that returning a closure that captures the old directory is clever, however I don't see the appeal for this type of function over using the defer argument evaluation semantics to give us the same thing in less lines (and arguably less cognitive overhead). 

As always, if you have a strong opinion on this (I understand that consistency is one, this feels slightly different to some of the other mocks I have seen however) I'm happy to take the closure approach.

dmitry-lyfar: @lachypjones this is about consistency, it should be done in a way as everywhere else unless there is a really good reason not to. 

Also, this allows for breaking locality for coupled entities, so, assignment and restoration happen in separate places. When reading restore, you will have to look for a value it is set to somewhere else and not in this line or, vice versa, when seeing the restoration logic, you'd need to look for where it is set.

dmitry-lyfar: Another reason is you always must think about the order. If called in a reverse order, there will be no proper restore.

lachypjones: Easy done. Thanks for the clear explanation.

docs/explanation/interfaces/desktop-interface.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code-block:: console
+
+   $ workshop connections --all
+
+     Interface  Plug                   Slot       Notes
+     ...
+     desktop    ws/desktop-sdk:desktop :desktop   manual
+
+
+This means the host's Wayland/X11 socket is available inside the workshop:
```
akcano:
```suggestion
This means the host's Wayland or X11 socket is available inside the workshop:
```

akcano (changes requested):

docs/explanation/interfaces/desktop-interface.rst
lachypjones: @akcano, it's not an 'or' relationship, it's an 'and/or' relationship.
I felt that 'Wayland/X11' was cleaner than 'Wayland and/or X11'.

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,133 @@
...
+		w.Connect = filepath.Join(xdg, wayland)
+		w.Listen = filepath.Join("/run/user/1000/", wayland)
+	}
+
+	// We pass through the X11 socket regardless of whether XAUTHORITY is present
+	// on the host. This then gives users the option to modify their xhost
+	// settings to allow connections from the container and container user.
+	if display != "" {
+		// Add X11 to the profile string
+		x := &desktop.X11
```
dmitry-lyfar: Not sure why do we need another indirection here when you can just:

```
d := desktop{}
// ...
// Add wayland to the profile string
d.Wayland.Name = plug.Sdk().Name + "-" + "wayland"
d.Wayland.Connect = filepath.Join(xdg, wayland)
d.Wayland.Listen = filepath.Join("/run/user/1000/", wayland)
```

internal/interfaces/builtin/desktop.go
```diff
@@ -0,0 +1,133 @@
...
+		w.Name = plug.Sdk().Name + "-" + "wayland"
+		w.Connect = filepath.Join(xdg, wayland)
+		w.Listen = filepath.Join("/run/user/1000/", wayland)
+	}
+
+	// We pass through the X11 socket regardless of whether XAUTHORITY is present
+	// on the host. This then gives users the option to modify their xhost
+	// settings to allow connections from the container and container user.
+	if display != "" {
+		// Add X11 to the profile string
```
dmitry-lyfar: What is a profile string?

internal/interfaces/lxd_device/backend.go
```diff
@@ -111,7 +115,7 @@ func installMount(user *user.User, fs workshop.WorkshopFs, dev workshop.Mount) (
 		// configuration which is not acceptable.
 		// The dir is being dynamically created (no source attribute
 		// provided by the slot).
-		if !osutil.IsDir(dev.What) {
+		if _, err := os.Stat(dev.What); err != nil {
```
dmitry-lyfar: `osutil.FileExists`

internal/interfaces/lxd_device/spec.go
```diff
@@ -92,15 +84,28 @@ func (s *Specification) AddMountEntry(dev workshop.Mount) error {
...
-	// see https://documentation.ubuntu.com/lxd/en/latest/reference/devices_proxy/#device-proxy-device-conf:bind
-	// bind denotes where the port is open (can be: instance, host)
 	s.Profile.Agent = &agent
+	addProxyEntry(s, (*workshop.ProxyEntry)(&agent), "ssh-agent")
+	return nil
+}
 
-	s.config[lxdbackend.DeviceTypeConfigKey(s.Profile.Sdk, agent.Name)] = "ssh-agent"
-	s.devices[agent.Name] = map[string]string{"type": "proxy", "connect": "unix:" + agent.Connect, "listen": "unix:" + agent.Listen, "uid": "1000", "gid": "1000", "bind": "instance"}
+func (s *Specification) SetDesktop(desktop *workshop.Desktop) error {
```
dmitry-lyfar: Other similar methods receive a copy here. The pointer value internally is used to check if a device exists in a workshop.

internal/interfaces/lxd_device/spec.go
```diff
@@ -162,3 +167,15 @@ func (s *Specification) SetCamera(camera workshop.Camera) error {
 
 	return nil
 }
+
+func addProxyEntry(s *Specification, entry *workshop.ProxyEntry, configKey string) {
```
dmitry-lyfar: Why is it a standalone method receiving `s`? It should be an unexported method of the Specification.

internal/interfaces/lxd_device/spec.go
```diff
@@ -92,15 +84,28 @@ func (s *Specification) AddMountEntry(dev workshop.Mount) error {
...
+// see https://documentation.ubuntu.com/lxd/en/latest/reference/devices_proxy/#device-proxy-device-conf:bind
+// bind denotes where the port is open (can be: instance, host)
+
 func (s *Specification) SetSshAgent(agent workshop.SshAgent) error {
-	// A network protocol proxy device, opens a port on the host or in a workhop.
-	// from, to are the source and destination addresses (paths in the case of unix sockets),
-	// see https://documentation.ubuntu.com/lxd/en/latest/reference/devices_proxy/#device-proxy-device-conf:bind
-	// bind denotes where the port is open (can be: instance, host)
 	s.Profile.Agent = &agent
+	addProxyEntry(s, (*workshop.ProxyEntry)(&agent), "ssh-agent")
```
dmitry-lyfar: It's OK to create an `SshAgent` struct that would contain a `workshop.ProxyEntry` member instead of using aliases. It is extendable and avoids something like this.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -721,6 +721,7 @@ users:
 		"user.workshop.project-id": projectId,
 		"user.user-data":           cloudInitConfig,
 		"user.workshop.file":       string(f),
+		"raw.lxc":                  "lxc.mount.entry = tmpfs tmp tmpfs defaults",
```
dmitry-lyfar: We need to comment this well (with a link to the discussion).

internal/workshop/lxd/lxd_backend_profile.go
```diff
@@ -91,3 +96,25 @@ func lxdToSdkProfile(profile string, devs map[string]map[string]string, config m
...
 	return pr, nil
 }
+
+func parseDesktop(name string, dev map[string]string, pr *workshop.SdkProfile, backend string) {
+	if pr.Desktop == nil {
+		pr.Desktop = &workshop.Desktop{}
+	}
+	switch backend {
+	case "wayland":
+		{
```
dmitry-lyfar: There are no (?) {} in the multi-line case statements elsewhere, as quick search shows.

internal/workshop/lxd/lxd_backend_profile.go
```diff
@@ -48,12 +48,17 @@ func lxdToSdkProfile(profile string, devs map[string]map[string]string, config m
 			pr.Gpu = &workshop.Gpu{Name: name}
 		case "proxy":
 			devtype := config[DeviceTypeConfigKey(profile, name)]
-			if devtype == "ssh-agent" {
+
+			switch devtype {
+			case "ssh-agent":
 				pr.Agent = &workshop.SshAgent{Name: name, Connect: dev["connect"], Listen: dev["listen"]}
-				continue
+			case "desktop-wayland":
```
dmitry-lyfar, line 54: why not:

```
			switch devtype {
			case "ssh-agent":
				pr.Agent = &workshop.SshAgent{Name: name, Connect: dev["connect"], Listen: dev["listen"]}
			case "desktop-wayland":
				pr.Desktop.Wayland = &workshop.ProxyEntry{Name: name, Connect: dev["connect"], Listen: dev["listen"]}
			case "desktop-x11":
				pr.Desktop.X11 = &workshop.ProxyEntry{Name: name, Connect: dev["connect"], Listen: dev["listen"]}
			default:
				logger.Noticef("On reading %q SDK profile: unknown device type: %s", profile, devtype)
			}
```

It would also use the same semantics (nil pointer indicates that value can be not set), so you don't have to compare the name of the ProxyEntry for "", if it is nil, it does not exist.

dmitry-lyfar (changes requested):
Smoke test caused this error (two plugs, gpu and desktop in a hack SDK):

```
- Disconnnect desk/hack:gpu from desk/system:gpu (/var/lib/workshop/run/.Xauthority is not a directory)
"desk" refresh aborted
```

internal/interfaces/builtin/desktop.go
lachypjones: No specific reason. I understand it's probably easier to read it without having to understand what d actually refers to. No problem

internal/interfaces/lxd_device/spec.go
lachypjones: Again, happy to change for consistency's sake. 
Why make a copy when it's cheaper to copy a pointer, legitimately interested. I understand it's a *tiny* data structure

internal/interfaces/lxd_device/spec.go
lachypjones: Fair call.
I wanted to avoid modifying all the unit tests in-case you were not happy with breaking things out into a struct. This was the smallest modification way of doing that.

internal/interfaces/lxd_device/spec.go
dmitry-lyfar: Having a copy tells you that you control its life-time vs. having a pointer that is allocated and managed by someone else. Also, the client sees there will be no modification. Pointers in these cases are an unnecessary level of indirection and performance considerations are not something I would optimise for. Also, https://go.dev/wiki/CodeReviewComments#pass-values

The usage of pointers internally is intentional; that allows to check whether a device installed without inspecting its fields.

docs/explanation/interfaces/desktop-interface.rst
akcano:
```suggestion
This means the host's socket (Wayland, X11 or both) is available inside the workshop:
```
@lachypjones I have *strong* feelings about the use of slashes as punctuation. Make English great again!

lachypjones: @akcano I like this last suggestion.
I'll make a minor edit 
```suggestion
This means the host's display socket (Wayland, X11 or both) is available inside the workshop:
```

internal/workshop/lxd/lxd_backend.go
lachypjones: I think I'll get rid of this.
If we're going to symlink the Xauth cookie, we mightaswell symlink the X and Wayland sockets also?

akcano (approved):

internal/interfaces/builtin/desktop_test.go
```diff
@@ -0,0 +1,288 @@
...
+	fake := testutil.FakeCommand(c, "sudo", `
+echo "XDG_RUNTIME_DIR=/tmp"
+echo "DISPLAY=:0"
+exit 0`)
+	defer fake.Restore()
+
+	c.Assert(deviceSpec.AddConnectedPlug(s.iface, connectedPlug, connectedSlot), check.IsNil)
+	_, ok := deviceSpec.Profile.Mounts["consumer-xauth"]
+	c.Assert(!ok, check.Equals, true)
+}
```
dmitry-lyfar, line 211:
```suggestion
}

```

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -175,6 +178,26 @@ func (m *InterfaceManager) StartUp() error {
 				}
 			}
 		}
+
+		// The .Xauthority cookie contains a 128bit key used to authenticate
+		// consumers of the X11 socket. It is generated on each boot with a random
+		// suffix, because of this we need to ensure there exists a
+		// consistently-named copy of the cookie for the LXC profile. We handle
+		// this consistency across reboots here.
+		usr, err := workshop.LookupUsername(user)
```
dmitry-lyfar: I would extract it as a separate function, similarly to `recreateInternalMounts`, e.g. `updateXAuthority`.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -699,6 +699,12 @@ func checkNvidia() (bool, error) {
 	return nvidiaRuntime, nil
 }
 
+// The following 'write-files' and 'runcmd' sections are for the desktop
```
dmitry-lyfar, line 702: Please extend why we do a copy instead of a symlink.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -699,6 +699,12 @@ func checkNvidia() (bool, error) {
...
+// interface. These create a systemd path/service unit pair to copy the
+// Xauthority cookie to /tmp when we mount it in the workshop. This is done to
+// work around file mount ordering complications with lxc. Although these will
```
```suggestion
// work around file mount ordering complications with lxc (files, mounted at /tmp, are not visible; mounts at /tmp are generally advised against in LXD). Although these will
```

internal/workshop/lxd/lxd_backend.go
```diff
@@ -721,6 +756,11 @@ users:
 		"user.workshop.project-id": projectId,
 		"user.user-data":           cloudInitConfig,
 		"user.workshop.file":       string(f),
+		// LXC appears to have a race condition wherein a proxy device mounted in
+		// a dynamically created directory has the potential to be 'masked' by this
+		// directory. We create an explicit mount for /tmp here (one such dymanic
+		// directory) to allow us to mount X11 sockets reliably
```
```suggestion
		// directory) to allow us to mount X11 sockets reliably
		// See: https://github.com/lxc/lxc/issues/434
```

tests/lib/sdk/test-sdk-desktop/sdk/hooks/setup-base
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/bash
```
dmitry-lyfar: FYI, this is not required.

tests/main/interface-desktop/task.yaml
```diff
@@ -0,0 +1,114 @@
+summary: Check desktop interface scenarios
```
dmitry-lyfar, line 1: This is a very nice test suite. I would add one more scenario (as a separate empty workshop definition here): refresh to a new workshop definition that does not have the desktop plug. Would be handy to guard scenarios we are going to have when we expand Workshop to support per-SDK refreshes or partial refresh.

dmitry-lyfar (approved):
This looks great. I'm happy to merge it, just a few remarks requested.

docs/explanation/interfaces/index.rst
```diff
@@ -51,8 +51,9 @@ This behaviour varies by interface,
 but the overall aim is to conduct reasonably in each case:
 the :ref:`mount <exp_mount_interface>`
 and the :ref:`GPU <exp_gpu_interface>` interfaces are auto-connected,
-whereas the :ref:`camera <exp_camera_interface>`
-and :ref:`SSH <exp_ssh_interface>` interfaces require manual connection.
+whereas the :ref:`camera <exp_camera_interface>`,
+:ref:`desktop <exp_desktop_interface>` and ref:`SSH <exp_ssh_interface>`
```
jonathan-conder:
```suggestion
:ref:`desktop <exp_desktop_interface>` and :ref:`SSH <exp_ssh_interface>`
```

PR #209, opened 2024-11-15
Preserve apt cache across workshop refreshes

jonathan-conder:
# Description

On creation, Workshop mounts a LXD volume at `/var/cache/apt/archives` so we can save `.deb` packages between refreshes. This can potentially speed up `workshop refresh` by several minutes.

A couple of things I'm not happy with:
- The cache has to be detached before stashing the workshop, maybe this should be lifted to the tasks/changes level, but if we reimplement stash using snapshots this might not be necessary anymore.
- The mounted volume has an unusual mode ([0711](https://github.com/canonical/lxd/blob/ba31c87cd892529bb2a225a5f532104e6d0fc142/lxd/storage/drivers/volume.go#L227)), so normal users can't see the cache contents. I think it's possible to fix this using `CreateStoragePoolVolumeFromBackup`, but generating a backup tarball with the correct metadata seems pretty involved. Might be more productive to ask if LXD can add an option to control the mode. See https://github.com/canonical/lxd/issues/14507.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/overlord/workshopstate/request_test.go
```diff
@@ -167,14 +167,14 @@ func (s *requestSuite) TestLaunchWorkshopNoSdk(c *check.C) {
...
-		"mount-project",
+		"mount-project", "create-apt-cache", "mount-apt-cache",
 		"start-workshop", "install-local-sdk", "link-sdk", "auto-connect"}
 	tasks := ts.Tasks()
 
 	verifyExpectedTasks(c, tasks, expected)
 
 	var wf workshop.File
-	err := tasks[1].Get("workshop-file", &wf)
+	err := tasks[2].Get("workshop-file", &wf)
```
jonathan-conder: I'm wondering if there's a cleaner way to refer to the task in question

internal/overlord/workshopstate/handlers.go
```diff
@@ -108,6 +134,23 @@ func (m *WorkshopManager) undoMountProject(task *state.Task, tomb *tomb.Tomb) er
...
+
+	ctx, cancel := BackendContext(tomb, user, prj.ProjectId)
+	defer cancel()
+
+	volume := workshop.AptCacheVolumeName(w, prj.ProjectId)
+	return m.backend.AttachStorage(ctx, w, volume, dirs.AptCachePath)
+}
+
+func (m *WorkshopManager) undoMountAptCache(task *state.Task, tomb *tomb.Tomb) error {
+	return nil
```
jonathan-conder, line 169: I just copied this from `undoMountProject`, but I think they could use comments explaining why the undo is unnecessary. I guess it's because removing the container will remove the mounts?

dmitry-lyfar: Yes.

internal/overlord/workshopstate/request_test.go
dmitry-lyfar: As per our conversation, it does not look like we need to test this. Side effects of a launch must be ensured in an integration test instead. The question how to ensure serial execution where expected is still open.

internal/overlord/workshopstate/handlers.go
jonathan-conder: is there any difference between having this function or just passing nil to `runner.AddHandler`?

dmitry-lyfar: No practical difference. However, in the `workshop tasks` output for this change, if it failed, these tasks will be in the Done state because no undo handler exists. I found it confusing, hence, the nil handler.

We still have tasks with no undo handlers (e.g. download-sdk), but in that case it is logical, we don't undo downloads.

internal/overlord/workshopstate/handlers.go
```diff
@@ -90,6 +93,37 @@ func (m *WorkshopManager) doCreateWorkshop(task *state.Task, tomb *tomb.Tomb) er
...
+func (m *WorkshopManager) doCreateAptCache(task *state.Task, tomb *tomb.Tomb) error {
+	user, prj, w, err := UserProjectWorkshop(task)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := BackendContext(tomb, user, prj.ProjectId)
+	defer cancel()
+
+	// TODO: ideally the root of the volume should have 0755 permissions
```
dmitry-lyfar: This comment is a bit difficult to understand without the context (maybe, explain the LXD issue here, or move the command to the actual implementation)?

internal/workshop/backend.go
```diff
@@ -68,19 +69,19 @@ type Stash interface {
...
 }
 
-type StateStorage interface {
-	// Create a temporary state storage volume for the workshop. It does not
+type Storage interface {
+	// Create a temporary storage volume for the workshop. It does not
 	// mount the device to the workshop, it must be mounted to the required
 	// workshop as a separate operation.
-	CreateStateStorage(ctx context.Context, name string) error
+	CreateStorage(ctx context.Context, name string) error
```
dmitry-lyfar: CreateStorage creates a storage volume that can be mounted to a workshop (as per the comment). So, maybe, the function name should be CreateVolume? It used to be CreateStateStorage, an attempt to abstract an implementation detail (storage in the sense of storage for the SDK states). Not relevant anymore.

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -112,43 +113,27 @@ func (f *wsOps) TestLxdBackendWorkshopStashRemove(c *check.C) {
...
 
 	// Execute
-	err = f.bd.DeleteStateStorage(f.ctx, "test")
+	err = f.bd.DeleteStorage(f.ctx, "test")
 
 	// Validate
 	c.Assert(err, check.IsNil)
 }
 
-func (f *wsOps) TestLxdBackendRemoveWorkshopStash(c *check.C) {
```
dmitry-lyfar, line 132: I forgot, why does this one has to go?

jonathan-conder:
Performance numbers (for launch, 5 refreshes, 3 more launches, 5 more refreshes):

| Action | Launch | Refresh | Refresh | Refresh | Refresh | Refresh | Launch | Launch | Launch | Refresh | Refresh | Refresh | Refresh | Refresh |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Without cache | 4m2.648s | 4m14.075s | 4m19.082s | 4m27.221s | 5m15.033s | 36m28.457s | 4m13.631s | 3m58.143s | 4m49.418s | 4m25.245s | 4m18.097s | 4m13.134s | 36m11.011s | 4m22.639s |
| With cache | 4m13.462s | 3m25.109s | 3m39.652s | 3m24.196s | 3m27.861s | 3m23.559s | 3m50.849s | 3m56.504s | 4m23.184s | 3m33.099s | 3m31.520s | 3m32.198s | 3m30.086s | 3m29.756s |

Workshop:
```yaml
name: ws
base: ubuntu@24.04
sdks:
  ros2:
    channel: latest/edge
```

Looks like there's something which occasionally delays `refresh` by a huge amount, might just be networking issues but probably worth investigating in more detail

internal/workshop/lxd/tests/integration/workshop_test.go
jonathan-conder: This one is virtually identical to `TestLxdBackendWorkshopStashRemove`, only needed to change 1 line. See 34a75c3b070b0b1a65c2b98d62cb16fefec06a0a

internal/workshop/backend.go
jonathan-conder: Yeah that makes sense

jonathan-conder:
New performance numbers (with snap install --no-wait):

| Action | Launch | Refresh | Refresh | Refresh | Refresh | Refresh | Launch | Launch | Launch | Refresh | Refresh | Refresh | Refresh | Refresh |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Without cache | 2m50.949s | 3m11.888s | 3m19.044s | 3m11.453s | 4m4.240s | 3m28.952s | 3m24.390s | 2m59.662s | 8m10.306s | 3m2.280s | 3m35.906s | 3m9.674s | 3m20.630s | 3m0.266s |
| With cache | 2m58.285s | 2m30.077s | 2m35.456s | 2m29.745s | 2m31.708s | 2m33.872s | 3m7.774s | 3m0.516s | 3m3.344s | 2m38.299s | 2m36.206s | 2m36.691s | 2m32.461s | 2m40.339s |

Workshop:
```yaml
name: ws
base: ubuntu@24.04
sdks:
  ros2:
    channel: jazzy/edge
```

SDK: https://github.com/canonical/sdks/tree/3f3a63f804b58c2a0208933faa94e38a2050d855/ros2

dmitry-lyfar (approved):

PR #212, opened 2024-11-20
Update LXD client library

dmitry-lyfar:
# Description

Recent LXD client library revision require go 1.23. Thus, go version and some other dependencies were updated too. The upgrade is done in preparation to update dependency on LXD server from 5.x to 6.x.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


jonathan-conder (approved):
LGTM

lachypjones (approved):

PR #214, opened 2024-11-25
Use systemd-run to spawn exec/shell sessions

lachypjones:
# Description

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"time"
 
@@ -92,9 +93,8 @@ Notes:
 
 - To start a workshop before running a terminal session, use 'workshop start'
 
-- The subcommand is a shorthand for 'workshop exec';
-  it launches the login shell for 'workshop',
-  the default non-privileged user in a workshop
+- You can set the working directory, environment variables, user and group ID
+  for the spawned shell; reasonable defaults are provided
 `
 
 func (c *CmdExec) Command() *cobra.Command {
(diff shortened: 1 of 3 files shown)


cmd/workshop/exec.go
```diff
@@ -187,9 +166,15 @@ func (c *CmdExec) Run(cmd *cobra.Command, av []string) error {
 		interactive = false
 	} else {
 		interactive = stdinIsTerminal && stdoutIsTerminal
+		// If both --pty and --pipe are specified, systemd-run will infer from
```
lachypjones, line 169: This is a dead comment, I'll remove it at a later time

tests/main/interface-mount/task.yaml
```diff
@@ -1,11 +1,11 @@
 summary: Check mount interface scenarios
 restore: |
-  workshop_exec remove ws-mount || true
```
lachypjones, line 3: utils.sh is not sourced by restore, meaning that this command doesn't remove the workshop

cmd/workshop/exec.go
```diff
@@ -92,9 +93,8 @@ Notes:
 
 - To start a workshop before running a terminal session, use 'workshop start'
 
-- The subcommand is a shorthand for 'workshop exec';
-  it launches the login shell for 'workshop',
-  the default non-privileged user in a workshop
+- You can set the working directory, environment variables, user and group ID
+  for the spawned shell; reasonable defaults are provided
```
akcano, line 97: I'd very much like to see the relevant defaults listed, not just mentioned here.

lachypjones: This matches the documentation for exec. Happy to expand.

cmd/workshop/exec.go
```diff
@@ -127,11 +127,17 @@ func (c *CmdShellAlias) Command() *cobra.Command {
...
+	cmd.Flags().StringArrayVar(&c.execCommand.Env, "env", []string{}, "Set an environment variable, e.g. 'FOO=bar'; if only the name is provided, the value is inherited from the CLI environment.")
+	cmd.Flags().IntVar(&c.execCommand.UserId, "uid", 1000, "Run as a specific workshop user")
+	cmd.Flags().IntVar(&c.execCommand.GroupId, "gid", 1000, "Run as a member of a specific workshop group")
+	c.execCommand.Interactive = true
 	return cmd
 }
 
 func (c *CmdShellAlias) Run(cmd *cobra.Command, av []string) error {
-	return c.execCommand.Run(cmd, []string{av[0], "su", "-l", "workshop"})
+	return c.execCommand.Run(cmd, []string{av[0], "/bin/bash", "-l"})
```
dmitry-lyfar, line 140: This does not set:

```
XDG_RUNTIME_DIR=/run/user/1000
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
```

lachypjones: Neither does shell. They'll be added to the environment inherited by the unit

lachypjones: Also I believe the DBUS address is automatically inferred (as $XDG_RUNTIME_DIR/bus) so we should not need to set it explicitly.

cmd/workshop/exec.go
akcano: @lachypjones I'd say the exact full form of the corresponding `exec` command will do just fine.

lachypjones:
Closing for now, to be addressed in conjunction with modifying interface environment variables.

PR #219, opened 2024-11-28
Doc: Merge with SDKcraft docs

akcano:


dmitry-lyfar (commented):
This looks (and feels) much better. I'm surprised how neat the merge came out.

Before merging, I'd like to address two issues:

1. The explanation section for "Workshop, SDKs" can be separated, so we have "Workshops" and "SDKs". So, things like "Dockerfiles vs. SDKs" do not live at the same level with "Projects" and "Workshops".
2. The same applies to the How-to section where topics about Workshop and SDKs are mixed. I'm not sure about using the same approach (i.e. separating topics into Workshop and SDK how-tos). 

The landing page bit should go on top of this PR, I think we are in a principal agreement to merge Workshop and SDKcraft documentation.

akcano:
1. Addressed.
2. No need to have an entire subsection, as we only have a single SDK-related how-to guide right now. I added a separate landing page section, though.

dmitry-lyfar (approved):

PR #224, opened 2024-12-05
Remove volumes after images when uninstalling snap

jonathan-conder:
# Description

Occasionally I get spread errors in remove-snap, which seems to be caused by LXD refusing to remove certain storage volumes. It seems to be caused by volumes of type 'image' so I'm hoping removing the images first will fix it.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/snap/local/hooks/remove
+++ b/snap/local/hooks/remove
@@ -23,18 +23,19 @@ for p in $(echo "$projects" | xargs -L1 echo); do
     fi
   done
   
-  volumes=$(lxc storage volume list default -c n -f compact --project "$p" | awk 'NR>1 {printf "%s ", $1}')
-  echo "Removing storage volumes: $volumes"
-  for volume in $(echo "$volumes" | xargs -L1 echo); do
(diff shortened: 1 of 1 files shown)


dmitry-lyfar (approved):

snap/local/hooks/remove
```diff
@@ -23,18 +23,19 @@ for p in $(echo "$projects" | xargs -L1 echo); do
...
-  done
-  
   images=$(lxc image list -f compact --project "$p"| awk 'NR>1 {printf "%s ", $1}')
   if [ -n "$images" ]; then
     echo "Removing images: $images"
     # shellcheck disable=SC2086
     lxc image delete $images --project "$p"
   fi
 
+  lxc storage volume list default --project "$p"
```
dmitry-lyfar, line 33: The header comment does not include volumes, also, could you fix s/configure/remove/ hook in the same comment?

PR #225, opened 2024-12-05
Implement wait on error for launch

lachypjones:
# Description

This PR allows use of the `--wait-on-error`, `--continue` and `--abort` flags with the `workshop launch` command.
The behavior mirrors that of the corresponding flags for `workshop refresh`. 

## Breaking Changes

- A launch/refresh aborted after using wait-on-error will now show `Undone` as the status of the change, rather than `Error`. This is done to work within the expectations of the state package. 
- 'RefreshMode' and all variations of the same concept (JSON representations, etc) have been renamed to 'Mode'.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


docs/how-to/debug-workshop-issues.rst
```diff
@@ -91,10 +91,13 @@ instead of reverting the workshop to its previous state,
 
    $ workshop refresh --wait-on-error golang-volatile
 
-     Error: cannot perform the following tasks:
-     - Run hook "setup-base" for "go" SDK (command failed with an error code (1))
-     Refresh aborted
+     YYYY-MM-DDT00:00:00 ERROR command exit code 1
```
lachypjones: Note this change is not specific to this PR, I believe the documentation was incorrect.
 
@dmitry-lyfar this does then raise an interesting point though, the way that --wait-on-error reports what failed is worse than a transactional change. I was thinking of raising a separate PR to attempt to address this?

docs/how-to/debug-workshop-issues.rst
```diff
@@ -91,10 +91,13 @@ instead of reverting the workshop to its previous state,
...
+     error: cannot launch; fix the errors reported,
+     then run "workshop launch --continue blank".
+     To abort and revert, run "workshop launch --abort blank"
 
+To help determine what went wrong, use the :command:`workshop_changes` and
+:command:`workshop_tasks` commands discussed above
```
akcano:
```suggestion
To determine what went wrong, use the :command:`workshop changes` and
:command:`workshop tasks` commands discussed above.
```

docs/how-to/debug-workshop-issues.rst
akcano: Command output in docs gets stale the moment you commit it; I'm looking at the ways to automate this as part of the general doc automation initiative.

tests/main/launch-abort/task.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+  echo "Update the workshop file to include a failing SDK"
+  sed -i 's/test-sdk-basic/test-sdk-setup-fail/g' .workshop/workshop.ws-launch-abort.yaml
+  workshop_exec launch --wait-on-error ws-launch-abort || true
+  sed -i 's/test-sdk-setup-fail/test-sdk-basic/g' .workshop/workshop.ws-launch-abort.yaml
+
+  echo "Check the workshop is in Pending state"
+  workshop_exec list | MATCH "^.+[[:space:]]+ws-launch-abort[[:space:]]+Pending[[:space:]]+wait-on-error$"
+
+  echo "Check the workshop launch --abort reverts to the previous state"
```
akcano:
```suggestion
  echo "Check the 'workshop launch --abort' command reverts to the previous state"
```

tests/main/launch-continue/task.yaml
```diff
@@ -0,0 +1,23 @@
...
+  workshop_exec launch --wait-on-error ws-launch-cont || true
+  sed -i 's/test-sdk-setup-fail/test-sdk-basic/g' .workshop/workshop.ws-launch-cont.yaml
+
+  echo "Check the workshop is in Pending state"
+  workshop_exec list | MATCH "^.+[[:space:]]+ws-launch-cont[[:space:]]+Pending[[:space:]]+wait-on-error$"
+
+  echo "\"Fix\" the SDK error by replacing the setup-base hook"
+  workshop_exec exec --uid 0 --gid 0 ws-launch-cont -- sed -i "s/exit 1/exit 0/g" /var/lib/workshop/sdk/test-sdk-setup-fail/current/sdk/hooks/setup-base
+
+  echo "Check the workshop launch --continue reverts to the previous state"
```
akcano: same as above

akcano (changes requested):
LGTM with minor fixes

client/projects.go
```diff
@@ -12,7 +12,7 @@ type Project struct {
 }
 
 type WorkshopActionOptions struct {
-	RefreshMode string `json:"refresh-mode,omitempty"`
+	ChangeMode string `json:"change-mode,omitempty"`
```
dmitry-lyfar:
```suggestion
	ActionMode string `json:"action-mode,omitempty"`
```

dmitry-lyfar: The mode is used in the context of an action (refresh, launch, etc.) for a workshop. Change is a broader underlying concept.

dmitry-lyfar: Alternatively, it can be just `Mode`, as it is part of an action request already.

dmitry-lyfar: Also, please mention this in the comment for the PR (see the checkbox about breaking / irreversible changes).

internal/daemon/api_workshops.go
```diff
@@ -272,14 +293,33 @@ func refresh(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopMa
...
+	if err != nil {
+		return change, taskset, err
 	}
+
+	switch {
+	case resume:
+		change, err = conflict.ResumeAfterWait(st, reqData.Names[0], pid, changeMode, "launch")
+	default:
+		change = newWorkshopChange(st, "launch", user, pid, reqData.Action, reqData.Names)
+		taskset, err = mgr.LaunchMany(ctx, reqData.Names, pid, change.ID())
```
dmitry-lyfar: This, compared to `RefreshMany`, does not check for the workshop status as previously it did not make sense (the operation failed if workshop existed).

Now:

```

$ workshop launch --wait-on-error io
$ workshop launch io
error: cannot launch "io": workshop already exists
$ workshop changes
1168  Wait    today at 12:18 NZDT  -                    Launch "io" workshop

# this second change exist because it was created due to the lack of status check
1169  Hold    today at 12:19 NZDT  -                    Launch "io" workshop
```

dmitry-lyfar:
```
workshop launch --wait-on-error io
# errors out
workshop launch io 
# reports workshop exists AND creates a new change that's on Hold because there is no 
# status check in LaunchMany (see how refresh prevents it).
```

dmitry-lyfar (changes requested):
Try to `workshop launch --continue` multiple times on a failed launch and then abort. There will be a lot of repeating errors reported and, also, an error logged by Errorf(task.Summary()). The error message on abort does already include errors, so there is no need to log something else. Also, as per the client code, it seems only the last error will be reported, not sure why duplicating errors are reported in the output.

Results of my smoke testing: another side effect of this change is that UnlinkSdk must check if the current link exists. As before it wasn't a possible scenario (there was always a 'current' link in a workshop for the SDK). Please address this (easy to reproduce with `test-sdk-content-broken`).

internal/daemon/api_workshops.go
lachypjones: Refresh doesn't prevent creating holds.
```
35  Done    today at 13:38 ACDT  today at 13:38 ACDT  Launch "blank" workshop
36  Wait    today at 13:39 ACDT  -                    Refresh "blank" workshop
37  Hold    today at 13:39 ACDT  -                    Refresh "blank" workshop
38  Hold    today at 13:39 ACDT  -                    Refresh "blank" workshop
```

Tested on the current main branch.

lachypjones: This happens because we always create the change before performing the operation. I originally modified this behavior (as I thought it was a bug) however then noticed refresh did the same thing so kept it consistent. 

The status check itself is still an omission, I'll fix that now. 

Maybe we fix the hold behavior in a new PR?

internal/daemon/api_workshops.go
```diff
@@ -261,22 +261,43 @@ func maybeSdkRefresh(names []string) (wp string, sk string, partial bool) {
...
+	case conflict.ChangeContinue, conflict.ChangeAbort:
+		resume = true
+	}
+
+	return mode, resume, nil
+}
+
+func refresh(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopManager, reqData *workshopReq, user, pid string) (*state.Change, []*state.TaskSet, error) {
+	var taskset []*state.TaskSet
+	var change = &state.Change{}
```
dmitry-lyfar: This initialisation is dangerous. It creates a non-nil change that does not exist in the state. See `st.NewChange` on how changes should be created. This must be a nil pointer here.

internal/daemon/api_workshops.go
```diff
@@ -261,22 +261,43 @@ func maybeSdkRefresh(names []string) (wp string, sk string, partial bool) {
...
+
+	mode, resume, err := maybeResume(reqData)
+	if err != nil {
+		return change, taskset, err
+	}
+
+	switch {
```
dmitry-lyfar: It is if-else on a boolean flag, I would not do the switch assuming there are other scenarios yet .

internal/daemon/api_workshops.go
```diff
@@ -287,14 +308,33 @@ func refresh(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopMa
...
+		var setup conflict.ChangeSetup
+		setup.Mode = mode.String()
+		change.Set("wait-setup", setup)
 	}
+	return change, taskset, err
+}
+
+func launch(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopManager, reqData *workshopReq, user, pid string) (*state.Change, []*state.TaskSet, error) {
+	var taskset []*state.TaskSet
+	var change = &state.Change{}
```
dmitry-lyfar: Same here.

internal/daemon/api_workshops.go
```diff
@@ -287,14 +308,33 @@ func refresh(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopMa
...
 
-	if refreshMode == conflict.RefreshContinue || refreshMode == conflict.RefreshAbort {
-		change, err = conflict.ResumeRefresh(st, reqData.Names[0], pid, refreshMode)
+	mode, resume, err := maybeResume(reqData)
+	if err != nil {
+		return change, taskset, err
 	}
+
+	switch {
```
dmitry-lyfar: Same here.

internal/daemon/api_workshops.go
```diff
@@ -328,8 +368,7 @@ func v1PostProjectWorkshop(c *Command, r *http.Request, _ *userState) Response {
 
 	switch reqData.Action {
 	case "launch":
```
dmitry-lyfar: The `maybeResume` and `Mode` can be simplified this way (actually, I think it should just validate and return the mode):

```
// conflict.go
func (s Mode) Resume() bool {
	return s == ChangeContinue || s == ChangeAbort
}

// maybeResume -> actionMode
func actionMode(reqData *workshopReq) (conflict.Mode, error) {
	var mode conflict.Mode
	var err error

	mode, err = conflict.ParseMode(reqData.Options.Mode)
	if err != nil {
		return mode, resume, fmt.Errorf("cannot %s: %v", reqData.Action, err)
	}

	if len(reqData.Names) > 1 && mode != conflict.ChangeTransactional {
		return mode, resume, fmt.Errorf("wait-on-error is not supported for multiple workshops")
	}
	return mode, nil
}
```

I like that some additional refresh logic went to a separate func, that would make dispatching (if we decide to have it) easier. But I would suggest moving the resume parameters validation (i.e. `maybeResume`) before the dispatch and have it together with other request validation logic. This would resolve duplication in launch() and refresh(), free them from deciding on whether to resume or start a new one, and result in something like:

```
/// ...validation
	mode, err := actionMode(&reqData)
	if err != nil {
		return statusBadRequest(...)
	}

	var change *state.Change
	var taskset = []*state.TaskSet{}

// dispatching
	if mode.Resume() {
		change, err = conflict.ResumeAfterWait(st, reqData.Names[0], projectId, mode, reqData.Action)
	} else {
		switch reqData.Action {
		case "launch":
			change = newWorkshopChange(st, "launch", user, projectId, reqData.Action, reqData.Names)
			taskset, err = wsmgr.LaunchMany(r.Context(), reqData.Names, projectId, change.ID())
			change.Set("wait-setup", conflict.ChangeSetup{Mode: mode.String()})
		case "refresh":
			// similar thing, but keep the function refresh() as it has more logic
		case "start":
			change = newWorkshopChange(st, "start", user, projectId, reqData.Action, reqData.Names)
			taskset, err = wsmgr.StartMany(r.Context(), reqData.Names, projectId, change.ID())
		case "stop":
			change = newWorkshopChange(st, "stop", user, projectId, reqData.Action, reqData.Names)
			taskset, err = wsmgr.StopMany(r.Context(), reqData.Names, projectId, change.ID())
		case "remove":
			change = newWorkshopChange(st, "remove", user, projectId, reqData.Action, reqData.Names)
			taskset, err = wsmgr.RemoveMany(r.Context(), reqData.Names, projectId, change.ID())
		default:
			return statusBadRequest("unknown action")
		}
	}
```

dmitry-lyfar (changes requested):

dmitry-lyfar:
@akcano can you please have a look into the new error and log messages, specifically, Aborting... and Continuing... for --continue and --abort?

internal/daemon/api_workshops.go
lachypjones: I really liked this approach, however there is a problem.
The mode is only set for 'launch' and 'refresh' commands, there is not a clean way to incorporate the above because it needs to check the mode first.

We can 'assume' the mode is transactional if it's otherwise unset, however I don't think that's a good solution.

dmitry-lyfar: @lachypjones the default mode for all actions is, indeed, transactional because it is an action's mode not refresh or launch mode anymore. For some actions, only the transactional mode is supported. What is the exact problem with this?

lachypjones: We only set the mode explicitly for 'launch' and 'refresh'. For all other modes it's inferred (and an optional parameter in the marshalled JSON). If we then infer here, we miss any cases where a caller did not explicitly set for 'launch' and 'refresh'. 

If you're happy with this as a solution, then there is no problem. I assumed from the error messages and current tests (ie. IncorrectInput) that we wanted this to be an explicit choice.

dmitry-lyfar: This should be encapsulated by `actionMode()` which:

1. Will validate compatibility of the mode and the action requested. So that action: start and mode: wait-on-error gets declined by the API call at the validation stage.
2. Infer the mode value.
3. Will ensure that the mode is provided for the actions that require it (so for the `launch` and `refresh` actions the piece of code that sets it to the default mode anyway, I don't think it is happening in your `maybeResume` now. So, if I call a launch with an empty mode attribute, it won't work).

cmd/workshop/launch.go
```diff
@@ -48,6 +54,15 @@ $ workshop launch nimble jazzy`,
...
 
+	cmd.PersistentFlags().BoolVar(&c.WaitOnError, "wait-on-error",
+		false,
+		"Pause the operation on error; to resume, use '--continue' or '--abort'.")
+	cmd.PersistentFlags().BoolVar(&c.Continue, "continue",
+		false,
+		"Continue the previously paused operation.")
+	cmd.PersistentFlags().BoolVar(&c.Abort, "abort",
+		false,
+		"Abort the previously paused operation, removing any changes.")
```
akcano:
```suggestion
		"Abort the previously paused operation, reverting any changes.")
```

cmd/workshop/launch.go
```diff
@@ -58,6 +73,22 @@ $ workshop launch nimble jazzy`,
...
+	if c.WaitOnError && c.Abort {
+		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with '--abort'")
+	}
+
+	if c.WaitOnError && c.Continue {
+		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with '--continue'")
+	}
+
+	if (c.Abort || c.Continue || c.WaitOnError) && len(av) > 1 {
+		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with multiple workshops")
```
akcano, line 93: Not sure what's going on with the conditional logic here.

cmd/workshop/launch.go
akcano: The auto-generated docs do not match this for some reason, so make sure to regenerate the CLI docs with the updated commit.

cmd/workshop/launch.go
lachypjones: This matches exec. 
I'll add a comment for clarity

cmd/workshop/launch.go
lachypjones: This is for the launch command.
I believe 'remove' makes more sense here.
(essentially we disconnect interfaces and remove the container, we don't revert to a previous state like with refresh)

internal/overlord/workshopstate/manager.go
```diff
@@ -55,19 +55,24 @@ func (w *WorkshopManager) Ensure() error {
 
 // Checks the provided workshop has one of the allowed health statuses.
 func (w *WorkshopManager) CheckStatus(ctx context.Context, name, pId string, allowedStatuses []healthstate.Status) error {
+	health := healthstate.HealthState{}
 	wp, err := w.Workshop(ctx, name, pId)
-	if err != nil {
+	switch {
```
dmitry-lyfar: You don't use CheckStatus for the launch anymore, is there a reason to keep these changes?

internal/daemon/api_workshops.go
```diff
@@ -261,39 +261,49 @@ func maybeSdkRefresh(names []string) (wp string, sk string, partial bool) {
...
-	var change *state.Change
-	var taskset []*state.TaskSet
-	var err error
+func actionMode(reqData *workshopReq) (conflict.Mode, error) {
+	var mode conflict.Mode
 
-	refreshMode, err = conflict.ParseRefreshMode(reqData.Options.Mode)
-	if err != nil {
-		return nil, nil, fmt.Errorf("cannot refresh: %v", err)
+	switch reqData.Action {
```
dmitry-lyfar: This is from our conversation:

> It's an action's mode, agnostic of action as per the name.
> That's why it has to be an optional parameter with a default value.

> Sounds good

That means that you don't have to 1. specify transactional mode for all other actions in the tests, 2. you only need to check if it is empty to set to the default value. Then you can validate compatibility between an action and the mode:

> this is the validation function's actionMode job: validate and infer; fail if options are invalid, incompatible, etc.

Also, the logic below, does it allow for the "launch" action with no mode provided?

dmitry-lyfar (changes requested):

internal/overlord/workshopstate/manager.go
lachypjones: I kept this because the previous healthcheck implementation has OffStatus as a valid case in the switch statement, however it's impossible to reach this state without these changes.

internal/daemon/api_workshops_test.go
```diff
@@ -1253,58 +1253,109 @@ func (s *apiSuite) TestValidatev1PostProjectWorkshopInputs(c *check.C) {
...
+				"abort":         `cannot start: mode "abort" is not valid with this command`,
+				"invalid-mode":  `cannot start: "invalid-mode" is not a valid mode`,
+			},
+		}, {
+			cmd:   "stop",
+			valid: true,
+			result: map[string]string{
+				"":              `cannot stop "basic": workshop not launched`,
+				"transactional": `cannot stop "basic": workshop not launched`,
+				"wait-on-error": `cannot stop: mode "wait-on-error" is not valid with this command`,
```
dmitry-lyfar: We should specify the command name, something like `is not valid with "stop" command`.

internal/daemon/api_workshops_test.go
```diff
@@ -1253,58 +1253,109 @@ func (s *apiSuite) TestValidatev1PostProjectWorkshopInputs(c *check.C) {
...
+			result: map[string]string{
+				"":              `cannot remove "basic": workshop not launched`,
+				"transactional": `cannot remove "basic": workshop not launched`,
+				"wait-on-error": `cannot remove: mode "wait-on-error" is not valid with this command`,
+				"continue":      `cannot remove: mode "continue" is not valid with this command`,
+				"abort":         `cannot remove: mode "abort" is not valid with this command`,
+				"invalid-mode":  `cannot remove: "invalid-mode" is not a valid mode`,
+			},
+		}, {
+			cmd:   "invalid-cmd",
```
dmitry-lyfar: This test looks much cleaner. What we can also do is to validate the action before it gets to dispatching, so you actually don't have to parse mode if the action is not supported. This will rid you of the last two rows in your table ("invalid-cmd" and "").

The `valid` flag can go away too if you move the unknown action into a separate test. So, this will check modes and there will be a separate small test that checks actions validity.

cmd/workshop/wait.go
```diff
@@ -169,7 +157,7 @@ func (wmx waitMixin) wait(cli *client.Client, id string, abortExpected bool) (*c
 				return chg, nil
 			}
 
-			// if the change finished as Ready and reported an error, check if
+			// if the change finished as Undone and reported an error, check if
```
dmitry-lyfar: @lachypjones I think the entire comment should be updated as it is no longer true.

lachypjones: See what I've worded it to now.
I think the bulk of the comment still accurately reflects what we're doing (iterating through the task history and checking for errors which reflect abort itself failing). The language around task log needed correcting.

dmitry-lyfar: Yes, also: it was an expected abortion of a failed refresh -> it was an expected abortion OR a failed change

internal/overlord/workshopstate/request.go
```diff
@@ -85,19 +84,19 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
 	taskset := make([]*state.TaskSet, 0, len(names))
 	var sdks []sdk.SdkResult
 	for _, name := range names {
+		// Make sure the workshop doesn't exist
+		_, err := w.Workshop(ctx, name, projectId)
+		if err == nil {
+			return nil, fmt.Errorf("cannot launch %q: workshop exists", name)
+		} else if err != workshop.ErrWorkshopNotLaunched {
```
dmitry-lyfar:
```suggestion
		} else if !errors.Is(err, workshop.ErrWorkshopNotLaunched) {
```

dmitry-lyfar (approved):

PR #226, opened 2024-12-09
Make workshop name optional for some commands

jonathan-conder:
# Description

If a project only has one workshop, the `launch`, `refresh`, `info`, `start`, `stop`, `remove`, `exec`, `shell` and `hack` commands no longer require a workshop name. Internally, these commands call `v1GetProjectWorkshops` and, if the result has a single entry, proceed as before.

The `v1GetProjectWorkshops` API has been extended slightly: the `state` query parameter can be `available`, `all`, or an actual state. The new state `available` is similar to `all`. It shows existing workshops and workshop files. If something is wrong with the workshop files, `available` downgrades the error to a warning and lists the existing workshops, whereas `all` returns the error. The actual states work the same as before (since they ignore workshop files).

Now `workshop list` uses `state=available`, but commands without workshop names use `state=all`, which remains the default if no `state` is provided.

The `exec` command is a special case, because in `workshop exec one two` it's unclear if `one` is the workshop name or the start of a command. The ambiguity is resolved by the presence or absence of a separator (`--`): `workshop exec -- one two` will treat `one two` as a command, whereas `workshop exec one two`, `workshop exec one -- two` and `workshop exec one two --` all run the `two` command in the workshop named `one`.

This PR also tells cobra to stop parsing `exec` arguments after the workshop name. We still remove the first `--` by hand if needed. So for example `workshop exec foo bar --project $P` will now run the command `bar --project $P` in the current project, rather than running `bar` in project `$P`. This is not required for the above `exec` changes, but I think it improves ergonomics, and without this I would be inclined to make `--` mandatory.

Where possible I've updated the tests to remove workshop names from commands. I think there's a nice mix of single- and multi-workshop projects so we should have pretty good coverage of both scenarios. Also updated some unit tests but I haven't added new ones just for this.

Following up on [Dmitry's observation](https://github.com/canonical/workshop/pull/226#issuecomment-2529863557), I've refactored the project tracking logic. I hope it's easier to follow now, but it should have the same behaviour as before, except fewer errors are suppressed (23367e8c583847b12c3a98359b556ccad6d424ba, 14edb67887a4e78c6f4b6b0278cdecc6bc07fbb1, 15ad5fd0100cc688640b1fdbb8de70a6ecfd11b1, 218365694d0e0b55be1907127d9e88408356acd5) and some steps have been reordered.

The latency seems to be caused by multiple projects/workshops and missing-project warnings. One step to address this in future might be to add a `Project()` method to the backend, to avoid the overhead of `Projects()` when the project ID is already known.

There also seems to be a bug where an empty `.workshop.lock` can be created, but I haven't been able to reproduce it reliably. Maybe these changes will fix it anyway.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/reference/cli/workshop-stop.rst
+++ b/docs/reference/cli/workshop-stop.rst
@@ -44,7 +44,7 @@ Stop the nimble and jazzy workshops in the current project directory:
    $ workshop stop nimble jazzy
 
 
-The name is optional if the project only has one workshop:
+The name is optional if the project has only one workshop:
 
 .. code-block:: console
 
--- a/docs/reference/cli/workshop-stop.rst
+++ b/docs/reference/cli/workshop-stop.rst
@@ -33,3 +33,21 @@ Notes:
 
 - To start a stopped workshop, use 'workshop start'
 
+
+.. rubric:: Examples
+
+
+Stop the nimble and jazzy workshops in the current project directory:
+
+.. code-block:: console
+
+   $ workshop stop nimble jazzy
+
+
+The name is optional if the project only has one workshop:
+
+.. code-block:: console
+
+   $ workshop stop
+
+
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -694,7 +694,7 @@ If you no longer need your workshop,
 
 .. code-block:: console
 
-   $ workshop remove golang
+   $ workshop remove
 
 
 This doesn't affect the files in the project directory,
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -345,7 +345,7 @@ To make it *Ready* again, :ref:`start <ref_workshop_start>` the workshop:
 
 .. code-block:: console
 
-   $ workshop start golang
+   $ workshop start
 
 
 Both commands work gracefully,
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -251,7 +251,7 @@ to see what went into your workshop:
 
 .. code-block:: console
 
-   $ workshop info golang
+   $ workshop info
 
      name:     golang
      base:     ubuntu@22.04
@@ -549,7 +549,7 @@ to a new location on the host:
    :emphasize-lines: 14
 
    $ workshop remount golang/go:mod-cache ~/mod/
-   $ workshop info golang
+   $ workshop info
 
      name:     golang
      base:     ubuntu@20.04
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -441,7 +441,7 @@ Next, build it *inside the workshop* using :ref:`exec <ref_workshop_exec>`:
 .. tip::
 
    Since :samp:`golang` is the only workshop in the project,
-   it can be omitted from most :samp:`workshop` commands.
+   it can be omitted from most :command:`workshop` commands.
    For :ref:`exec <ref_workshop_exec>`,
    a name or a separator (:samp:`--`) is required to avoid ambiguity.
    The above command can also be written as:
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -596,7 +596,7 @@ opens an :ref:`SDK definition <exp_sdk_definition>`:
 
 .. code-block:: console
 
-   $ workshop hack golang
+   $ workshop hack
 
 
 Initially, the editor shows a very basic setup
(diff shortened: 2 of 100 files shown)


dmitry-lyfar:
@jonathan-conder thank you! I will get to reviewing it shortly. Whilst the review is in progress, could you think of the following issue: now, commands like `info` have a noticeable delay compared to how it was before when it had only a single API call. These are my `info` calls on the server:

```
GET /v1/projects/984d2192/workshops 242.042578ms 200
GET /v1/projects/984d2192/workshops/optional 117.13134ms 200

GET /v1/projects/984d2192/workshops 247.191314ms 200
GET /v1/projects/984d2192/workshops/optional 128.266784ms 200
```

These are the results for `workshop list --global` (I have several projects, some with warnings):

```
GET /v1/projects 117.29467ms 200
GET /v1/projects/fbc4e283/workshops?state=available 236.121388ms 200
GET /v1/projects/0ce2e11f/workshops?state=available 234.223879ms 200
GET /v1/projects/502e3034/workshops?state=available 226.706331ms 200
GET /v1/projects/4d42157d/workshops?state=available 237.659068ms 200
GET /v1/projects/984d2192/workshops?state=available 256.764649ms 200
GET /v1/projects/24d984c1/workshops?state=available 242.050375ms 200
GET /v1/projects/ad51bb1d/workshops?state=available 259.657018ms 200
```

internal/daemon/api_workshops.go
```diff
@@ -221,27 +222,42 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
...
 	}
 
 	info := Workshops{}
 	info.Workshops = make([]*WorkshopInfo, 0, len(workshops))
 	for _, w := range workshops {
 		health := wrkmgr.WorkshopHealth(w)
-		if wstate != "all" && strings.ToLower(health.Status.String()) != wstate {
-			continue
+		switch wstate {
+		case "all":
```
dmitry-lyfar:
```suggestion
		case "all", "available":
```

internal/daemon/api_workshops.go
```diff
@@ -221,27 +222,42 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
...
-		state.Warnf("%v", err)
-	}
-	for name, path := range files {
-		info.Files = append(info.Files, workshopFileToInfo(projectId, name, path))
+	// If the client queried everything available,
+	// we add workshop files to the response.
+	// Some of these may only exist as files, not instances.
+	// The "available" query is a best-effort version of "all":
+	// if something is wrong with the files we still return the instances.
+	if wstate == "all" || wstate == "available" {
```
dmitry-lyfar: We only have two valid states here and state is an optional field. Could you add the query parameters validation before the actual logic, so this condition would become not relevant, as it will be executed for any request.

dmitry-lyfar (changes requested):
Great work. Some minor issues.

jonathan-conder:
@akcano This should be ready in terms of changes to the docs at least. The main things are:
1. Removing workshop names from example commands (my rule of thumb was to do this only if the workshop definition file is listed earlier on).
2. Adding a tip about how this works for `exec` in the tutorial.
3. Adding nameless examples to the affected commands.

dmitry-lyfar (approved):

docs/reference/cli/workshop-stop.rst
```diff
@@ -33,3 +33,21 @@ Notes:
...
+
+
+Stop the nimble and jazzy workshops in the current project directory:
+
+.. code-block:: console
+
+   $ workshop stop nimble jazzy
+
+
+The name is optional if the project only has one workshop:
```
akcano:
```suggestion
The name is optional if the project has only one workshop:
```

Please apply globally.

docs/tutorial/index.rst
```diff
@@ -438,6 +438,18 @@ Next, build it *inside the workshop* using :ref:`exec <ref_workshop_exec>`:
 
    $ workshop exec golang go build main.go
 
+.. tip::
+
+   Since :samp:`golang` is the only workshop in the project,
+   it can be omitted from most :samp:`workshop` commands.
```
akcano:
```suggestion
   it can be omitted from most :command:`workshop` commands.
```

akcano (approved):
LGTM, a couple of minor changes.

PR #227, opened 2024-12-09
Doc: Improve tutorial, fix SDKcraft merge artefacts

akcano:


dmitry-lyfar (approved):
Yes and I like how the itemised list is used to explain the tasks output.

PR #229, opened 2024-12-12
Sketch SDK

dmitry-lyfar:
# Description

This PR updates the `sketch-sdk` (formerly, `hack`) workflow logic as follows:

```bash
$ workshop sketch-sdk           # Edits current Sketch SDK (opens an $EDITOR with a sketch template).
$ workshop sketch-sdk --stash   # Stashes; overwrites stash (if exists).
$ workshop sketch-sdk --restore # Restores from stash; fails if there is a current Sketch SDK.
$ workshop sketch-sdk --remove  # Removes current Sketch SDK.
```

`workshop remove` will remove both current and stashed `sketch` SDKs to be consistent with the command's approach to resources created by a workshop (e.g. the default mount directories will be removed as well if weren't `remount`).

## Breaking Changes (API)

- [Change /v1/workshops to return workshops and files separately](https://github.com/canonical/workshop/pull/229/commits/befd10497a336c2ede5948777acc231ec2392bd6).
- [Change /v1/projects/{id}/workshops to include file info if present](https://github.com/canonical/workshop/pull/229/commits/2cead7e336070fb770aed9f6f7f44bbbccd8c3b1).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+		Use:   "sketch-sdk [--stash|--restore|--remove] <WORKSHOP>",
+		Args:  cobra.ExactArgs(1),
+		Short: "Edit the sketch SDK and graft it onto the workshop",
+		Long: `
+This command opens the default text editor to configure the 'sketch' SDK
+and immediately installs it in the specified workshop,
+enabling rapid experiments and tweaks at the SDK level.
+
+Saving and exiting causes a refresh,
+which installs the updated 'sketch' SDK in the workshop.
```
akcano:
```suggestion
This opens the 'sketch' SDK definition in the default text editor,
enabling rapid experiments and tweaks at the SDK level.

Saving the definition and exiting the editor causes a refresh,
which installs the configured 'sketch' SDK in the workshop.
```

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+
+The '--stash' and '--restore' options stash the 'sketch' SDK,
+reversing the changes, and quickly restore it to the workshop.
+
+The '--remove' option removes the 'sketch' SDK permanently.
```
akcano:
```suggestion
The '--stash' and '--restore' options respectively stash the SDK,
reversing the changes, and quickly restore it to the workshop.
The '--remove' option removes the SDK permanently.
```

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+	cmd.MarkFlagsMutuallyExclusive("stash", "restore", "remove")
+
+	return cmd
+}
+
+var sketchTemplate = `# Sketch SDK for %s
+# Sketch SDK provides local customisation of this specific workshop.
+
+# To read more about SDKs, their components and syntax, see:
+# https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/
```
akcano, line 77: I'd suggest parameterizing the URLs and adding a `curl`-based test for them.

akcano (changes requested):

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+- The 'sketch' SDK doesn't appear in the workshop definition
+  and cannot include build-time data such as parts
+
+- In addition to hooks, the 'sketch' SDK can use interfaces,
+  define plugs, slots, connections and bindings
+
+- You can partially refresh the workshop, targeting the 'sketch' SDK
+  with the 'workshop refresh <WORKSHOP>/sketch' command
+`,
+		RunE: c.Run,
```
jonathan-conder, line 61: The `Examples` seem to have disappeared

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+    # apt-get install -y --no-install-recommends PACKAGE...
+    # snap install SNAP...
+
+  # EXAMPLE: check-health runs after all SDK setup completes, call 'workshopctl set-health okay' for OK.
+  # See https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/hooks/
+  # check-health: |
+    # if CHECK_HEALTH_COMMAND ; then
+    #   workshopctl set-health okay
+    # else
+    #   workshopctl set-health --code=installation-fails error "Installation fails"
```
jonathan-conder:
```suggestion
    #   workshopctl set-health --code=installation-failed error "Installation failed"
```

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+		storedir := sdk.WorkshopSketchSdkStash(user.HomeDir, p.Id, wp.Name)
+
+		if err = restoreSketch(sketchdir, storedir); err != nil {
+			return err
+		}
+
+		// Run refresh with the stored sketch SDK. We do not revert dirs exchange
+		// on a failed refresh here as it is run with the content from "stored"
+		// and with --wait-on-error. Hence, there is always a possibility to
+		// workshop refresh --abort and workshop sketch --restore to restore the
```
jonathan-conder:
```suggestion
		// workshop refresh --abort and workshop sketch-sdk --restore to restore the
```

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+		if err := removeSketch(sketchdir); err != nil {
+			return err
+		}
+
+		cmdrefresh := &CmdRefresh{root: c.root}
+		return cmdrefresh.Run(cmd, av[0:1])
+	}
+
+	metafile := filepath.Join(sketchdir, "meta", "sdk.yaml")
+	boilerplate := fmt.Sprintf(sketchTemplate, workshop.Filepath(p.Path, wp.Name), wp.Base)
```
jonathan-conder: With the new conventions, `workshop.Filepath` might not be correct here (it will always give `.workshop/<NAME>.yaml`)

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+	var rec workshop.SdkRecord
+	r := revert.New()
+	defer r.Fail()
+
+	if err := yaml.Unmarshal(content, &rec); err != nil {
+		return err
+	}
+
+	if rec.Name != sdk.Sketch {
+		return fmt.Errorf("cannot sketch: SDK name must be %q", sdk.Sketch)
```
jonathan-conder:
```suggestion
		return fmt.Errorf("cannot sketch: SDK must be named %q (now: %q)", sdk.Sketch, rec.Name)
```
For consistency with https://github.com/canonical/workshop/blob/c97e38cf668bb111f713c98fa61227a49db081d8/internal/workshop/project.go#L76

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+		r.Add(func() { os.RemoveAll(hooksdir) })
+	}
+	for _, hook := range []string{"setup-base", "save-state", "restore-state", "check-health"} {
+		hookpath := filepath.Join(hooksdir, hook)
+		if script := rec.Hooks[hook]; len(script) > 0 {
+			if err := os.WriteFile(hookpath, []byte(script), 0644); err != nil {
+				return err
+			}
+		} else {
+			if err := os.Remove(hookpath); err != nil && !osutil.IsDirNotExist(err) {
```
jonathan-conder:
```suggestion
			if err := os.Remove(hookpath); err != nil && !errors.Is(err, os.ErrNotExist) {
```
This confused me a bit because `hookpath` isn't a directory

cmd/workshop/sketch.go
```diff
@@ -0,0 +1,326 @@
...
+		return cmdrefresh.Run(cmd, []string{fmt.Sprintf("%s/sketch", av[0])})
+	}
+
+	if c.remove {
+		if err := removeSketch(sketchdir); err != nil {
+			return err
+		}
+
+		cmdrefresh := &CmdRefresh{root: c.root}
+		return cmdrefresh.Run(cmd, av[0:1])
```
jonathan-conder, line 246:
```suggestion
		return cmdrefresh.Run(cmd, av)
```
I think there's another line with this somewhere as well. I might change it in my PR anyway so feel free to leave as-is.

cmd/workshop/sketch_test.go
```diff
@@ -0,0 +1,399 @@
...
+		}
+	})
+}
+
+func (m *workshopSketch) TestSketchSdkMetaOnlySuccess(c *check.C) {
+	cmd := &CmdSketch{root: &CmdRoot{}}
+
+	m.mockSketchHappyRefreshPath(c, "ws/sketch", "wait-on-error")
+
+	sketchContent := fmt.Sprintf(sketchTemplate, workshop.Filepath(m.prjDir, "ws"), "ubuntu@22.04")
```
jonathan-conder: If you change the other `Filepath`, this one should change too

cmd/workshop/sketch_test.go
```diff
@@ -0,0 +1,399 @@
...
+		default:
+			c.Errorf("expected 5 calls, now on %d", n)
+		}
+	})
+
+	metadir, _ := m.mockMinimalSketchSdk(c, true, []byte(simpleSketchMeta))
+
+	err := cmd.Run(nil, []string{"ws"})
+	c.Assert(err, check.NotNil)
+
```
jonathan-conder, line 282: In https://github.com/canonical/workshop/commit/6c97ccbbd96f6f8e0a7956ee2e68c441eafcfbd0 there was a test similar to this which wasn't calling all the mocked endpoints, so I think it's a good idea to check that `n==5` here

cmd/workshop/sketch_test.go
```diff
@@ -0,0 +1,399 @@
...
+		return []byte(sketchContent), nil
+	})
+	defer restore()
+
+	err := cmd.Run(cmd.Command(), []string{"ws"})
+	c.Assert(err, check.IsNil)
+	c.Assert(filepath.Join(meta, "sdk.yaml"), testutil.FileEquals, sketchContent)
+	c.Assert(filepath.Join(hooks, "setup-base"), testutil.FileAbsent)
+	c.Assert(filepath.Join(hooks, "check-health"), testutil.FileAbsent)
+	c.Assert(filepath.Join(hooks, "save-state"), testutil.FileEquals, `# saves state`)
```
jonathan-conder: I think we should append newlines to hooks even if the user does something like `|-`. At least that's what I'm doing for scripts, and it would be good to be consistent here. I can add it as part of the scripts work if you prefer.

cmd/workshop/sketch_test.go
```diff
@@ -0,0 +1,399 @@
...
+	c.Assert(err, check.IsNil)
+
+	sketchContent := `name: sketch
+base: ubuntu@22.04
+plugs:
+  gpu:
+    interface: gpu
+`
+	restore := MockTextEditor(func(inPath string, inContent []byte) ([]byte, error) {
+		err := os.WriteFile(inPath, []byte(sketchContent), 0644)
```
jonathan-conder: I don't understand why this mock is writing the file. The other mocks just return the content.

docs/reference/cli/index.rst
```diff
@@ -24,7 +24,7 @@ such as :option:`!--help` or :option:`!-h`.
 .. include:: workshop-exec.rst
 
 
-.. include:: workshop-hack.rst
+.. include:: workshop-sketch-sdk.rst
```
jonathan-conder: This is no longer in alphabetical order - I suggest running `workshop generate-docs` and just copy the files over. This is the diff I see:
```diff
diff --color -Nru docs/reference/cli/index.rst docs-gendocs/index.rst
--- docs/reference/cli/index.rst	2024-12-13 11:59:59.809975659 +1300
+++ docs-gendocs/index.rst	2024-12-13 12:01:37.449277664 +1300
@@ -24,9 +24,6 @@
 .. include:: workshop-exec.rst
 
 
-.. include:: workshop-sketch-sdk.rst
-
-
 .. include:: workshop-info.rst
 
 
@@ -51,6 +48,9 @@
 .. include:: workshop-shell.rst
 
 
+.. include:: workshop-sketch-sdk.rst
+
+
 .. include:: workshop-start.rst
 
 
diff --color -Nru docs/reference/cli/workshop-refresh.rst docs-gendocs/workshop-refresh.rst
--- docs/reference/cli/workshop-refresh.rst	2024-12-13 11:59:59.809975659 +1300
+++ docs-gendocs/workshop-refresh.rst	2024-12-13 12:01:37.448277621 +1300
@@ -52,7 +52,7 @@
 
 - If the optional <SDK> is supplied,
   the operation is limited to this SDK;
-  currently, it can only be 'hack'
+  currently, it can only be 'sketch'
 
 
 .. rubric:: Options
@@ -106,8 +106,10 @@
    $ workshop refresh nimble --continue
 
 
-Refresh the :samp:`sketch` SDK in :samp:`nimble`:
+Refresh the hack SDK under 'nimble':
 
 .. code-block:: console
 
-   $ workshop refresh nimble/sketch
+   $ workshop refresh nimble/hack
+
+
diff --color -Nru docs/reference/cli/workshop-sketch-sdk.rst docs-gendocs/workshop-sketch-sdk.rst
--- docs/reference/cli/workshop-sketch-sdk.rst	2024-12-13 11:59:59.809975659 +1300
+++ docs-gendocs/workshop-sketch-sdk.rst	2024-12-13 12:01:37.448277621 +1300
@@ -54,3 +54,5 @@
 --stash
 
    Stash the sketch SDK and remove it from the workshop.
+
+
```
Looks like `cmd/workshop/refresh.go` still has some `hack` references

jonathan-conder (changes requested):
Looks good, some minor issues though

cmd/workshop/sketch.go
dmitry-lyfar: It turns out that the client cannot infer it anymore. See my latest commit, I went down the path of including the workshop's file path into the response for both, active and file-only workshops. It is a breaking change for the existing workshops that, I believe, we should do as the convention changes made reliance on the Filepath incorrect.

I also cleaned up the Filename function from the workshop package as it is incorrect too (made it unexported for the case it is currently used for). I think we should retire the Filepath function too as a separate change as it is mainly used in tests.

internal/workshop/workshop_file.go
```diff
@@ -140,10 +143,19 @@ func (p *SdkList) UnmarshalYAML(value *yaml.Node) error {
...
+	buf, err := os.ReadFile(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, fmt.Errorf("workshop definition %q not found", path)
+		}
+		return nil, err
+	}
+	file.Path = path
+
 	if err = yaml.Unmarshal(buf, &file); err != nil {
```
jonathan-conder: I think the path should be set after `Unmarshal`, or we should exclude `Path` from unmarshalling if that's possible

cmd/workshop/sketch.go
jonathan-conder: I was a bit hesitant to go down this path myself, I think the path for running workshops can potentially be out-of-date. In `v1GetProjectWorkshop`, when listing the files, I think we should update the paths for the existing workshops. That way it will only be out of sync if the request specifies `state=ready` or something.

I agree `Filepath` makes less sense now.

jonathan-conder (approved):

cmd/workshop/sketch.go
dmitry-lyfar: We have those links under auth wall; any possible ways to do link checks similarly to how we do it in the docs?

akcano: The most simple way now is to check the two parts:

- the base URL (`https://canonical-workshop.readthedocs-hosted.com/en/latest/`), which we are certain of and can trim during the test
- the path (`explanation/sdks/`), which can be verified inside the repo by probing the respective directory under `docs/`.

internal/workshop/project.go
```diff
@@ -104,40 +113,36 @@ func (w *Project) ReadWorkshops() (map[string]string, error) {
 }
 
 // Read single workshop file if it exists and is unique.
```
jonathan-conder: This comment should be adjusted since it doesn't read the file anymore

cmd/workshop/sketch.go
dmitry-lyfar: Thanks. Will be moving it to the backlog to implement the actual link checks for when the documentation is open.

PR #230, opened 2024-12-16
Doc: Add hack SDK to tutorial, fix issues

akcano:


docs/tutorial/index.rst
```diff
@@ -206,7 +218,7 @@ in the project directory:
    $ workshop list
 
      Project                Workshop   Status  Notes
-     ~/hello-workshop       golang     Off     -
+     ./hello-workshop       golang     Off     -
```
dmitry-lyfar: hmm, do we have a dot in the output there?

dmitry-lyfar: Just checked the last main, it is ~, if the project is in the user's home.

docs/tutorial/index.rst
```diff
@@ -511,7 +527,8 @@ named :samp:`golang/go:mod-cache`.
 As seen in the :command:`workshop info` output,
 it was automatically connected at :ref:`launch <tut_launch>`
 to the :samp:`golang/system:mount` slot,
-here abbreviated by convention as :samp:`:mount`.
+indicated by the ellipsis in the :samp:`host-source` path
```
dmitry-lyfar, line 530: thanks. It (the ellipsis) will go away in the latest info design changes, but it is good to have it here in the meantime.

docs/tutorial/index.rst
```diff
@@ -551,15 +568,129 @@ to a new location on the host:
 This makes :file:`/home/user/mod/` on the host
 act as the Go modules cache for the workshop.
 
-We're nearing the end of our tutorial;
-the only thing left is the cleanup.
+
+.. _tut_hack:
+
+Hack a workshop (optional)
```
dmitry-lyfar, line 574: It's OK for this iteration to have the hack as part of the tutorial as agreed. The question I wanted to discuss, no need to address it in this PR: how a user will continue from a tutorial?

I noticed this issue in our live sessions. The tutorial seemed a bit too long (they often take slightly more than 30 mins to navigate it and after that what should they continue with?

My thinking:

OK, I learnt basic concepts, I have a project, what do I do next? Supposedly, I move to a more advanced "tutorial" that will guide me through building an environment for the actual project via hacking.

dmitry-lyfar: Probably the issue we can address right now (in this PR) is: what's the next step? Preferably, a recommended action to continue with, so they don't have to choose (not ready for choosing in this very moment, just getting themselves familiar with the tool). It ends up training them to remove a workshop and that's it. From there they would wonder.

docs/tutorial/index.rst
```diff
@@ -551,15 +568,129 @@ to a new location on the host:
...
+the revision will be incremented.
+
+Another form of the :command:`workshop hack` command
+enables you to edit a specific :ref:`hook <exp_hooks>`:
+
+.. code-block:: console
+
+   $ workshop hack golang check-health
+
+This opens the :samp:`check-health` hook that |ws_markup| runs
```
dmitry-lyfar: Hacking is about customising the environment in the first place, this should be about setup-base. To continue on the topic of Go projects here, this is what I suggest for the setup-base:

```
apt-get update
apt-get install -y delve
apt-get install -y clang # if a Go project uses clang.
```

dmitry-lyfar (changes requested):
Very interesting to see hack in the tutorial. Left some comments, some of the hack commands are already outdated, but it is ok, we will change it all to the sketch-sdk in the following PR.

docs/tutorial/index.rst
akcano: I added a section for next steps. Also, my experience from our recent UX sessions tells me that we need to have a dedicated hack/sketch how-to guide.

As for the length, we're around the optimum, I think. However, adding major sections will change that.

docs/tutorial/index.rst
```diff
@@ -551,15 +568,127 @@ to a new location on the host:
...
+includes lines similar to the following:
+
+.. code-block:: console
+
+   content:
+     hack:
+       channel:  ~   2024-12-15  (x1)
+
+
+Note that the hack SDK lists the time of last update and the revision
```
dmitry-lyfar: It lists a local revision (starts with `x` as in snaps). The channel name is empty because it is a local SDK.

dmitry-lyfar (approved):
Looks good! I like the next action options. One bit about the local revisions / channels left.

Also, there is a spellcheck issue.

PR #231, opened 2024-12-16
Prevent connection removal on backend setup failure

lachypjones:
# Description
If an error occurs during the backend setup call within doDisconnect, the connection being disconnected will be left in a partially disconnected state. This error can be either transient (ie. a backend connection issue) or non-transient (ie. a broken workshop missing essential packages).

This PR restores the connection to the workshop state on failure, noting that this will not take any action to revert modifications to the workshop that were run during the disconnect (ie. remove<interface> functions).  

## One-way decisions
It was agreed that it's better to keep a potentially broken connection connected than to disconnect a potentially still-connected connection. Some of the factors going into this decision were the following:

- Transient errors 
  - It's better to let a user try again
- Ramifications for connect 
  - If we disconnect forcibly, it's highly possible we break any subsequent re-connections
- General guarantees
  - We should not allow an action (without making it explicitly clear) to break another action

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/ifacestate/handlers.go
+++ b/internal/overlord/ifacestate/handlers.go
@@ -652,8 +652,8 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err 
 
 	rev.Add(func() {
 		oldConn := &schema.ConnState{}
-		if err := task.Get("old-conn", oldConn); err != nil {
-			logger.Noticef("On doDisconnect: Cannot recover disconnected %q", cref.ID())
+		if err1 := task.Get("old-conn", oldConn); err1 != nil {
+			err = fmt.Errorf("On Disconnect: %w\nWhen attempting to revert: internal error: previous connection not found in state", err)
 		}
 		conns[cref.ID()] = oldConn
 		setConns(st, conns)
@@ -661,7 +661,7 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err 
 		_, err1 := m.repo.Connect(&cref, conn.StaticPlugAttrs, conn.DynamicPlugAttrs, conn.StaticSlotAttrs,
 			conn.DynamicSlotAttrs, nil)
 		if err1 != nil {
-			logger.Noticef("On doDisconnect: Cannot recover disconnected %q", cref.ID())
+			err = fmt.Errorf("On Disconnect: %w\nWhen attempting to revert: Cannot recover disconnected %q", err, cref.ID())
 		}
 	})
 
@@ -687,7 +687,7 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err 
 		ctx, cancel := handlersetup.BackendContext(tomb, user, ref.ProjectId)
 		defer cancel()
 		for _, backend := range m.repo.Backends() {
-			if err := backend.Setup(ctx, ref, m.repo); err != nil {
+			if err = backend.Setup(ctx, ref, m.repo); err != nil {
 				return err
 			}
 		}
--- a/internal/overlord/ifacestate/handlers.go
+++ b/internal/overlord/ifacestate/handlers.go
@@ -658,9 +658,10 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err 
 		}
 	})
 
-	oldConns := maps.Clone(conns)
+	oldConn := *conn
 	rev.Add(func() {
-		setConns(st, oldConns)
+		conns[cref.ID()] = &oldConn
+		setConns(st, conns)
 	})
 
 	switch {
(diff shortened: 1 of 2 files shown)


lachypjones:
Only unit tests have been implemented. The coverage report for doDisconnect indicates that nearly all functionality (with the exception of some error conditions that are not impacted by this change) is covered.

internal/overlord/ifacestate/handlers.go
```diff
@@ -658,6 +658,11 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err
 		}
 	})
 
+	oldConns := maps.Clone(conns)
```
dmitry-lyfar: Try change your `TestDoDisconnectSetupFailure` this way:

1. Set the "connect" task `t1` to be an auto-connection by `t1.Set("auto", true)`
2. Have this plug definiotion for `consumerManyPlugs`:
```
  plug-ssh:
    interface: mock-ssh-agent
    attribute: four
```

dmitry-lyfar (changes requested):

internal/overlord/ifacestate/handlers_test.go
```diff
@@ -1764,3 +1766,240 @@ func (s *interfaceHandlersSuite) TestUndoDiscardConnsSuccess(c *check.C) {
...
+	c.Assert(c2.Err(), check.ErrorMatches, `cannot perform the following tasks:
+-  \(setup failed\)`)
+
+	// Ensure that the connection is not removed from state
+	newConns := map[string]*schema.ConnState{}
+	s.state.Get("conns", &newConns)
+	c.Assert(oldConns, check.DeepEquals, newConns)
+
+	// Ensure the connection is not removed from the repo
+	ref, err := repo.Connected(s.prj.ProjectId, "ws", "consumer", "plug-ssh")
```
dmitry-lyfar: Instead of checking if the reference exists, try to get the Connection object from the repo and ensure that the connection remained the same. There is a similar issue with a reverting Connect that should use a copy of the old connection in the 'auto' connection state.

internal/overlord/ifacestate/handlers.go
```diff
@@ -658,6 +658,12 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err
 		}
 	})
 
+	oldConn := *conn
```
dmitry-lyfar: This solution solved the particular case for the auto, but it does not make full copy either. It happens to work because references to attrs are nullified, but if any of them was just changed, you'd get the same problem.

lachypjones: This consequence I did pick up on.
It didn't think it mattered in our case? I didn't see a situation where we would be modifying the map before the revert.

dmitry-lyfar: It works here, but if it did change, the intention to create a copy that is indicated by dereferencing would not work. The copy that is created is a shallow copy either way. So it is implicit. 

There is this code that executes earlier exactly for this reason:

```
	// store old connection for undo
	task.Set("old-conn", conn)
```

You can use that in your revert logic and get the connection (full copy of it) from the task. That would be explicit and future-proof.

Potentially, you can even combine the repo and state reverters into a single one, they follow each other.

internal/overlord/ifacestate/handlers.go
```diff
@@ -651,6 +651,11 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err
 	defer rev.Fail()
 
 	rev.Add(func() {
+		oldConn := new(schema.ConnState)
+		task.Get("old-conn", oldConn)
```
dmitry-lyfar: Error handling? Get returns an error.

internal/overlord/ifacestate/handlers.go
```diff
@@ -651,6 +651,11 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err
 	defer rev.Fail()
 
 	rev.Add(func() {
+		oldConn := new(schema.ConnState)
```
dmitry-lyfar: We don't use new elsewhere, so just create on the stack and give its address to Get.

internal/overlord/ifacestate/handlers.go
```diff
@@ -651,6 +651,13 @@ func (m *InterfaceManager) doDisconnect(task *state.Task, tomb *tomb.Tomb) (err
 	defer rev.Fail()
 
 	rev.Add(func() {
+		oldConn := &schema.ConnState{}
+		if err := task.Get("old-conn", oldConn); err != nil {
```
dmitry-lyfar: The lack of "old-conn" in the task's data should be an internal error. This is not a simple error, it's an integrity issue. So instead of proceeding and setting an empty connection to the state, we should return an internal error (see how these cases are handled a few lines above):

```
	if err := task.Get("forget", &forget); err != nil && !errors.Is(err, state.ErrNoState) {
		return fmt.Errorf("internal error: cannot read 'forget' flag: %w", err)
	}

```

dmitry-lyfar: In our case you don't have to check for ErrNoState, any error here is an internal error.

dmitry-lyfar (approved):

PR #233, opened 2024-12-18
Doc: Add coverage tool

akcano:
```
python coverage.py
open coverage.html
```


lachypjones:
It might be helpful to separate docs coverage from code coverage. 
Suggest considering moving coverage.py into the docs/ directory (and/or renaming)

dmitry-lyfar:
I like being able to say `@artefact`. Interesting in how you would approach topic artefacts from the original coverage map (that's category now) and, as discussed, simple coverage calculation which used to be a traffic light.

Would you imagine this tag to be used / updated by devs or by you during the coverage map review after an epic was finished?

akcano:
Closed to merge updates properly.

PR #236, opened 2024-12-20
Add `workshop run` command

jonathan-conder:
# Description

Adds scripts to workshop definition files and a command to `run` them.

At the moment these are only documented in the reference section. I think we should add a how-to guide for CI in future, and scripts would make sense to include there.

There's also a small change to `exec` - if the command returns a nonzero exit code, `workshop` will silently return the same code.

At the moment `workshop run foo` works (with no workshop name), but `workshop exec ./foo.sh` raises an error. At some point we might want to be more consistent about this. I'm on the fence but might make up my mind after using it a bit more.

The scripts are currently copied to `/var/lib/workshop/run/scripts` before running inside the container. There isn't yet an easy way of running them from `workshop shell`. I've attempted to pad the scripts with newlines so they line up with the workshop definition.

If the copy step fails, there's an issue where `gorilla/websocket` truncates the HTTP response body. This could lead to issues if our error messages get too long. For now I'm just ignoring the issue, because LXD does the same thing, and it's probably easier to fix upstream than here.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


akcano (approved):

internal/overlord/cmdstate/handlers.go
```diff
@@ -346,3 +349,42 @@ func (e *execution) controlLoop(execId string, conn *websocket.Conn, stop <-chan
...
+	var script string
+	st := task.State()
+	st.Lock()
+	err = task.Get("name", &name)
+	st.Unlock()
+	if err != nil {
+		return fmt.Errorf("cannot get script name for task %q: %w", task.ID(), err)
+	}
+	st.Lock()
+	err = task.Get("script", &script)
```
dmitry-lyfar: There are some known performance issues connected with that kind of serialisation in pebble (a frequent command with a fairly heavy set of attributes added to the state would make state.Unlock() longer every time which would cause lock contention). 

In this scenario, consider using `state.Cache` for the script's content, so it would be copied from memory into the workshop fs without intermediate json serialisation. That brings another issue -- the cache value would be lost if the daemon is restarted, so, the change would not be successful on a retry, but I think it is fine.

Alternatively, why don't we pull the script's content in this task's logic instead of doing it in the API layer?

dmitry-lyfar (commented):
Good stuff. I smoke tested it as well and it feels intuitive to use. See my notes on the `copy-script`.

internal/overlord/cmdstate/handlers.go
dmitry-lyfar: After thinking about it a bit, I tend to prefer doing it in the task's logic, the API layer does feel a bit alien to do that sort of logic (pulling the script's content, formatting the command, etc.)

jonathan-conder: OK, I think it makes sense to move it to the task. I don't remember why I did it this way initially but will see if anything crops up when I move it

jonathan-conder: The main issue isn't relevant now, but in future the workshop definition might modify the command more, e.g.

```yaml
scripts:
  build:
    shell: python
    env:
      FOO: BAR
    script: |
      print('hello world')
```

If the `copy-script` task is in charge of reading the workshop file, it will need some way to propagate this info to the `exec` task. I don't see any existing pattern for this, so I think it makes sense to use `state.Cache` for that as well. So `copy-script` will take an `ExecArgs` from the cache, change the command, and then cache the result for `exec` to use. That way the `exec` handler can remain unaware of scripts.

It should make things easier in future. Although there might be a minor issue: if the definition specifies some options, I think CLI arguments should override them. But the task can't tell the difference between `--uid 1000 --gid 1000` and no arguments. Doesn't really change my preference here though.

jonathan-conder: Should be working now. I added this and https://github.com/canonical/pebble/pull/478 as 2 new commits, but also rebased on `main` and amended an earlier commit to fix the example commands for `workshop run`.

dmitry-lyfar:
Smoke tested performance of `workshop run` vs. `workshop exec` does not indicate a significant difference on average. The exec is slightly faster in most runs due to the lack of copy-script, apparently. 

```bash
#!/usr/bin/bash

for i in {1..50}; do
   workshop run info > /dev/null 2>&1 &
done

wait
echo "Done."

# ./perf.sh  0.22s user 0.35s system 3% cpu 15.357 total
```

vs.

```bash
#!/usr/bin/bash

for i in {1..50}; do
   workshop exec -- lsb_release -a > /dev/null 2>&1 &
done

wait
echo "Done."

# ./perf_exec.sh  0.21s user 0.32s system 3% cpu 15.214 total
```

PR #237, opened 2024-12-20
Enable default linters except for errcheck

jonathan-conder:
# Description

This enables all but one of the default linters for `golangci-lint` and also `nilerr`, which checks for patterns like this:
```go
if err != nil {
    return nil
}
```
I already fixed a few manually but this caught another one, so I think it's worth keeping it on.

The `errcheck` linter triggers a lot of errors, so it will take some time to fix them all, if we want to enable it. Also `staticcheck` has some false positives, but mostly for the `state` tests because `Lock()` and `Unlock()` have side effects. There's also an issue with `rand.Seed` which I can't really fix until `x-go/randutil` does.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


dmitry-lyfar (approved):
Beautiful, thank you.

PR #239, opened 2025-01-06
Add code coverage action

lachypjones:
# Description
This PR adds go cover profile generation to unit and spread tests, surfacing this information in the readme (for main) and in an automatically-updating comment for PRs. 

Two outputs are generated:
- A traditional 'badge' showing coverage as an overall %
- The go cover html representation 

For a pull-request, these will appear as a comment created in the PR which includes the badge inline and a link to the html output. Subsequent commits will update this comment. Note, due to limitations with the upload-artifact action, this will download as a zip. 

For the main branch, the badge will appear at the top of the README. Clicking the badge will link to the raw HTML output. 

## Notes on coloring
Currently this matches the default coloring of [codecov](https://docs.codecov.com/docs/coverage-configuration). 
That is values below 60% are red, values over 80% are green, and values between these are yellow. 

# Breaking Changes
- Spread and unit tests now run as a callee of the coverage checks. This is done to make sharing artifacts between the (previously separate) workflows as stable as possible. There were alternatives, however most interacted directly with the GH cli and as such felt quite brittle. 
- The wiki is used to store artifacts used by the README. As we host out docs via RTD, it was felt that this shouldn't be an issue. If at a later date we wish to incorporate the wiki, alternative hosting for the artifacts will need to be found (it is also likely this would coincide the the project being released, at which point we could use codecov or something similar)

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -37,10 +37,10 @@ Environment setup
       workshopd run --create-dirs
 
    The client can connect using the daemon's Unix domain socket:
-   
+
    .. code-block:: console
 
-      
+
       export WORKSHOP=~/workshop
       workshop list
 
@@ -214,6 +214,39 @@ of ``Spread``:
 .. code-block:: console
 
    spread tests/<TestPathName>
+
+
+To check code coverage:
+
+.. code-block:: console
+
+  go test --coverpkg=<./...|package> covermode=<set|count|atomic> -coverprofile=<OutputFile> <./...|package>
+
+  // For example to measure coverage using all tests:
+  go test covermode=count -coverprofile=cover.out ./...
+
+
+To generate a html representation:
+
+.. code-block:: console
+
+  go tool cover -html=<OutputFile> -o <OutputHTML>
+
+  // For example:
+  go tool cover -html=cover.out -o cover.html
+
+  // The output flag can be omitted to open in the default browser:
+  go tool cover -html=cover.out
+
+
+The above will work for unit and integration tests instrumented directly with
+`go test`. Integration tests run using spread will create the coverprofile
+automatically, however the artifacts will need to be collected from the VM.
+This can be accomplished by using the `-artifacts` flag when running spread.
+
+.. code-block:: console
+
+  spread -artifacts=<path-to-dest> tests/integration/
 
 
 How to run a local SDK Store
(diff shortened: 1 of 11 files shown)


docs/contributing.rst
```diff
@@ -216,6 +216,39 @@ of ``Spread``:
...
 
 
+To check code coverage:
+
+.. code-block:: console
+
+  go test --coverpkg=<./...|package> covermode=<set|count|atomic> -coverprofile=<OutputFile> <./...|package>
+
+  // For example to measure coverage using all tests:
+  go test covermode=count -coverprofile=cover.out ./...
```
akcano: Three-space indent here and below; the first character should be vertically aligned with the 'c' in `.. code-block`.

docs/contributing.rst
```diff
@@ -216,6 +216,39 @@ of ``Spread``:
...
+
+
+To generate a html representation:
```
```suggestion
To generate an HTML representation:
```

docs/contributing.rst
```diff
@@ -216,6 +216,39 @@ of ``Spread``:
...
+  go tool cover -html=cover.out -o cover.html
+
+  // The output flag can be omitted to open in the default browser:
+  go tool cover -html=cover.out
+
+
+The above will work for unit and integration tests instrumented directly with
+`go test`. Integration tests run using spread will create the coverprofile
+automatically, however the artifacts will need to be collected from the VM.
+This can be accomplished by using the `-artifacts` flag when running spread.
```
```suggestion
`go test`. Integration tests run using `spread` will create the coverprofile
automatically, however the artifacts will need to be collected from the VM.
This can be accomplished by using the `-artifacts` flag when running `spread`.
```

akcano (changes requested):

docs/contributing.rst
lachypjones: @akcano you lead me down a very interesting grammatical rabbit hole with this one! 
https://english.stackexchange.com/questions/468475/should-one-use-an-or-an-a-before-an-acronym-beginning-with-the-letter-h

I'm accepting the suggestion, I believe you are correct for the majority case of 'aitch'.

dmitry-lyfar (approved):

dmitry-lyfar (approved):

PR #240, opened 2025-01-08
Doc: Update tutorial, improve formatting

akcano:


tests/documentation/tutorial/task.yaml
```diff
@@ -54,5 +54,7 @@ execute: |
   workshop_exec remount nimble/test-sdk-go:mod-cache ~/mod
   workshop_exec info
 
+  # TODO: sketch-sdk
```
dmitry-lyfar: This can be done using the same approach as in `tests/sketch-sdk/task.yaml`:

To edit `setup-base`:

```
  sudo TERM=ansi EDITOR=ex -u ubuntu -- workshop sketch-sdk <<EOF
  10
  i
    setup-base: |
      apt-get update
      apt-get install -y --no-install-recommends delve
      apt-get install -y --no-install-recommends clang
  .
  wq
  EOF
```

To add a plug interactively (plugs go after hooks, so it would be easier to do this way in the test):

```
  sudo TERM=ansi EDITOR=ex -u ubuntu -- workshop sketch-sdk <<EOF
  $
  a
    ssh-agent:
  .
  wq
  EOF
```

dmitry-lyfar (approved):

PR #250, opened 2025-01-09
Don't propogate a non-zero exit code to the host when running 'shell'

lachypjones:
# Description
Small PR that stops the propagation of exit codes to the host when running the `shell` command.

As a shell is inherently interactive (and therefore surfaces all errors during a session to the end-user), there is no need to re-alert the user that a command they attempted to run failed.   

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -158,7 +158,7 @@ func maybeNameAndCommand(cmd *cobra.Command, av []string) error {
 func (c *CmdExec) Run(cmd *cobra.Command, av []string) error {
 	// Infer workshop name if first positional argument is --
 	if cmd.ArgsLenAtDash() == 0 {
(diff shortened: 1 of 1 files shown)


jonathan-conder:
Just FYI I did something similar in 5b558b5bcd39eaa65947daa7a638243e7a3af496 (part of https://github.com/canonical/workshop/pull/236). I think it makes sense to remove the error message (rather than the code) for all 3 of `exec`, `shell` and `run`. Not so keen on making a special case for `shell` though.

lachypjones:
Closed in favor of Jonathan's changes

PR #254, opened 2025-01-12
Fix autostart behavior

lachypjones:
# Description
This PR does two things:

1. Changes the autostart behavior so that if a container was last in the `stopped` state, it won't automatically start it on subsequent boots. 
2. Uses an lxd proxy device in place of a bind mount for the untrusted workshop socket. This fixes an issue where a workshop would be unable to start on boot if the workshop daemon had not started quickly enough. 

# Breaking Changes
Workshops will need to be recreated to use the new proxy device. Workshops using the old mount will still function. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/ifacestate/ifacemgr.go
+++ b/internal/overlord/ifacestate/ifacemgr.go
@@ -300,21 +300,11 @@ func (m *InterfaceManager) ResolveDisconnect(
 	}
 }
 
-// Ensure the mounts required by a workshop to function properly were created:
-// workshopctl, socket. These mounts are created at the time of launch but can
-// become invalid on the daemon restart / update. Thus, recreating them upon
-// every daemon restart makes sure they still point to the correct files.
+// Ensure the workshopctl mount (required by a workshop to function properly)
+// is created. This mount is created at the time of launch but can become
+// invalid on the daemon restart / update. Thus, recreating it upon every
+// daemon restart makes sure it still points to the correct files.
 func (m *InterfaceManager) recreateInternalMounts(pctx context.Context, w string) error {
-	hostpath := dirs.SocketPath + ".untrusted"
-	sname := filepath.Base(hostpath)
-	wspath := filepath.Join(dirs.WorkshopRunDir, sname)
-	socket := workshop.Mount{Name: "workshop.socket", What: hostpath, Where: wspath}
-
-	_ = m.backend.RemoveWorkshopMount(pctx, w, socket.Name)
-	if err := m.backend.AddWorkshopMount(pctx, w, socket); err != nil {
-		return err
-	}
-
 	// Recreate workshopctl bind mount, this has to be done if, for example,
 	// workshopctl was updated to a new version and is shown as /deleted in a
 	// workshop.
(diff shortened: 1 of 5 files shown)


akcano:
A side note; please tick the appropriate doc boxes in the PR description next time so I can see the relevant PRs.

lachypjones:
@akcano 
This PR was (is) still in draft. With the exception of checking rough direction and change surface area, I wouldn't expect you to check these often. It's still a WIP so the relevant boxes are not checked. 

For example, other than a quick grep to check for mentions of any existing workshop autostart behavior (of which there is none), I haven't yet manually searched the documentation for any related topics. So at this stage I'm unsure if this will affect docs or not.

Hope that makes sense.

akcano:
yes, absolutely, thanks for the elaboration. I was just cautious that this may trickle farther

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -300,21 +300,11 @@ func (m *InterfaceManager) ResolveDisconnect(
...
-	hostpath := dirs.SocketPath + ".untrusted"
-	sname := filepath.Base(hostpath)
-	wspath := filepath.Join(dirs.WorkshopRunDir, sname)
-	socket := workshop.Mount{Name: "workshop.socket", What: hostpath, Where: wspath}
-
-	_ = m.backend.RemoveWorkshopMount(pctx, w, socket.Name)
-	if err := m.backend.AddWorkshopMount(pctx, w, socket); err != nil {
-		return err
-	}
-
```
lachypjones, line 317: We could optionally leave this behavior in place (with a check to see if the mount exists) and fully remove it at a later date. This would then convert existing mounts to the new proxy device (fixing the breaking change).

dmitry-lyfar: Currently, everyone installs workshop from a local snap which does not have an update option. It will execute the remove hook first which would delete the existing workshops and install a new snap. This helps us to get breaking changes through at the cost of some inconvenience of relaunching workshops.

Later on, when the snap is published, such breaking changes would have to be done with something like this in mind.

tests/main/autostart/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+  if [ $SPREAD_REBOOT = 0 ]; then
+    REBOOT
+  fi
+  if [ $SPREAD_REBOOT = 1 ]; then
+    workshop_exec list | MATCH 'Ready'
+  fi
+
+  echo "Make sure a workshop can start independently of the workshop daemon"
+  if [ $SPREAD_REBOOT = 1 ]; then
+    echo "snapctl stop --disable workshop.workshopd" | snap run --shell workshop
```
dmitry-lyfar: Why not:

```
snap stop --disable workshop
```
?

tests/main/autostart/task.yaml
```diff
@@ -0,0 +1,36 @@
...
+    REBOOT
+  fi
+  if [ $SPREAD_REBOOT = 2 ]; then
+    echo "snapctl start --enable workshop.workshopd" | snap run --shell workshop
```
dmitry-lyfar: Same as above.

tests/main/autostart/task.yaml
lachypjones: This is better. I'll change now.

dmitry-lyfar (approved):

PR #256, opened 2025-01-14
Workshop definition lists SDKs in installation order

jonathan-conder:
# Description

This is a breaking change, workshop definitions will need to be updated (see the diff for many examples). Running workshops should be removed before running the server with these changes.

If an old workshop is still running, you may see an error like this when trying to remove it:
```
cannot load workshop: yaml: unmarshal errors:
  line 4: cannot unmarshal !!map into []workshop.SdkRecord
```
You can either remove all workshops using [this script](https://github.com/canonical/workshop/blob/main/snap/local/hooks/remove), or revert 67decf66f63e537cb96db00c6c6141b8c0c4a066 and remove or refresh the old workshops manually.

Another potentially breaking change is that lists of SDKs and mounts returned from the API will no longer be sorted by default. At the moment this is done client-side, but in future we might provide extra arguments for clients to opt-in to server-side sorting. The list of SDKs has been renamed from `content` to `sdks`.

Finally, the uniqueness of SDK names is now only validated when the workshop file is read (not when it's deserialized from LXD). This is consistent with other types of validation, i.e. if someone wants to mess with the LXD metadata they were already able to do that in other ways. The SDKs are now stored in LXD using `user.workshop.sdks` rather than `user.workshop.content`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


dmitry-lyfar (approved):
Great. I'll wait for the "content -> sdks" change in the info output before merging.

jonathan-conder:
I went through and removed all the references to `content` I could find. This includes some old references to the mount interface. Introduces further breaking changes in the API (content renamed to SDKs) and LXD config keys (`user.workshop.content` is now `user.workshop.sdks`). I'll update the description

docs/reference/definitions/workshop-definition.rst
```diff
@@ -79,8 +79,6 @@ and includes a number of mandatory and optional keys:
        are set by default.
 
 
-Any entry in :samp:`sdks` must be named after an existing SDK
-that is available from the SDK store.
```
akcano, line 83: I would add an admonition of sorts:

```
The SDKs are installed in the order they appear in the :samp:`sdks` list.
```

```

PR #257, opened 2025-01-14
Add 'read-only' option for mounts

lachypjones:
# Description
Adds a 'read-only' parameter for mounts. This enables the enforcement of read-only access to workshop mounts from within a workshop. 

Syntax:
```
plugs:
  example_plug:
    interface: mount
    workshop-target: <target>
    [read-only: <true, false, "true", "false">] 
```

The `read-only` parameter accepts a boolean or any string representation thereof (ie. "true", "false", "TRUE", etc)

SDKcraft validation implemented in https://github.com/canonical/sdkcraft/pull/139

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/explanation/interfaces/mount-interface.rst
+++ b/docs/explanation/interfaces/mount-interface.rst
@@ -36,7 +36,8 @@ which is declared in the SDK definition.
 
 A basic structure would include the name of the plug itself,
 the interface (:samp:`mount`)
-and the intended target path inside the workshop (:samp:`workshop-target`).
+the intended target path inside the workshop (:samp:`workshop-target`)
+and, optionally, whether the mount should be read-only (:samp:`read-only`).
 
 Defining the plug in an SDK designates the target directory inside the workshop;
 a directory on the host system that |ws_markup| will create at run-time
--- a/docs/explanation/sdks/data-persistence-sharing.rst
+++ b/docs/explanation/sdks/data-persistence-sharing.rst
@@ -72,6 +72,7 @@ per each directory you want to retain during the workshop's life cycle.
      training-data:
        interface: mount
        workshop-target: /opt/training
+       read-only: true
 
 
 This SDK defines two mount plugs;
@@ -80,6 +81,10 @@ for each,
 Both :samp:`workshop-target` directories inside the workshop
 can be used by the SDK-specific logic
 implemented via :ref:`hooks <exp_sdk_hooks>` and other features.
+
+Additionally, you can mark a directory as `read-only`.
+|ws_markup| will then enforce the immutability of resources in this directory
+when they are accessed from inside the workshop.
 
 Here's a corresponding workshop definition:
 
--- a/internal/interfaces/builtin/mount.go
+++ b/internal/interfaces/builtin/mount.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/canonical/workshop/internal/interfaces"
@@ -62,7 +63,7 @@ const mountBaseDeclarationPlugs = `
           auto-explicit: true
 `
 
-var knownPlugAttributes = []string{"workshop-target"}
+var knownPlugAttributes = []string{"workshop-target", "read-only"}
 var knownSlotAttributes = []string{"workshop-source", "host-source"}
 
 // mountInterface allows sharing content between sdks
@@ -98,12 +99,32 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
 			return fmt.Errorf(`unknown attribute for mount interface plug: %q`, name)
 		}
 	}
+
 	target, ok := plug.Attrs["workshop-target"].(string)
 	if !ok || len(target) == 0 {
 		return fmt.Errorf("mount plug must contain target path")
 	}
+
 	if err := validatePath(target); err != nil {
 		return err
+	}
+
+	ro, ok := plug.Attrs["read-only"]
+	if !ok {
+		ro = false
+	}
+
+	switch ro := ro.(type) {
+	case bool:
+		plug.Attrs["read-only"] = ro
+	case string:
+		roBool, err := strconv.ParseBool(ro)
+		if err != nil {
+			return fmt.Errorf(`unknown value %q in key "read-only" for mount interface plug. Accepted values are 'true' or 'false'. String representations (ie. '"true"') are also permitted.`, ro)
+		}
+		plug.Attrs["read-only"] = roBool
+	default:
+		return fmt.Errorf(`unknown value type %T in key "read-only" for mount interface plug. Accepted types are 'bool' or 'string'.`, ro)
 	}
 	return nil
 }
@@ -141,6 +162,12 @@ func (iface *mountInterface) target(attrs interfaces.Attrer) string {
 		return target
 	}
 	return ""
+}
+
+func (iface *mountInterface) readOnly(attrs interfaces.Attrer) bool {
+	var ro bool
+	attrs.Attr("read-only", &ro)
+	return ro
 }
 
 func (iface *mountInterface) workshopSource(slot *interfaces.ConnectedSlot) (string, error) {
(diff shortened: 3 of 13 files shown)


jonathan-conder:
I feel like `mode` isn't quite the right term for this, although I'm blanking on a better one. Maybe just a `readonly` boolean like in LXD.

For example `fat` filesystems have the `fmask` and `dmask` options which I think are closer to `mode`. Or we might reserve `mode` for the option `X-mount.mkdir[=mode]`.

internal/interfaces/builtin/mount.go
```diff
@@ -98,10 +99,30 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+	}
+	// Check if 'read-only' is a boolean
+	if _, ok := ro.(bool); !ok {
+		// If not, check whether read only is a string that can be interpreted as
+		// a boolean. We don't care if the string typecast fails, either way
+		// it's an invalid value
+		roString, _ := ro.(string)
+		roBool, err := strconv.ParseBool(roString)
+		if err != nil {
+			return fmt.Errorf(`unknown value %q in key "read-only" for mount interface plug. Accepted values are 'true' or 'false'. String representations (ie. '"true"') are also permitted.`, roString)
```
lachypjones: This error does not match the 'content interface' errors, this is done deliberately to be compatible with Jonathan's upcoming changes.

internal/interfaces/builtin/mount.go
```diff
@@ -143,6 +164,13 @@ func (iface *mountInterface) target(attrs interfaces.Attrer) string {
 	return ""
 }
 
+func (iface *mountInterface) readonly(attrs interfaces.Attrer) bool {
+	var ro bool
+	fmt.Println(strconv.FormatBool(ro))
```
dmitry-lyfar: ?

docs/explanation/interfaces/mount-interface.rst
```diff
@@ -36,7 +36,8 @@ which is declared in the SDK definition.
 
 A basic structure would include the name of the plug itself,
 the interface (:samp:`mount`)
-and the intended target path inside the workshop (:samp:`workshop-target`).
+the intended target path inside the workshop (:samp:`workshop-target`)
+and optionally whether the mount should be read only (:samp:`read-only`).
```
akcano:
```suggestion
and, optionally, whether the mount should be read-only (:samp:`read-only`).
```

docs/explanation/sdks/data-persistence-sharing.rst
```diff
@@ -81,6 +82,10 @@ Both :samp:`workshop-target` directories inside the workshop
 can be used by the SDK-specific logic
 implemented via :ref:`hooks <exp_sdk_hooks>` and other features.
 
+Additionally, you can mark a directory as `read-only`.
+|ws_markup| will then enforce the immutability of resources in this directory
+when accessed from inside the workshop.
```
akcano:
```suggestion
when they are accessed from inside the workshop.
```

akcano (changes requested):

internal/interfaces/lxd_device/backend.go
```diff
@@ -171,8 +178,9 @@ func runMountCommand(conn lxd.InstanceServer, pid, w string, cmd []string) error
 
 func reloadMounts(conn lxd.InstanceServer, pid, w string) error {
 	return runMountCommand(conn, pid, w, []string{
-		"mount",
-		"-a",
+		"/bin/bash",
+		"-c",
+		"systemctl daemon-reload && systemctl restart local-fs.target",
```
dmitry-lyfar, line 188: The workarounds that I've seen so far only mention `daemon-reload`. Do we need to restart the target?

lachypjones: From our discussion this morning. The target does need to be restarted. This action is tightly scoped to the entries in fstab: https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html#local-fs.target

internal/interfaces/builtin/mount.go
```diff
@@ -98,10 +99,30 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+	}
+	// Check if 'read-only' is a boolean
+	if _, ok := ro.(bool); !ok {
+		// If not, check whether read only is a string that can be interpreted as
+		// a boolean. We don't care if the string typecast fails, either way
+		// it's an invalid value
+		roString, _ := ro.(string)
+		roBool, err := strconv.ParseBool(roString)
+		if err != nil {
+			return fmt.Errorf(`unknown value %q in key "read-only" for mount interface plug. Accepted values are 'true' or 'false'. String representations (ie. '"true"') are also permitted.`, roString)
```
akcano:
```suggestion
			return fmt.Errorf(`unknown value %q in key "read-only" for mount interface plug. Accepted values are 'true' or 'false'. String representations (e.g., '"true"') are also permitted.`, roString)
```

akcano (changes requested):

internal/interfaces/builtin/mount.go
```diff
@@ -98,10 +99,30 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+		}
+		// Update value so we don't need to convert later
+		plug.Attrs["read-only"] = roBool
+	}
+
```
dmitry-lyfar: this separates `validatePath(target)` call and parsing the `target` attribute. It should finish with target first and then move on to the next attribute.

internal/interfaces/builtin/mount.go
```diff
@@ -98,10 +99,30 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+	ro, ok := plug.Attrs["read-only"]
+	if !ok {
+		ro = false
+	}
+	// Check if 'read-only' is a boolean
+	if _, ok := ro.(bool); !ok {
+		// If not, check whether read only is a string that can be interpreted as
+		// a boolean. We don't care if the string typecast fails, either way
+		// it's an invalid value
+		roString, _ := ro.(string)
```
dmitry-lyfar: Consider:

```
	switch ro := ro.(type) {
	case bool:
		plug.Attrs["read-only"] = ro
	case string:
		roBool, err := strconv.ParseBool(ro)
		if err != nil {
			return fmt.Errorf(`unknown value %q in key "read-only" for mount interface plug. Accepted values are 'true' or 'false'. String representations (e.g., "true") are also permitted.`, ro)
		}
		plug.Attrs["read-only"] = roBool
	default:
		return fmt.Errorf(`unknown value type %T in key "read-only" for mount interface plug. Accepted types are 'bool' or 'string'.`, ro)
	}
```

dmitry-lyfar: This is how it is (mostly) done elsewhere if there are multiple possible types, so I think it should be like this here.

internal/interfaces/builtin/mount.go
```diff
@@ -143,6 +164,12 @@ func (iface *mountInterface) target(attrs interfaces.Attrer) string {
 	return ""
 }
 
+func (iface *mountInterface) readonly(attrs interfaces.Attrer) bool {
+	var ro bool
+	attrs.Attr("read-only", &ro)
```
dmitry-lyfar, line 169: Please rebase to the main, this will likely be picked up by the linter.

internal/interfaces/lxd_device/backend.go
dmitry-lyfar: @lachypjones I think we need to link it in the comments. Although, reading what is in the link, how does it mandate the target reload? The fstab generator adds it as a local mount dependency, but do you have to restart the target?

internal/interfaces/lxd_device/backend.go
```diff
@@ -352,10 +360,16 @@ func (b *Backend) Setup(ctx context.Context, sdkInfo sdk.Ref, repo *interfaces.R
 
 	reload := false
 	for _, mnt := range spec.Profile.Mounts {
-		if reload, err = installMount(user, fs, mnt); err != nil {
+		maybeReload, err := installMount(user, fs, mnt)
+		if err != nil {
 			return err
 		}
+
+		if maybeReload {
```
dmitry-lyfar: This is fine, but the name is confusing a bit. We use maybe prefix to indicate that something may happen, here the reload is a must, so it is not a maybe.

internal/interfaces/lxd_device/tests/integration/backend_test.go
```diff
@@ -154,6 +155,16 @@ func (f *backendDeviceSuite) TestSetupWorkshopMounts(c *check.C) {
 	c.Assert(f.repo.AddSdk(cinfo), check.IsNil)
 	c.Assert(f.repo.AddSdk(pinfo), check.IsNil)
 
+	// A mountpoint for /media is required as it's configured as a mount
+	// dependency for workshop-workshop mounts.
+	fs, err := f.be.WorkshopFs(f.ctx, "test")
```
dmitry-lyfar: This can be done by adjusting lxd/tests/helper:

```
func DefaultTestDevices() map[string]map[string]string {
	return map[string]map[string]string{
		"root":             {"type": "disk", "pool": "default", "path": "/"},
		"workshop.network": {"type": "nic", "network": "lxdbr0", "name": "eth0"},
                "workshop.project": {...}
	}
}
```

or creating another defaultTestDevices for this specific test suite.

dmitry-lyfar (approved):

PR #264, opened 2025-01-16
Mount SDKs as volumes

dmitry-lyfar:
# Description

Managing SDKs as read-only volumes is required to support restoring from a previous workshop snapshot for `workshop refresh`. An SDK is attached to a workshop(s) as a read-only volume before running other tasks. 

One caveat to consider is that the system will have as many versions of an SDK volume as many revisions are used by the running workshops at the time. Without a zfs deduplication or compression that suggests unwanted duplication assumingan average SDK does not have significant content differences amongst its revisions. The possible duplication issue will not be solved under this scope. Based on the assumed use cases, it can be tolerated or solved with zfs
de-duplication in the following Workshop releases.

On another implementation note, Workshop has to transform SDKs into an acceptable LXD backup tar format to import them as volumes. This is a temporary solution and requires an extension for the LXD API to be able to import an arbitrary tarball into a custom volume without attaching it to a workshop or using the backup functionality. 

## Managing zfs storage

Workshop will create and use a zfs loop-backend using LXD API. This is a 30 GB non auto-expandable zfs pool with compression and trimming enabled by default. Thus, extending zfs pool needs addressing in the documentation: https://documentation.ubuntu.com/lxd/en/stable-4.0/storage/#growing-a-loop-backed-zfs-pool. (cc @akcano where would you suggest it to go?).

## Breaking changes

- Workshop starts using "zfs" LXD storage backend. That requires relaunch of the existing workshops.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/daemon/api_workshops_test.go
+++ b/internal/daemon/api_workshops_test.go
@@ -743,10 +743,10 @@ func mockSdkVolume(sdkdir string, metayaml string) error {
 
 func (s *apiSuite) mockSdkVolumes(c *check.C, sdks map[string]testSdk) {
 	for _, info := range sdks {
-		err := s.b.CreateVolume(s.ctx, info.s.VolumeName())
+		err := s.b.CreateVolume(s.ctx, sdk.VolumeName(info.s.Name, info.s.Revision.String()))
 		c.Assert(err, check.IsNil)
 
-		vfs := s.b.WorkshopVolumeContents[info.s.VolumeName()]
+		vfs := s.b.WorkshopVolumeContents[sdk.VolumeName(info.s.Name, info.s.Revision.String())]
 		err = mockSdkVolume(vfs, info.meta)
 		c.Assert(err, check.IsNil)
 	}
@@ -1953,8 +1953,8 @@ func (s *apiSuite) TestSDKInstallationOrder(c *check.C) {
 
 	c.Assert(s.b.AttachVolumeCalls, check.DeepEquals, []fakebackend.AttachVolumeCall{
 		{Workshop: "manysdks", Name: workshop.AptCacheVolumeName("manysdks", s.project.ProjectId)},
-		{Workshop: "manysdks", Name: s2.VolumeName()},
-		{Workshop: "manysdks", Name: s1.VolumeName()},
+		{Workshop: "manysdks", Name: sdk.VolumeName(s2.Name, s2.Revision.String())},
+		{Workshop: "manysdks", Name: sdk.VolumeName(s1.Name, s1.Revision.String())},
 	})
 	s.b.AttachVolumeCalls = s.b.AttachVolumeCalls[:0]
 
@@ -1977,8 +1977,8 @@ func (s *apiSuite) TestSDKInstallationOrder(c *check.C) {
 
 	c.Assert(s.b.AttachVolumeCalls, check.DeepEquals, []fakebackend.AttachVolumeCall{
 		{Workshop: "manysdks", Name: workshop.AptCacheVolumeName("manysdks", s.project.ProjectId)},
-		{Workshop: "manysdks", Name: s1.VolumeName()},
-		{Workshop: "manysdks", Name: s2.VolumeName()},
+		{Workshop: "manysdks", Name: sdk.VolumeName(s1.Name, s1.Revision.String())},
+		{Workshop: "manysdks", Name: sdk.VolumeName(s2.Name, s2.Revision.String())},
 	})
 }
 
--- a/internal/daemon/api_workshops_test.go
+++ b/internal/daemon/api_workshops_test.go
@@ -14,16 +14,18 @@ import (
 	"sync"
 	"time"
 
-	"github.com/spf13/afero"
 	"gopkg.in/check.v1"
 
 	"github.com/canonical/workshop/internal/dirs"
 	"github.com/canonical/workshop/internal/interfaces"
+	"github.com/canonical/workshop/internal/osutil"
 	"github.com/canonical/workshop/internal/overlord/conflict"
 	"github.com/canonical/workshop/internal/overlord/state"
+	"github.com/canonical/workshop/internal/progress"
 	"github.com/canonical/workshop/internal/sdk"
 	"github.com/canonical/workshop/internal/testutil"
 	"github.com/canonical/workshop/internal/workshop"
+	"github.com/canonical/workshop/internal/workshop/fakebackend"
 )
 
 var (
@@ -268,16 +270,22 @@ plugs:
 `
 )
 
-var testsdks = map[string]string{
-	"test-sdk":   testsdk,
-	"test-sdk-2": testsdk2,
+type testSdk struct {
+	s    sdk.Setup
+	meta string
 }
 
-func (s *apiSuite) launchWorkshop(c *check.C, name, yaml string, sdks map[string]string) {
+var apiSuiteSdks = map[string]testSdk{
+	"test-sdk":   testSdk{s: sdk.Setup{Name: "test-sdk", Revision: sdk.R(1)}, meta: testsdk},
+	"test-sdk-2": testSdk{s: sdk.Setup{Name: "test-sdk-2", Revision: sdk.R(1)}, meta: testsdk2},
+}
+
+func (s *apiSuite) launchWorkshop(c *check.C, name, yaml string, sdks map[string]testSdk) {
 	s.createWFile(c, name, yaml)
 
 	defer s.store.SetActionCallback(storeAction)()
-	defer s.mockDoInstallSdk(c, name, sdks)()
+	defer s.store.SetDownloadCallback(storeDownload)()
+	s.mockSdkVolumes(c, sdks)
 
 	reqbuf := bytes.NewBufferString(fmt.Sprintf(`{"names":["%s"],"action":"launch"}`, name))
 	s.vars = map[string]string{"id": s.project.ProjectId}
@@ -303,8 +311,8 @@ func (s *apiSuite) TestGetWorkshops(c *check.C) {
 	s.d.Overlord().Loop()
 	defer s.d.Overlord().Stop()
 
-	s.launchWorkshop(c, "manysdks", manysdks, testsdks)
-	s.launchWorkshop(c, "basic", basic, map[string]string{})
+	s.launchWorkshop(c, "manysdks", manysdks, apiSuiteSdks)
+	s.launchWorkshop(c, "basic", basic, map[string]testSdk{})
 
 	projectsCmd := apiCmd("/v1/projects/{id}/workshops")
 	s.vars = map[string]string{"id": s.project.ProjectId}
@@ -370,7 +378,7 @@ func (s *apiSuite) TestGetWorkshopInfo(c *check.C) {
 	s.d.Overlord().Loop()
 	defer s.d.Overlord().Stop()
 
-	s.launchWorkshop(c, "manysdks", manysdks, testsdks)
+	s.launchWorkshop(c, "manysdks", manysdks, apiSuiteSdks)
 
 	w, ok := s.b.Workshops[s.project.ProjectId]["manysdks"]
 	c.Assert(ok, check.Equals, true)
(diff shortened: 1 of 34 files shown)


akcano:
@dmitry-lyfar this type of info would probably look nice in a reference-style article about what a workshop is under the hood, which we already have planned for the nearest pulse.

jonathan-conder:
Those LXD ZFS docs don't exist anymore in the latest version, maybe they aren't needed anymore?

internal/daemon/api_workshops_test.go
```diff
@@ -700,56 +722,38 @@ func storeAction(ctx context.Context, actions []sdk.SdkAction) ([]sdk.SdkResult,
...
+	}
+	defer file.Close()
+	if _, err = file.Write([]byte(metayaml)); err != nil {
+		return err
+	}
+
+	hooksdir := filepath.Join(sdkdir, "sdk", "hooks")
+	if err = os.MkdirAll(hooksdir, 0755); err != nil {
+		return err
+	}
```
```suggestion
	return os.MkdirAll(hooksdir, 0755)
```
It wouldn't let me remove the `return nil`

internal/daemon/api_workshops_test.go
```diff
@@ -1923,47 +1943,47 @@ func (s *apiSuite) TestSDKInstallationOrder(c *check.C) {
...
+		{Workshop: "manysdks", Name: sdk.VolumeName(s1.Name, s1.Revision.String())},
+	})
+	s.b.AttachVolumeCalls = s.b.AttachVolumeCalls[:0]
 
 	// Install test-sdk first this time.
 	s.createWFile(c, "manysdks", manysdks)
 
 	requests = []*bytes.Buffer{
-		bytes.NewBufferString(`{"names":["manysdks"],"action":"refresh"}`),
+		bytes.NewBufferString(`{"names":["manysdks"],"action":"launch"}`),
```
jonathan-conder, line 1965: do `launch` and `refresh` share enough code that we don't need to test `refresh` here?

internal/overlord/sdkstate/handlers_test.go
```diff
@@ -193,17 +195,11 @@ func (s *sdkStateSuite) TestDoInstallSdkSuccess(c *check.C) {
...
-	})
 	c.Check(chg.Err(), check.IsNil)
-	c.Check(t1.Status(), check.Equals, state.DoneStatus)
+	c.Check(chg.Status(), check.Equals, state.DoneStatus)
+
+	c.Assert(s.backend.WorkshopVolumeMountPoints, check.HasLen, 1)
+	c.Assert(s.backend.WorkshopVolumeMountPoints["test-2-2"], check.Equals, "/var/lib/workshop/sdk/test-2/2")
 }
 
 func (s *sdkStateSuite) TestDoInstallSdkSuccessWhenLocked(c *check.C) {
```
jonathan-conder:
```suggestion
func (s *sdkStateSuite) TestDoInstallSdkWhenVolumeExists(c *check.C) {
```

internal/workshop/fakebackend/backend.go
```diff
@@ -386,26 +428,51 @@ func (s *FakeWorkshopBackend) DetachVolume(ctx context.Context, wp, name string)
 	}
 	defer wfs.Close()
 
-	afero.Walk(wfs, target, func(path string, info fs.FileInfo, err error) error {
-		s.WorkshopVolumeContents[name][path] = true
-		return nil
-	})
-
 	err = wfs.RemoveAll(target)
```
```suggestion
	err = wfs.Remove(target)
```

internal/workshop/lxd/lxd_backend_project.go
```diff
@@ -19,7 +19,7 @@ import (
 var lxdProjectConfig = map[string]string{
 	"features.images":          "false",
 	"features.profiles":        "true",
-	"features.storage.volumes": "true",
+	"features.storage.volumes": "false",
```
jonathan-conder, line 22: Maybe we should put our volumes in a `workshop` project instead of `default`?

internal/workshop/lxd/lxd_backend_volume.go
```diff
@@ -0,0 +1,228 @@
...
+
+func volumeIndexContent(name string) string {
+	return fmt.Sprintf(volumeIndex, name, storagePool, name)
+}
+
+func lockVolume(volume string) chan struct{} {
+	volumes.Lock()
+	vchan, ok := volumeLocks[volume]
+	if !ok {
+		vchan = make(chan struct{}, 1)
```
jonathan-conder: would `sync.Once` make sense here?

snap/local/hooks/remove
```diff
@@ -30,12 +30,22 @@ for p in $(echo "$projects" | xargs -L1 echo); do
...
-    lxc storage volume delete default "$volume" --project "$p"
+
+done 
+
+# Storage pool volumes are shared among projects and removed last.
+for p in $(echo "$projects" | xargs -L1 echo); do
+  lxc storage volume list --project "$p"
+  # Using awk to filter by type as LXD won't accept the type filter without a
+  # storage name.
+  volumes=$(lxc storage volume list -c pnt -f compact --project "$p" | awk '$3 == "custom" {printf "%s %s ", $1, $2}')
```
jonathan-conder: I think all our volumes are in the `workshop` pool so the previous command should still work (with `workshop` instead of `default`). But most volumes are in the `default` project (only exception I know of is the container rootfs). They seem to be listed with both `--project default` and `--project workshop.<USER>` though.

jonathan-conder (approved):
I really like how this affects the SDK mocks. I have a few questions/suggestions but nothing major.

dmitry-lyfar:
> Those LXD ZFS docs don't exist anymore in the latest version, maybe they aren't needed anymore?

I think the approach is still relevant for loop backs. Unless there is a feature in LXD that does automatic file expansion, so that zfs can then auto-expand into newly allocated blocks (I could not find any confirmation of that). @jonathan-conder

jonathan-conder:
> I think the approach is still relevant for loop backs. Unless there is a feature in LXD that does automatic file expansion, so that zfs can then auto-expand into newly allocated blocks (I could not find any confirmation of that). @jonathan-conder

I managed to track down where it changed: https://github.com/canonical/lxd/pull/11541. Looks like it's still manual but the command is simpler now: https://documentation.ubuntu.com/lxd/en/latest/howto/storage_pools/#resize-a-storage-pool. I guess it should be possible using the API if we want to automate expansion ourselves

internal/daemon/api_workshops_test.go
dmitry-lyfar: Both launch and refresh requests share the SDK installation logic from `installSdks()`. Refresh adds state management to it, but those exist as separate task sets.

internal/workshop/lxd/lxd_backend_volume.go
dmitry-lyfar: Do you mean instead of checking for existence? Not sure I get the idea, could you elaborate?

That actually made me think that unlockVolume() should panic if the volume wasn't locked (e.g the channel did not exist for the given volume).

internal/workshop/lxd/lxd_backend_project.go
dmitry-lyfar: They are in the `workshop` project. This flag means that if you create a volume in a project (say, workshop.user), it still will be visible to those users / projects that have access to the storage pool (`workshop`).

jonathan-conder: In my branch (which I think was up to date this morning, maybe not now though), they are in the `workshop` storage pool but not the `workshop` project:
```console
$ lxc storage volume list --all-projects workshop type=custom
+---------+--------+------------------------+-------------+--------------+---------+
| PROJECT |  TYPE  |          NAME          | DESCRIPTION | CONTENT-TYPE | USED BY |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | dev-1540ebfb-cache-apt |             | filesystem   | 1       |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | go-158                 |             | filesystem   | 0       |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | go-440                 |             | filesystem   | 0       |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | go-45                  |             | filesystem   | 0       |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | test-sdk-basic-35      |             | filesystem   | 1       |
+---------+--------+------------------------+-------------+--------------+---------+
| default | custom | ws-9ba9473c-cache-apt  |             | filesystem   | 1       |
+---------+--------+------------------------+-------------+--------------+---------+
```

internal/workshop/lxd/lxd_backend_volume.go
jonathan-conder: I haven't fully thought it through, I just think if you can use a `sync.Once` instead of a channel it would be more readable. I'm not sure what code exactly would go in the `sync.Once` callback though.

jonathan-conder: this part would be something like
```go
if !ok {
    vchan = &sync.Once{}
    volumeLocks[volume] = vchan
}
```
and `ImportVolume` could call `vchan.Do(callback)` instead of the `select`.

The callback would do the actual import, remove the `Once` from `volumeLocks`, and set a flag (local variable in `ImportVolume`). If the flag is set, `ImportVolume` can return right away. Otherwise it can probably just return `ErrVolumeAlreadyExists`. Seems a bit unorthodox though, maybe the current way is better.

internal/workshop/lxd/lxd_backend_project.go
dmitry-lyfar: As per the conversation: custom volumes will be created in a default project for the legacy reasons. These will still be visible in other workshop projects. The default project can no longer be deleted from LXD (it exists before `lxd init` was called).

The snap remove hook, therefore, should remove all volumes from the workshop storage pool.

internal/workshop/lxd/lxd_backend_volume.go
dmitry-lyfar: That's an interesting idea. Returning `ErrVolumeAlreadyExists` from ImportVolume should happen when the volume actually exists. If the local flag is not set after the `Once` has completed, the `ImportVolume` would have to check if the volume exists and return something like `ErrVolumeNotFound` which, for the `ImportVolume`, does not seem to make sense given its goal is to create the volume.

In the current implementation, if a go routine waits for the other one to finish import and continues once that other one was finished, it can detect whether a volume was created or not to decide what it wants to do (e.g. the first go routine failed to import the volume because of a transient or a permanent error, so the volume does not actually exist; hence, the go routine would attempt to import itself and would either resolve the issue or fail with the same permanent error).

I would probably stick to the channel's based sync here as it allows the client to wait for the ongoing import to finish or cancel waiting if the launch or refresh was aborted.

Also, it allows other methods to use a more generic lock/unlock logic if required (e.g. syncing between two task sets or changes that work with the same volume). We don't have a scenario where Import and Create would concurrently try to operate a volume with the same name yet. However, I think DeleteVolume may conflict with ImportVolume at some point, when the SDK management will be extended to smth. like `sdk delete <name>`.

jonathan-conder: Yeah I think being able to cancel waiting is a good reason to prefer channels. I still think it's a bit confusing though. Maybe we could rename `volumeLocks` to something else, e.g. `onceChannels` or something

snap/local/hooks/remove
```diff
@@ -29,13 +29,16 @@ for p in $(echo "$projects" | xargs -L1 echo); do
...
 
-  lxc storage volume list default --project "$p"
-  volumes=$(lxc storage volume list default type=custom -c n -f compact --project "$p" | awk 'NR>1 {printf "%s ", $1}')
-  echo "Removing storage volumes: $volumes"
-  for volume in $(echo "$volumes" | xargs -L1 echo); do
-    lxc storage volume delete default "$volume" --project "$p"
-  done
+# Storage pool volumes are shared among projects and removed last.
+storage_pool="workshop"
+volumes=$(lxc storage volume list "$storage_pool" -c pnt type=custom -f compact | awk 'NR>1 {printf "%s ", $1}')
```
jonathan-conder: We might want `--project default` or `--all-projects` here, in case the user previously ran `[sudo] lxc project switch other-project` (I'm guessing `sudo` is needed but not sure which user runs the snap hook).

dmitry-lyfar: True. I assumed that storage volume list ignores projects for those volumes that are not per-project given that custom volumes creation ignore the project setting.

internal/workshop/lxd/lxd_backend_volume.go
dmitry-lyfar: I reworked the locker interface so it doesn't expose the channel. It's a common pattern to use channels for synchronization. Our case is a bit more specific given we want it to be cancelable.

PR #269, opened 2025-01-20
Add command to list workshop scripts

jonathan-conder:
# Description

For example:
```console
$ cat workshop.yaml
name: ws
base: ubuntu@24.04
scripts:
  foo: echo foo
  bar: |
    echo bar
$ workshop scripts ws
bar:
  script: |
    echo bar
foo:
  script: echo foo
```
The response format includes `script:` so it can be extended without breaking anything. In future we will probably support both formats in workshop definitions, e.g.:
```yaml
...
scripts:
  foo: echo foo
  bar:
    script: echo bar
    uid: 0
    gid: 0
```

Like `workshop run`, the docs are a bit lacking, but it would make sense to include scripts in a future how-to guide for CI.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -26,6 +26,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"golang.org/x/sys/unix"
+	"gopkg.in/yaml.v3"
 
 	"github.com/canonical/workshop/client"
 	"github.com/canonical/workshop/internal/logger"
@@ -151,6 +152,11 @@ Notes:
 
 - You can set the working directory, environment variables, user and group ID
   for running the script in the workshop; reasonable defaults are provided.
+`
+
+var shortScriptsHelp = "List workshop scripts"
+var longScriptsHelp = `
+This command enumerates all scripts in the workshop, printing a YAML map.
 `
 
 func (c *CmdExec) Command() *cobra.Command {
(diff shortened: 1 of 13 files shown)


dmitry-lyfar (approved):

internal/daemon/api_workshops.go
```diff
@@ -456,3 +460,34 @@ func v1GetProjectWorkshop(c *Command, r *http.Request, _ *userState) Response {
 
 	return SyncResponse(rsp, http.StatusOK)
 }
+
+func v1GetProjectWorkshopScripts(c *Command, r *http.Request, _ *userState) Response {
+	projectId := muxVars(r)["id"]
+	name := muxVars(r)["name"]
+
+	if projectId == "" {
+		return statusBadRequest("project-id must be provided")
```
akcano:
```suggestion
		return statusBadRequest("project-id required")
```

internal/daemon/api_workshops.go
```diff
@@ -456,3 +460,34 @@ func v1GetProjectWorkshop(c *Command, r *http.Request, _ *userState) Response {
...
+	}
+
+	if name == "" {
+		return statusBadRequest("workshop name must be provided")
```
akcano:
```suggestion
		return statusBadRequest("workshop name required")
```

cmd/workshop/exec.go
```diff
@@ -153,6 +154,11 @@ Notes:
   for running the script in the workshop; reasonable defaults are provided.
 `
 
+var shortScriptsHelp = "List workshop scripts"
+var longScriptsHelp = `
+This command enumerates all scripts in the workshop, printing a YAML map.
```
akcano, line 159: This prompts me to add workshop scripts to the coverage map, otherwise we're missing some context. The description is fine, though.

akcano (changes requested):

dmitry-lyfar (approved):

PR #274, opened 2025-01-24
Cleanup StartWorkshop progress if failed

dmitry-lyfar:
# Description

The StartWorkshop backend method does multiple things. If, for example, waiting for the systemd to be ready fails, the
method must clean up its progress made so far. In most cases it means that the instance must be set back to Stopped and the auto-start option reverted.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/workshop/lxd/tests/integration/workshop_test.go
+++ b/internal/workshop/lxd/tests/integration/workshop_test.go
@@ -265,3 +265,23 @@ func (f *wsOps) TestLxdBackendDownloadMultipleBasesConcurrently(c *check.C) {
 		c.Assert(f.deleteimage(c, fp), check.IsNil)
 	}
 }
+
+func (f *wsOps) TestLxdBackendWorkshopStartFailed(c *check.C) {
+	helper.LaunchTestWorkshop(c, f.ctx, f.bd, f.project.Path)
+	defer helper.RemoveTestWorkshop(c, f.ctx, f.bd)
+
+	err := f.bd.StopWorkshop(f.ctx, "test", true)
+	c.Check(err, check.IsNil)
+
+	// Leaves the workshop instance in a started state with a failed start
+	// command. The StartWorkshop API must clean up its previous progress, i.e.
+	// set the workshop to the Stopped state.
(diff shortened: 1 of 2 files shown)


internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -265,3 +265,23 @@ func (f *wsOps) TestLxdBackendDownloadMultipleBasesConcurrently(c *check.C) {
...
+	helper.LaunchTestWorkshop(c, f.ctx, f.bd, f.project.Path)
+	defer helper.RemoveTestWorkshop(c, f.ctx, f.bd)
+
+	err := f.bd.StopWorkshop(f.ctx, "test", true)
+	c.Check(err, check.IsNil)
+
+	// Leaves the workshop instance in a started state with a failed start
+	// command. The StartWorkshop API must clean up its previous progress, i.e.
+	// set the workshop to the Stopped state.
+	lxdbackend.FakeStartCommand("exit 1")
```
jonathan-conder: this should be reverted at the end of the test

lachypjones:
This doesn't appear to fix the issue for launch:
```
❯ workshop launch blank
Create new "blank" workshop                                                                                                                       \
- Create new "blank" workshop (context canceled)
"blank" launch aborted
❯ workshop launch blank
error: cannot launch "blank": workshop exists
```

Then:
```
❯ workshop start blank
- Start "blank" workshop (Error occurred when starting proxy device: Error: Failed to listen on /var/lib/workshop/run/workshop.socket.untrusted: listen unix /var/lib/workshop/run/workshop.socket.untrusted: bind: no such file or directory)
```

jonathan-conder (approved):
I think this is good on its own, but I'm pretty sure I'm going to forget to call `context.WithoutCancel` when I write something like this myself. Maybe we should abstract it at some point, similar to `AddCleanup`. But I'll need to think about it more.

lachypjones (approved):

PR #275, opened 2025-01-28
Add code completion for command arguments

lachypjones:
# Description
Adds dynamic shell completion to all commands that would benefit from it. 
All completions are checked for validity. That is, where a completion is provided, the user should expect to be able to hit `enter` and have the command succeed. 

Highlights include `connect` and `remount`.

Commands will now not provide file/folder completions unless relevant to the command in question (ie. remount). 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/changes.go
+++ b/cmd/workshop/changes.go
@@ -57,24 +57,10 @@ func (c *CmdChanges) Run(cmd *cobra.Command, av []string) error {
 		return err
 	}
 
-	clientOpts := client.ChangesOptions{
-		ProjectPath: c.root.project,
-		Selector:    client.ChangesAll,
-	}
-
-	chngs, err := cli.Changes(&clientOpts)
+	chngs, err := c.changes(cli)
 	if err != nil {
 		return err
 	}
-
-	slices.SortFunc(chngs, func(a, b *client.Change) int {
-		if a.SpawnTime.Before(b.SpawnTime) {
-			return -1
-		} else if a.SpawnTime.After(b.SpawnTime) {
-			return 1
-		}
-		return 0
-	})
 
 	if len(chngs) > 0 {
 		w := tabWriter()
@@ -98,4 +84,26 @@ func (c *CmdChanges) Run(cmd *cobra.Command, av []string) error {
 	}
 
 	return nil
+}
+
+func (c *CmdChanges) changes(cli *client.Client) ([]*client.Change, error) {
+	clientOpts := client.ChangesOptions{
+		ProjectPath: c.root.project,
+		Selector:    client.ChangesAll,
+	}
+
+	chngs, err := cli.Changes(&clientOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	slices.SortFunc(chngs, func(a, b *client.Change) int {
+		if a.SpawnTime.Before(b.SpawnTime) {
+			return -1
+		} else if a.SpawnTime.After(b.SpawnTime) {
+			return 1
+		}
+		return 0
+	})
+	return chngs, nil
 }
--- a/cmd/workshop/connect.go
+++ b/cmd/workshop/connect.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"fmt"
+	"maps"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -54,7 +56,8 @@ $ workshop connect nimble/go:mod-cache :mount
 
 A full version of the command that also lists the target SDK ('system'):
 $ workshop connect nimble/go:mod-cache nimble/system:mount`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.complete(),
 	}
 
 	cmd.PersistentFlags().BoolVar(&c.NoWait, "no-wait",
@@ -131,4 +134,50 @@ func (c *CmdConnect) Run(cmd *cobra.Command, av []string) error {
 	}
 
 	return nil
+}
+
+func (c *CmdConnect) complete() ValidArgsFunction {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		cli, err := c.root.client()
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		project, err := cli.Project(c.root.project)
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		connections, err := cli.Connections(&client.ConnectionOptions{ProjectId: project.Id, All: true})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		// Create map of endpoint string and associated plugs
+		disconnectedPlugs := make(map[string]client.Plug)
+		for _, plug := range connections.Plugs {
+			if len(plug.Connections) == 0 {
+				plugString := endpoint(plug.Workshop, plug.Sdk, plug.Name)
+				disconnectedPlugs[plugString] = plug
+			}
+		}
+
+		var completions []string
+		switch len(args) {
+		case 0:
+			// No arguments, show list of plugs
+			completions = slices.Collect(maps.Keys(disconnectedPlugs))
+		case 1:
+			plug, ok := disconnectedPlugs[args[0]]
+			if !ok {
+				break
+			}
+			for _, slot := range connections.Slots {
+				if plug.Interface == slot.Interface && plug.Workshop == slot.Workshop {
+					completions = append(completions, endpoint(slot.Workshop, slot.Sdk, slot.Name))
+				}
+			}
+		}
+		return completions, cobra.ShellCompDirectiveNoFileComp
+	}
 }
--- a/cmd/workshop/connections.go
+++ b/cmd/workshop/connections.go
@@ -42,7 +42,8 @@ $ workshop connections nimble
 
 List connections for all workshops in the current project directory:
 $ workshop connections`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Pending", "Stopped"}),
 	}
 
 	cmd.Flags().BoolVar(&c.all, "all", false, "Include disconnected plugs in the output.")
--- a/cmd/workshop/disconnect.go
+++ b/cmd/workshop/disconnect.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"maps"
+	"slices"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -51,7 +53,8 @@ $ workshop disconnect nimble/go:mod-cache nimble/system:mount
 Disconnect all plugs connected to the 'mount' slot of the 'system' SDK
 under the 'nimble' workshop in the current project directory:
 $ workshop disconnect nimble/system:mount`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.complete(),
 	}
 
 	cmd.PersistentFlags().BoolVar(&c.forget, "forget",
@@ -113,4 +116,46 @@ func (c *CmdDisconnect) Run(cmd *cobra.Command, av []string) error {
 	}
 
 	return nil
+}
+
+func (c *CmdDisconnect) complete() ValidArgsFunction {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		cli, err := c.root.client()
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		project, err := cli.Project(c.root.project)
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		connections, err := cli.Connections(&client.ConnectionOptions{ProjectId: project.Id, All: true})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		// Create map of endpoint strings and associated plugs
+		connectedPlugs := make(map[string]client.Plug)
+		for _, plug := range connections.Plugs {
+			if len(plug.Connections) > 0 {
+				ref := endpoint(plug.Workshop, plug.Sdk, plug.Name)
+				connectedPlugs[ref] = plug
+			}
+		}
+
+		var completion []string
+		switch len(args) {
+		case 0:
+			// No arguments, show list of plugs
+			completion = slices.Collect(maps.Keys(connectedPlugs))
+		case 1:
+			if plug, ok := connectedPlugs[args[0]]; ok {
+				for _, slot := range plug.Connections {
+					completion = append(completion, endpoint(slot.Workshop, slot.Sdk, slot.Name))
+				}
+			}
+		}
+		return completion, cobra.ShellCompDirectiveNoFileComp
+	}
 }
--- a/cmd/workshop/refresh.go
+++ b/cmd/workshop/refresh.go
@@ -81,7 +81,8 @@ $ workshop refresh --continue
 
 Refresh the sketch SDK in the 'nimble' workshop:
 $ workshop refresh nimble/sketch`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Pending"}),
 	}
 
 	cmd.PersistentFlags().BoolVar(&c.WaitOnError, "wait-on-error",
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -42,7 +42,8 @@ $ workshop remove nimble jazzy
 
 The name is optional if the project has only one workshop:
 $ workshop remove`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Stopped"}),
 	}
 
 	return cmd
--- a/cmd/workshop/root.go
+++ b/cmd/workshop/root.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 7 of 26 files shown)


lachypjones:
The runner seems to be having issues, works locally:
```
- Create new "golang" workshop (Failed creating instance from image: Unable to unpack image, run out of disk space)
"golang" launch aborted
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
 act as the Go modules cache for the workshop.
 
 
+Install shell completions
+-------------------------
+
+|ws_markup| features shell completion for popular shells such as `bash`, `zsh`
+and `fish`. Bash completion is installed automatically,
```
akcano:
```suggestion
|ws_markup| features shell completion for popular shells
such as :program:`bash`, :program:`zsh` and `fish`.
Bash completion is configured automatically;
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+to install completion for other shells,
+use :ref:`completion`<ref_workshop_cli_completion>`
```
akcano:
```suggestion
use :ref:`completion`<ref_workshop_cli_completion>`.
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+
+To see how this can be useful, let's look at the `connect` command above.
```
akcano:
```suggestion
To see how this can be useful, let's look at the :command:`connect` command above.
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+This command contains up to six sections that must be remembered:
+
+* The respective plug and slot\ :sup:`*` workshop.
+* The respective plug and slot\ :sup:`*` sdk.
+* The respective plug and slot name.
```
akcano:
```suggestion
- The names of the plug and the slot themselves.
- Their respective workshop\ :sup:`*` names.
- Their respective SDK\ :sup:`*` names.
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+
+:sup:`*` *the workshop and sdk may be omitted for system slots*
```
akcano:
```suggestion
:sup:`*` *Workshop and SDK names may be omitted for system SDK slots*
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+This would look something like:
+
+.. code-block:: console
+
+   $ workshop connect <workshop>/<sdk>:<plug> <workshop>/<sdk>:<slot>
+
+
+Shell completion can simplify this by automatically finding appropriate
+candidate plugs and slots for you, eliminating the need to run
+`workshop connections <workshop>` to determine what is available.
```
akcano:
```suggestion
:command:`workshop connections` to see what's available.
```

docs/tutorial/index.rst
```diff
@@ -590,8 +590,41 @@ This makes :file:`/home/user/mod/` on the host
...
+
+After installation, completion can be accessed by pressing the `tab` key during
+or after entering a command.
```
akcano:
```suggestion
After installation, you can access command completion by pressing the :kbd:`Tab` key
while typing or after entering a command.
```

akcano (changes requested):
great job! some minor changes

cmd/workshop/changes.go
```diff
@@ -99,3 +85,25 @@ func (c *CmdChanges) Run(cmd *cobra.Command, av []string) error {
...
+		ProjectPath: c.root.project,
+		Selector:    client.ChangesAll,
+	}
+
+	chngs, err := cli.Changes(&clientOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	slices.SortFunc(chngs, func(a, b *client.Change) int {
```
dmitry-lyfar, line 100: why sorting as part of the client library? Sorting is a representation logic.

lachypjones: I wanted the completion ordering to be sorted identically to the 'changes' output. 
For instance:
```
workshop tasks 7 *tab*
771 772 773 774 
```

cmd/workshop/completions.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	workshopInfo, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
+	if err != nil {
+		return nil, err
+	}
+
+	var workshops []string
+	for _, workshop := range workshopInfo {
+		if allowedWorkshopStatus == nil || slices.Contains(allowedWorkshopStatus, workshop.Status) {
```
dmitry-lyfar: No need for nil check.

cmd/workshop/completions.go
```diff
@@ -0,0 +1,48 @@
...
+	Project(path string) (*client.Project, error)
+	List(opts *client.ListOptions) ([]*client.WorkshopInfo, []*client.WorkshopFile, error)
+	Connections(*client.ConnectionOptions) (client.Connections, error)
+	Changes(*client.ChangesOptions) ([]*client.Change, error)
+}
+
+// Returns a list of workshops in the current project that have a status
+// permitted in allowedWorkshopStatus. If allowedWorkshopStatus is nil,
+// returns all workshops
+func workshopInfoList(allowedWorkshopStatus []string, cli Client, projectId string) ([]string, error) {
```
dmitry-lyfar: allowed is, perhaps, not the best choice here. It is a filter by the status property. They are all allowed. `statusFilter`?

cmd/workshop/completions.go
```diff
@@ -0,0 +1,48 @@
...
+		if allowedWorkshopStatus == nil || slices.Contains(allowedWorkshopStatus, workshop.Status) {
+			workshops = append(workshops, workshop.Name)
+		}
+	}
+
+	return workshops, nil
+}
+
+// Returns the elements in a that are not present in b
+func subtractSlice[T comparable](a, b []T) (result []T) {
```
dmitry-lyfar: Any reason to use generics? This is string in all cases.

cmd/workshop/completions.go
```diff
@@ -0,0 +1,48 @@
+package main
+
+import (
+	"slices"
+
+	"github.com/canonical/workshop/client"
+)
+
+type Client interface {
```
dmitry-lyfar: This Client interface interferes with the implementation and it's an adhoc solution that is retrofitted for tests. Even if we decided to introduce an interface for the client library, it should live in the client package and not in the unit which is responsible for the completions.

But, firstly, what is the reason not to test completions in the same way as for the other commands where you can generate a bunch of testing json responses and write some wrappers to minimise the tests themselves?

cmd/workshop/refresh.go
```diff
@@ -82,6 +82,19 @@ $ workshop refresh --continue
 Refresh the sketch SDK in the 'nimble' workshop:
 $ workshop refresh nimble/sketch`,
 		RunE: c.Run,
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
dmitry-lyfar: I checked the other commands briefly. My initial observations:

1. `remove`, `refresh`, `start`, `stop` share the logic.
2. Commands like `launch` implement something called `generationCompletions`, but do the same thing.

So:

1. It is possible to do something like this:
```
// refresh.go
// ...
		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Pending"}),
	}

// root.go
type ValidArgsFunction func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)

func (c *CmdRoot) completeWorkshopName(statuses []string) ValidArgsFunction {
	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		cli, err := c.client()
		if err != nil {
			return nil, cobra.ShellCompDirectiveError
		}

		workshops, err := nameCompletion(statuses, cli, c.project)
		if err != nil {
			return nil, cobra.ShellCompDirectiveError
		}

		return subtractSlice(workshops, args), cobra.ShellCompDirectiveNoFileComp
	}
}


```

This avoids duplication, does not inline function into the length cobra.Command structure initialisation (see how Run function is not inline there, the rest should be the same way).

I did not go any further, but I suppose it is possible to deduplicate it more.

dmitry-lyfar (changes requested):

cmd/workshop/changes.go
dmitry-lyfar: Ah, yes, I thought you moved it under the client for some reason, my mistake. But the changes command doesn't seem to have ValidArgsFunc?

dmitry-lyfar: I mean, why don't the CmdTasks `generateCompletions` take care of sorting changes?

lachypjones: It could, but then if we change sorting it will need to be changed in both places (something that could be easily missed, and I don't think sorting numbers lexicographically makes the most sense so it is possible it may change). Considering I would be using exactly the same logic I thought it would be better to just break out the core part (retrieval + sorting) of the changes cmd. Happy to change if you think otherwise :)

cmd/workshop/connect.go
```diff
@@ -132,3 +135,54 @@ func (c *CmdConnect) Run(cmd *cobra.Command, av []string) error {
...
+			for _, conn := range connections.Established {
+				if endpoint(plug.Workshop, plug.Sdk, plug.Name) == endpoint(conn.Plug.Workshop, conn.Plug.Sdk, conn.Plug.Name) {
+					// We ignore plugs that are currently connected
+					connected = true
+					break
+				}
+			}
+			if !connected {
+				plugString := endpoint(plug.Workshop, plug.Sdk, plug.Name)
+				unboundPlugs[plugString] = plug
```
dmitry-lyfar: why is it an unbound plug? Bound means that the plug has .Bind != nil. Also, why filter bound plugs? They can still be connected and disconnected.

cmd/workshop/connect.go
```diff
@@ -132,3 +135,54 @@ func (c *CmdConnect) Run(cmd *cobra.Command, av []string) error {
...
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		// Create list of unbound plugs and their endpoint strings
+		unboundPlugs := make(map[string]client.Plug)
+		for _, plug := range connections.Plugs {
+			connected := false
+			for _, conn := range connections.Established {
+				if endpoint(plug.Workshop, plug.Sdk, plug.Name) == endpoint(conn.Plug.Workshop, conn.Plug.Sdk, conn.Plug.Name) {
```
dmitry-lyfar: It's better to compare references than endpoints because the latter does not include project id:

```
if plug.Ref() == conn.Plug {

```

cmd/workshop/connect.go
```diff
@@ -132,3 +135,54 @@ func (c *CmdConnect) Run(cmd *cobra.Command, av []string) error {
...
+		connections, err := cli.Connections(&client.ConnectionOptions{ProjectId: project.Id, All: true})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		// Create list of unbound plugs and their endpoint strings
+		unboundPlugs := make(map[string]client.Plug)
+		for _, plug := range connections.Plugs {
+			connected := false
+			for _, conn := range connections.Established {
```
dmitry-lyfar: does it mane sense to iterate here when every plug and slot from the client library has an attribute called `Connections`. So you can check if the Connections is not empty.

cmd/workshop/connect.go
dmitry-lyfar: (it can use endpoint as connections are requested per project, just so you know if the list contains conns from other projects this logic may not work)

cmd/workshop/disconnect.go
```diff
@@ -114,3 +117,43 @@ func (c *CmdDisconnect) Run(cmd *cobra.Command, av []string) error {
...
+		}
+
+		connections, err := cli.Connections(&client.ConnectionOptions{ProjectId: project.Id, All: true})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		// Create list of connected plugs and their endpoint strings
+		establishedConns := make(map[string]client.Connection)
+		for _, conn := range connections.Established {
```
dmitry-lyfar: Can you just find a plug by its reference from the `connections.Plugs` and then get the list of the slots it is connected to? Similar to the logic for `connect` above, but in a reverse way.

cmd/workshop/root.go
```diff
@@ -90,6 +91,37 @@ func (c *CmdRoot) postRun(cmd *cobra.Command, args []string) {
...
+		}
+
+		workshopInfo, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		var workshops []string
+		for _, workshop := range workshopInfo {
+			if (status == nil ||
```
dmitry-lyfar: No need to compare it against `nil`, slices.Contains handles this.

dmitry-lyfar (changes requested):

cmd/workshop/root.go
lachypjones: The nil check isn't a safety thing, it's because passing `nil` as a filter should return all workshops in the project (ie. `workshop info`)

cmd/workshop/connect.go
lachypjones: Bad wording on my part. I used `bound` to refer to a plug with a bound slot (i.e. connection). I'll reword some comments and vars.

cmd/workshop/disconnect.go
lachypjones: The logic for connect (including your suggestions) still requires an iteration step to build a list of endpoint strings (both to return, and to compare with the current argument). The only improvement I see here would be to compare with arg0 as I go, however when testing (with 20 workshops) it didn't make a meaningful difference to the completion speed, so I elected not to)

connections.Established will always be <= to that of connections.Plugs, so it made more sense to iterate through this. When there are no args, I need all the connections, when there are args, the only parameter I have to work with is an endpoint string which I then need something to compare with.

If i'm missing something apologies.

cmd/workshop/root.go
dmitry-lyfar: OK, that makes sense. This is a suggestion to make it more readable:

```
		matchesStatus := func(s string) bool {
			if status == nil {
				return true
			}
			return slices.Contains(status, s)
		}
                // ...
		for _, workshop := range workshopInfo {
			if matchesStatus(workshop.Status) && !slices.Contains(args, workshop.Name) {
				workshops = append(workshops, workshop.Name)
			}
		}
```

cmd/workshop/connect.go
```diff
@@ -132,3 +135,46 @@ func (c *CmdConnect) Run(cmd *cobra.Command, av []string) error {
...
+			}
+		}
+
+		var completion []string
+		switch len(args) {
+		case 0:
+			// No arguments, show list of plugs
+			completion = slices.Collect(maps.Keys(disconnectedPlugs))
+		case 1:
+			plug := disconnectedPlugs[args[0]]
```
dmitry-lyfar: disconnect checks if the element is present, connect does not.

lachypjones: It doesn't matter as it won't find a candidate slot. But yes, It's better with an explicit check.

cmd/workshop/disconnect.go
dmitry-lyfar: Neither of the options make any noticeable difference from the performance perspective as both are O(N) with N being, mostly, tiny. All the significant time spent in this operation is getting a bunch of json over the socket and parsing it, i.e. I/O.

It's about readability. There is a list of plugs. Filter them to get a list of connected ones. Then make suggestions: all connected plugs OR a list of slots that the arg[0] plug connected to. So, the code should be approximately this plain for such a simple piece of logic.

It works with `connections` too, but the intention is less clear. You comment says `// Create list of connected plugs and their endpoint strings` but the variable name is `establishedConns`. Is it a list of connected plugs or established connections? Also, a plug can be connected to multiple slots (not realistic right now, but possible).

I gave it a go:

```
		connectedPlugs := make(map[string]client.Plug)
		for _, plug := range connections.Plugs {
			if len(plug.Connections) > 0 {
				ref := endpoint(plug.Workshop, plug.Sdk, plug.Name)
				connectedPlugs[ref] = plug
			}
		}

		var completion []string
		switch len(args) {
		case 0:
			// No arguments, show list of plugs
			completion = slices.Collect(maps.Keys(connectedPlugs))
		case 1:
			if plug, ok := connectedPlugs[args[0]]; ok {
				for _, slot := range plug.Connections {
					completion = append(completion, endpoint(slot.Workshop, slot.Sdk, slot.Name))
				}
			}
		}
```

FYI, disconnect works with a slot as a first argument. I don't think we should support this in the completion when no arguments are provided, though as it would be a confusing list for the first-time user.

cmd/workshop/connections.go
```diff
@@ -42,7 +42,8 @@ $ workshop connections nimble
 
 List connections for all workshops in the current project directory:
 $ workshop connections`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Pending"}),
```
dmitry-lyfar: `connections` can work for the Stopped workshops too.

cmd/workshop/remove.go
```diff
@@ -42,7 +42,8 @@ $ workshop remove nimble jazzy
 
 The name is optional if the project has only one workshop:
 $ workshop remove`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready"}),
```
dmitry-lyfar: `remove` can remove Stopped workshops too.

cmd/workshop/tasks.go
```diff
@@ -112,3 +113,29 @@ func (c *CmdTasks) Run(cmd *cobra.Command, av []string) error {
...
+		}
+
+		changes, err := changesCmd.changes(cli)
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveError
+		}
+
+		chgs := make([]string, len(changes))
+		for _, chg := range changes {
+			chgs = append(chgs, chg.ID)
```
dmitry-lyfar: I think tasks completion should be more informative than ids:

```
func (c *CmdTasks) complete() ValidArgsFunction {
	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		cli, err := c.root.client()

		if err != nil {
			return nil, cobra.ShellCompDirectiveError
		}

		changesCmd := CmdChanges{
			root: c.root,
		}

		changes, err := changesCmd.changes(cli)
		if err != nil {
			return nil, cobra.ShellCompDirectiveError
		}

		l := len(changes)
		num := min(l, 10)
		completions := make([]string, l)

		for _, chg := range changes[l-num : l] {
			completions = append(completions, fmt.Sprintf("%s\t%-5s %s\n", chg.ID, chg.Status, chg.Summary))
		}

		return completions, cobra.ShellCompDirectiveNoFileComp
	}
}
```

Also, note, it should stop suggesting arguments if len(args) == 1 as the command only accepts one positional argument.

lachypjones: This is great!

dmitry-lyfar (approved):

docs/tutorial/index.rst
dmitry-lyfar: :program:`fish`

docs/tutorial/index.rst
dmitry-lyfar: this renders with an extra ` after the word "completion", see https://canonical-workshop--275.com.readthedocs.build/en/275/tutorial/#install-shell-completions

PR #277, opened 2025-01-30
Use separate LXD bridge for workshops

jonathan-conder:
# Description

At the moment this just provides another level of separation between Workshop containers and other LXD containers users might have. In future, we may want to lock down this interface (to prevent issues like https://github.com/moby/moby/issues/14041) or apply a workaround for the [UFW issue](https://documentation.ubuntu.com/lxd/en/latest/howto/network_bridge_firewalld/#ufw-add-rules-for-the-bridge). We could also make workshops addressable as `<NAME>-<ID>.workshop` via `systemd-resolved`.

The network is named `workshopbr0`, similar to `lxdbr0`. I think the maximum length for an interface name is 15 bytes, so we should be able to go all the way to `workshopbr99999` in the unlikely event we need more than one network in future.

I don't think we need to document this, but there are a few new error messages. One of them is based on an existing message, and I hope `LXD not initialized` will be enough to direct users to https://canonical-workshop.readthedocs-hosted.com/en/latest/how-to/use-workshops/troubleshoot/#install-and-start-lxd.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -29,6 +29,8 @@ const (
 	LxdSock           = "/var/snap/lxd/common/lxd/unix.socket"
 	storagePool       = "workshop"
 	storagePoolDriver = "zfs"
+	networkName       = "workshopbr0"
+	networkType       = "bridge"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


dmitry-lyfar (approved):

internal/workshop/lxd/lxd_backend.go
```diff
@@ -119,6 +142,35 @@ func New() (*Backend, error) {
 		}
 	}
 
+	networkExists := false
+	for _, network := range networks {
+		if network.Name == networkName {
+			if network.Type != networkType {
+				return nil, fmt.Errorf("network %q already exists with different type %q", networkName, network.Type)
```
akcano:
```suggestion
				return nil, fmt.Errorf("network %q already exists with a different type: %q", networkName, network.Type)
```

akcano (changes requested):

PR #279, opened 2025-01-30
Doc: Auto-generate coverage map with a GitHub workflow or manually

akcano:


dmitry-lyfar (approved):
As discussed, it looks pretty good. Some fixup commits need auto squashing.

Also, not sure if it is meant to be viewable from the GitHub renderer, it appears that the header is a bit broken here: https://github.com/akcano/workshop/blob/main/docs/coverage.md

akcano:
@dmitry-lyfar I dropped the header with the latest fix; also, added line numbers for GitHub.

PR #283, opened 2025-02-02
Ensure minimal workable zfs pool size

dmitry-lyfar:
# Description

Ensure the storage pool is no less than 5GiB, otherwise Workshop runs out of space in tests and environments with less than ~14GiB of available space. LXD defaults to 20% in those cases which results in a ~2GiB pool size for Workshop.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/snap/local/hooks/remove
+++ b/snap/local/hooks/remove
@@ -35,7 +35,7 @@ lxc network delete workshopbr0
 
 # Storage pool volumes are shared among projects and removed last.
 storage_pool="workshop"
-volumes=$(lxc storage volume list "$storage_pool" -c pnt type=custom -f compact --all-projects | awk 'NR>1 {printf "%s ", $1}')
+volumes=$(lxc storage volume list "$storage_pool" -c nt type=custom -f compact --all-projects | awk 'NR>1 {printf "%s ", $1}')
 echo "$volumes" | xargs -n1 -r | while read -r volume; do
   lxc storage volume delete "$storage_pool" "$volume"
 done
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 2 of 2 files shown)


.github/workflows/spread.yaml
```diff
@@ -56,6 +56,9 @@ jobs:
         run: |
           mkdir ${{ github.workspace }}.cover
           spread -artifacts=${{ github.workspace }}.cover tests/${{matrix.suite}}/
+      - name: Setup tmate session
+        if: ${{ failure() }}
+        uses: canonical/action-tmate@main
```
jonathan-conder:
```suggestion
```

jonathan-conder (changes requested):
Look like this fixes the spread issue? I'm happy with it, but assume you don't want to keep the `tmate` step?

.github/workflows/spread.yaml
dmitry-lyfar: I was considering leaving it as is for debugging. However, I think that due to the transient issues we have sometimes, it's not very practical for regular runs.

jonathan-conder (approved):

snap/local/hooks/remove
```diff
@@ -35,7 +35,7 @@ lxc network delete workshopbr0
 
 # Storage pool volumes are shared among projects and removed last.
 storage_pool="workshop"
-volumes=$(lxc storage volume list "$storage_pool" -c pnt type=custom -f compact --all-projects | awk 'NR>1 {printf "%s ", $1}')
+volumes=$(lxc storage volume list "$storage_pool" -c nt type=custom -f compact --all-projects | awk 'NR>1 {printf "%s ", $1}')
 echo "$volumes" | xargs -n1 -r | while read -r volume; do
   lxc storage volume delete "$storage_pool" "$volume"
```
jonathan-conder, line 40: not related to these changes, but I just noticed this line needs a `--project default`

dmitry-lyfar: Will address in a separate PR.

PR #285, opened 2025-02-03
Align shell/exec/run sessions and environment

lachypjones:
# Description
Ensures that workshop `shell`, `exec` and `run` sessions contain user and group information as configured in the instance, and an environment derived from a login shell.

For example:

- Commands that rely on _instance provided_ information for user information (groups, home, etc.) will now work as expected. 
- Environment variables set by interfaces will now be present in all cases

Commands will now at a minimum contain the following:
```
 sudo -u <uid> -g <gid> --preserve-env -- bash -l
```

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/signal"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
@@ -251,7 +252,7 @@ $ workshop shell`,
 }
 
 func (c *CmdShell) Run(cmd *cobra.Command, av []string) error {
-	args := &ExecArgs{command: []string{"sudo", "-i", "-u", "workshop", "bash", "-c", "cd /project; exec bash"}}
+	args := &ExecArgs{command: []string{"bash"}}
 
 	if len(av) > 0 {
 		args.workshop = av[0]
@@ -259,7 +260,13 @@ func (c *CmdShell) Run(cmd *cobra.Command, av []string) error {
 		args.implicit = true
 	}
 
-	return exec(c.root, &ExecFlags{WorkingDir: "/project"}, args)
+	flags := &ExecFlags{
+		WorkingDir: "/project",
+		UserId:     1000,
+		GroupId:    1000,
+	}
+
+	return exec(c.root, flags, args)
 }
 
 func (c *CmdRun) Command() *cobra.Command {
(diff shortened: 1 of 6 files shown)


cmd/workshop/exec.go
```diff
@@ -251,7 +253,7 @@ $ workshop shell`,
 }
 
 func (c *CmdShell) Run(cmd *cobra.Command, av []string) error {
-	args := &ExecArgs{command: []string{"sudo", "-i", "-u", "workshop", "bash", "-c", "cd /project; exec bash"}}
+	args := &ExecArgs{command: []string{"sudo", "-i", "-u", "workshop", "bash", "-c", "cd /project; exec bash"}, shell: true}
```
dmitry-lyfar: Why not:

```
func (c *CmdShell) Run(cmd *cobra.Command, av []string) error {
	args := &ExecArgs{command: []string{"bash"}}

	if len(av) > 0 {
		args.workshop = av[0]
	} else {
		args.implicit = true
	}

	return exec(c.root, &ExecFlags{UserId: 1000, GroupId: 1000}, args)
}
```

cmd/workshop/exec.go
```diff
@@ -433,6 +436,22 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 		}
 	}
 
+	// Obtain an exec session as close to a real login shell as possible.
+	// This is required as an lxd exec call is a simple namespace exec, and LXD
+	// ignores the instance configuration by design:
+	// https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#user-groups-and-working-directory
+	if !args.shell && !args.script {
```
dmitry-lyfar: There is already a condition that checks if it is a script or not. So, in addition to logging:
```
	if args.script {
		logger.Debugf("Running script %q", args.command)
	} else {
		// Obtain an exec session as close to a real login shell as possible.
		// This is required as an lxd exec call is a simple namespace exec, and LXD
		// ignores the instance configuration by design:
		// https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#user-groups-and-working-directory
		command := []string{"sudo", "-E", "-u", "#" + strconv.Itoa(flags.UserId), "-g", "#" + strconv.Itoa(flags.GroupId)}
		command = append(command, environment...)
		command = append(command, "--", "bash", "-l", "-c", `exec -- "$0" "$@"`)
		args.command = append(command, args.command...)

		logger.Debugf("Running %q", args.command)
	}

```

internal/overlord/cmdstate/handlers.go
```diff
@@ -388,8 +389,15 @@ func (m *CommandManager) doInstallScript(task *state.Task, tomb *tomb.Tomb) erro
 		return errors.New("script not found")
 	}
 
+	var env []string
+	for k, v := range args.Environment {
```
dmitry-lyfar: The environment has already been built and will be passed to the exec API. We should not build a new environment just for this command as it ignores the API arguments. You can use `--preserve-env` flag of the sudo command to preserve the environment from the API.

dmitry-lyfar (changes requested):

cmd/workshop/exec.go
lachypjones: I wasn't sure whether you still wanted the real login shell, it was (at least at one point) a concern so I assumed it still was. If it no longer is, then that indeed helps clean things up nicely.

dmitry-lyfar:
I think this PR does have implications for the documentation. I would leave possible updates to the Tutorial, Reference or Explanation to @akcano as a separate PR if we find this required. We don't have details of the `exec` environment, nor of the workshop user as the default user, and it requires triaging. E.g. if you don't know this as a user, it would take some trial and error to install a tool (`golangci-lint`) locally in `setup-base`.

internal/overlord/cmdstate/handlers.go
```diff
@@ -389,7 +390,18 @@ func (m *CommandManager) doInstallScript(task *state.Task, tomb *tomb.Tomb) erro
...
 
 	path := filepath.Join(dirs.WorkshopScriptsDir, name)
-	command := []string{"bash", "-ue", "-o", "pipefail", path}
+	command := []string{"sudo",
+		"-u",
+		"#" + strconv.Itoa(args.UserId),
+		"-g", "#" + strconv.Itoa(args.GroupId),
+		"--preserve-env",
+		"--",
+		"bash",
```
jonathan-conder, line 399: I don't think this will read from `/etc/profile`, correct me if I'm wrong though.

I think `sudo -u ... bash -l ...` and the extra environment variables should be inserted in `lxdbackend.Backend.Exec` rather than the client. That will also make the SDK hooks use the same environment. Or `execCommand`, which affects the `startCommand` and `findmnt` (not really needed but doesn't make much difference)

lachypjones: On your first point, yes you're correct (thankyou!), I inadvertently removed a section when simplifying the command string,  looks like a good sign I need to write a test.  

To your second point, the reason for the client-side implementation was to keep the backend unopinionated in this respect. A different client may reasonably want a pure exec call. Apologies if i'm misunderstanding your point.

jonathan-conder (approved):

dmitry-lyfar (approved):

lachypjones:
As raised by Dmitry, there are some caveats to this PR that should be highlighted in documentation, namely:
- Workshop allows the user to specify the `uid` and `gid` used for an exec session. We enable lingering for the workshop user, which creates a persistent user-manager for uid #1000 (workshop) on boot, this creates side-effects of a normal login session which will not be present for other users. Namely:
  - XDG_RUNTIME_DIR is created
  - The dbus session bus service and socket are created
- As a partial consequence of the above, ssh/desktop interfaces will likely not work for other users. The socket needs to be created somewhere, and the standard directory for this is often the runtime dir. This could be fixed (and is likely worth looking into).

PR #286, opened 2025-02-04
Use lxc --force-local in snap remove hook

dmitry-lyfar:
# Description

Use lxc --force-local in snap remove hook to make sure the default project and a local Unix socket are used by the `lxc` commands.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/snap/local/hooks/remove
+++ b/snap/local/hooks/remove
@@ -1,46 +1,52 @@
 #!/bin/sh -e
 
+# Enforce local unix socket and the default project (if not specified).
+run_lxc() {
+    lxc --force-local "$@"
+}
+
 # The workshop services are stopped by the moment the remove 
 # hook is called. Therefore, for now, we walk the LXD projects created 
 # by the workshop and remove all instances/profiles/images/volumes.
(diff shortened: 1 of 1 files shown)


snap/local/hooks/remove
```diff
@@ -1,46 +1,50 @@
 #!/bin/sh -e
 
+# Enforce local unix socket and the default project (if not specified).
+alias run_lxc='lxc --force-local'
```
jonathan-conder: I was surprised that this works, but looks like `sh` will let you use aliases in scripts, whereas `bash` won't. Might be worth using a function for portability.

jonathan-conder (approved):

snap/local/hooks/remove
dmitry-lyfar: `bash` can do that too with `shopt -s expand_aliases` in non-interactive scripts. But yes, a function would make sense.

PR #287, opened 2025-02-04
Add pre-commit config for golangci-lint

jonathan-conder:
# Description

This should match what the workflow does, but will only check new code.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -196,6 +196,35 @@ Code structure
   This improves its structure and makes it easier to comprehend.
 
 
+Linting
+-------
+
+Code should be formatted consistently
+and avoid common pitfalls.
+Contributions will be checked for some of these issues
+using `golangci-lint <https://golangci-lint.run/>`_.
+To run these checks locally:
+
+.. code-block:: console
+
+   golangci-lint run
+
+
+Some issues can be fixed automatically:
+
+.. code-block:: console
+
+   golangci-lint run --fix
+
+
+If `pre-commit <https://pre-commit.com/index.html#install>`_ is available,
+:command:`git` can run these checks on every commit:
+
+.. code-block:: console
+
+   pre-commit install
+
+
 Testing
 -------
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -40,7 +40,6 @@ Environment setup
 
    .. code-block:: console
 
-
       export WORKSHOP=~/workshop
       workshop list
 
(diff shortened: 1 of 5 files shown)


dmitry-lyfar (approved):
Thanks! Looks good, there is some spelling issue.

docs/contributing.rst
```diff
@@ -197,6 +196,35 @@ Code structure
...
+Contributions will be checked for some of these issues
+using `golangci-lint <https://golangci-lint.run/>`_.
+To run these checks locally:
+
+.. code-block:: console
+
+   $ golangci-lint run
+
+
+Some issues can be fixed automatically using:
```
```suggestion
Some issues can be fixed automatically:
```

docs/contributing.rst
```diff
@@ -216,30 +244,42 @@ of ``Spread``:
...
 
 .. code-block:: console
 
-   go test --coverpkg=<./...|package> covermode=<set|count|atomic> -coverprofile=<OutputFile> <./...|package>
+   $ go test --coverpkg=<./...|package> covermode=<set|count|atomic> -coverprofile=<OutputFile> <./...|package>
+
 
-   // For example to measure coverage using all tests:
-   go test covermode=count -coverprofile=cover.out ./...
+For example to measure coverage using all tests:
```
```suggestion
For example, to measure coverage using all tests:
```

docs/contributing.rst
```diff
@@ -31,35 +31,34 @@ Environment setup
 
    .. code-block:: console
 
-      go install ./...
-      export WORKSHOP=~/workshop
-      export WORKSHOP_DEBUG=1
-      workshopd run --create-dirs
+      $ go install ./...
```
akcano: This file has to be rendered on GitHub, just as README does.
However, GitHub does a poor job of rendering these code blocks and doesn't prevent the prompts from being copied (unlike Sphinx/RTD), so it's best to drop them here.

akcano (changes requested):

docs/contributing.rst
jonathan-conder: I reckon there should be a way to insert them as part of the render. Probably not gonna look into it right now though.

jonathan-conder: I had a bit of a crack at it over here: https://github.com/canonical/workshop/commit/92342b4b62ae8330cc20e169769f961307a475e5. Let me know if you think it's worth including. I should have added a filename-based filter. Shouldn't be too hard but I'll wait to see what you think.

akcano: OK, let's try to live with this and see if it works out.

jonathan-conder: so I should add that to the PR?

dmitry-lyfar: @jonathan-conder yes, please.

dmitry-lyfar (approved):

PR #289, opened 2025-02-05
Prevent sketching when the workshop is not ready

lachypjones:
# Description
Fixes two small issues:
- If a workshop is stopped during a sketch, the subsequent partial refresh will fail (workshop is not running)
- If a workshop is pending during a sketch, the current wait will be aborted (due to sketch auto-aborting)

# Self review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/sketch_test.go
+++ b/cmd/workshop/sketch_test.go
@@ -23,8 +23,64 @@ type workshopSketch struct {
 
 var _ = check.Suite(&workshopSketch{})
 
-var mockWorkshopWithSdksReady = `{"type":"sync","status-code":200,"status":"OK","result":{"name":"ws","base":"ubuntu@22.04","project-id":"42424242","status":"Ready","sdks":[{"name":"go","version":"1.8.0","channel":"latest/edge","revision":"1","build-time":"2017-02-19T17:23:05.592623Z","install-time":"2017-03-22T09:01:00.0Z"},{"name":"sketch","channel":"","revision":"x1","install-time":"2017-03-22T09:01:00.0Z"}],"notes":["missing-project"],"path":"/home/project/.workshop/ws.yaml"},"warning-timestamp":"2017-03-22T10:01:00.0Z","warning-count":1}`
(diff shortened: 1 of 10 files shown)


cmd/workshop/sketch_test.go
```diff
@@ -23,6 +24,7 @@ type workshopSketch struct {
 
 var _ = check.Suite(&workshopSketch{})
 
+var mockWorkshopWithSdksReady = strings.Replace(mockWorkshopWithSdks, "Error", "Ready", 1)
```
dmitry-lyfar: I would not couple it for no reason this way. So, either a function, that returns a JSON of a workshop with the required status to minimise duplication, or just duplicate and make the status "Error" (kind of ok in tests to have some duplication).

cmd/workshop/sketch_test.go
```diff
@@ -549,3 +550,34 @@ func (m *workshopSketch) TestSketchesEmpty(c *check.C) {
 
 	c.Assert(m.stdout.String(), check.Matches, "")
 }
+
+func (m *workshopSketch) TestSketchSdkWorkshopStatusNotReady(c *check.C) {
+	cmd := &CmdSketch{root: &CmdRoot{}, stash: true}
+
+	status := []string{"Pending", "Error", "Off", "Stopped"}
```
dmitry-lyfar: We don't have the "Off" status anymore. If there is only a workshop file, it won't be returned in the API response. 

Can you, in the scope of this PR, remote the `OffStatus` status from the server (and update its corresponding tests and functions that still use it)?

cmd/workshop/sketch_test.go
```diff
@@ -23,6 +23,7 @@ type workshopSketch struct {
 
 var _ = check.Suite(&workshopSketch{})
 
+var mockWorkshopWithSdksReady = `{"type":"sync","status-code":200,"status":"OK","result":{"name":"ws","base":"ubuntu@22.04","project-id":"42424242","status":"Ready","sdks":[{"name":"go","version":"1.8.0","channel":"latest/edge","revision":"1","build-time":"2017-02-19T17:23:05.592623Z","install-time":"2017-03-22T09:01:00.0Z"},{"name":"sketch","channel":"","revision":"x1","install-time":"2017-03-22T09:01:00.0Z"}],"notes":["missing-project"],"path":"/home/project/.workshop/ws.yaml"},"warning-timestamp":"2017-03-22T10:01:00.0Z","warning-count":1}`
```
dmitry-lyfar: I think it's time to tidy up out JSON responses in the client's tests. See how this is done (formatting) in refresh_test.go (e.g. line 31 or 42).

dmitry-lyfar (approved):

PR #291, opened 2025-02-09
Fix sshAgent and Desktop profile equality checks

lachypjones:
# Description
As reproduced and solved by Dmitry, connecting sshAgent and Desktop in a workshop would result in the former having it's remove logic executed. This fixes the equality check causing this behavior. 

This also modifies the desktop removal logic to return an error in cases where the `profile.d` entry no longer exists. This now matches the behavior of ssh-agent.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/builtin/desktop_test.go
+++ b/internal/interfaces/builtin/desktop_test.go
@@ -4,13 +4,14 @@ import (
 	"os/user"
 	"path/filepath"
 
+	"gopkg.in/check.v1"
+
 	"github.com/canonical/workshop/internal/dirs"
 	"github.com/canonical/workshop/internal/interfaces"
 	"github.com/canonical/workshop/internal/interfaces/builtin"
 	"github.com/canonical/workshop/internal/interfaces/lxd_device"
 	"github.com/canonical/workshop/internal/testutil"
 	"github.com/canonical/workshop/internal/workshop"
-	"gopkg.in/check.v1"
 )
 
 type desktopSuite struct {
@@ -76,11 +77,15 @@ exit 0`)
 	defer fake.Restore()
 
 	c.Assert(deviceSpec.AddConnectedPlug(s.iface, connectedPlug, connectedSlot), check.IsNil)
-	expectedProxy := &workshop.Desktop{}
-	expectedProxy.Wayland = &workshop.ProxyEntry{}
-	expectedProxy.Wayland.Name = "consumer-wayland"
-	expectedProxy.Wayland.Connect = "/tmp/wayland-1"
-	expectedProxy.Wayland.Listen = "/run/user/1000/wayland-1"
+	expectedProxy := &workshop.Desktop{
+		Wayland: &workshop.ProxyEntry{
+			Name: "consumer-wayland",
+			Connect: workshop.ProxyTarget{
+				Address:  "/tmp/wayland-1",
+				Protocol: "unix"},
+			Listen: workshop.ProxyTarget{
+				Address:  "/run/user/1000/wayland-1",
+				Protocol: "unix"}}}
 	c.Assert(deviceSpec.Profile.Desktop, check.DeepEquals, expectedProxy)
 }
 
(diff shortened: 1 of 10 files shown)


internal/interfaces/builtin/desktop_test.go
```diff
@@ -287,3 +288,9 @@ exit 0`)
 
 	c.Assert(deviceSpec.AddConnectedPlug(s.iface, connectedPlug, connectedSlot), check.ErrorMatches, "XDG_RUNTIME_DIR is either empty.*")
 }
+
+// Make sure desktop has two fields (Wayland and X11). If this changes, the
+// equality check in lxd_device/backend.go will need to be modified
+func (s *desktopSuite) TestDesktopValidateMethodCount(c *check.C) {
```
dmitry-lyfar: The comparison (Equal) logic can be put into a desktop's method and live next to the struct itself ensuring locality and no need for this kind of tests.

internal/interfaces/lxd_device/backend.go
```diff
@@ -234,7 +234,12 @@ func installSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent, workshop str
 }
 
 func removeSshAgent(fs workshop.WorkshopFs, dev workshop.SshAgent) error {
-	return fs.Remove(filepath.Join("/etc/profile.d", dev.Name+".sh"))
+	if err := fs.Remove(filepath.Join("/etc/profile.d", dev.Name+".sh")); err != nil {
+		if !errors.Is(err, afero.ErrFileNotFound) {
```
dmitry-lyfar: If the agent is installed the profile is there, if it's uninstalled, the profile is not there. What are the possible use cases that makes this to ignore the file existence?

lachypjones: A user may remove the file in profile.d. The remove logic should not fail if the file is not present, as the outcome is identical either way. 

In any case, I think this is an argument in favor of moving the environment population responsibility from linux primatives to things that are not mutable by users.

internal/interfaces/builtin/desktop_test.go
lachypjones: I'd still argue in favor of the test (as it's still possible to miss, there's no enforcement external to locality), but happy to take your advice here. 

My reasoning for not using an 'Equal' method was I didn't see one in use elsewhere (nor in Snapd for that matter), so I assumed it wasn't an accepted norm here. More than happy to implement as it seemed like a nicer solution to me.

internal/interfaces/lxd_device/backend.go
dmitry-lyfar: This logic will hide possible issues with implementation, i.e. it will keep ignoring the file non-existence when it's actually our mistake and it does not exist not because a user removed it but because we haven't created it (or removed it ourselves when we should not have to). `removeSshAgent` must not protect the workshop from the integrity issues that a user may introduce (i.e. from `sudo rm -f /etc/profile.d/desktop.sh`).

There are ways to deal with the integrity to make it more fool-proof, but not at this level. E.g. SDK volumes are now read-only and that helps if someone tries to change it.

internal/interfaces/builtin/desktop_test.go
dmitry-lyfar: It's local (adjacent) to the struct itself and can accommodate comparison with `nil`. I think snapd favours `reflect.DeepEqual` which is another possible way. But there are still things like `slotDeviceAttrEqual`.

internal/interfaces/lxd_device/backend.go
```diff
@@ -409,15 +408,18 @@ func (b *Backend) Setup(ctx context.Context, sdkInfo sdk.Ref, repo *interfaces.R
...
+
 		if prevp.Agent != nil {
 			if spec.Profile.Agent == nil || *prevp.Agent != *spec.Profile.Agent {
 				if err = removeSshAgent(fs, *prevp.Agent); err != nil {
 					return err
 				}
 			}
 		}
+
 		if prevp.Desktop != nil {
```
dmitry-lyfar: The Equals function should handle nil cases, thus the following should read as:
```
if !prevp.Desktop.Equal(spec.Profile.Desktop)
```

internal/workshop/device.go
```diff
@@ -35,6 +35,13 @@ type Desktop struct {
 	X11     *ProxyEntry
 }
 
+func (d *Desktop) Equals(other *Desktop) bool {
```
dmitry-lyfar:
```
if d == nil || other == nil {
   return d == other
}
//...
```

internal/workshop/device.go
```diff
@@ -30,11 +35,27 @@ type SshAgent struct {
...
+
+	return *s == *other
+}
+
 type Desktop struct {
 	Wayland *ProxyEntry
 	X11     *ProxyEntry
 }
 
+func (d *Desktop) Equals(other *Desktop) bool {
```
dmitry-lyfar: I think we landed at naming it `Equal`.

internal/workshop/device.go
```diff
@@ -30,11 +35,27 @@ type SshAgent struct {
...
+	if d == nil || other == nil {
```
dmitry-lyfar, line 52: So, if these are both nil, they are not equal?

dmitry-lyfar (approved):

PR #284, opened 2025-02-03
Add tunnel interface

jonathan-conder:
# Description

Adds the `tunnel` interface, allowing communication between network services in the workshop and on the host. This adds extra data to the `workshop info` API and output. It's also included in the `sketch-sdk` template.

Interfaces which don't define plug/slot sanitize functions are no longer allowed to have attributes. This is to prevent users from faking meaningful dynamic attributes like `auto-explicit`. The downside is that it may be harder to introduce new attributes in future (e.g. SDKs might be tied to new versions of Workshop).

The spread tests require `netcat`, and `systemd` version 230 or later. Should be fine since the prepare step already depends on `netcat`, and `focal` has `system` 245.

## Known issues:

- If connect fails, the LXD profile is still updated. A second connect will erroneously succeed (but can then be safely disconnected). Seems unlikely to be fully resolved upstream so we'll have to work around this at some point.
- Disconnecting a unix socket plug doesn't remove the socket: https://github.com/canonical/lxd/issues/14972.
- Port conflict error messages are currently inconsistent: https://github.com/canonical/lxd/issues/15023.
- Proxy devices closed too early: https://github.com/canonical/lxd/issues/15024.
- `workshop info` uses the current connections instead of the SDK profiles. This complicates the code, but is required for now because the system SDK is hidden from `workshop info`.
- Can't listen on port 0. There's not much point doing so unless we provide a way for users to query the randomly allocated port. I've already created a lot of work for the LXD team, so this can wait until we have users who want this feature.

# Self-review quick check

 * [x] Make decisions that cost a lot to reverse explicit in the PR description.
 * [x] Avoid nested conditions.
 * [x] Delete dead code and redundant comments.
 * [x] Normalise symmetries by sticking to doing identical things identically. 
 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 // one way to handle multiple returns
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```
 * [x] Check that coupled code elements, files, and directories are adjacent. For example, test data is stored as close as possible to a test.
 * [x] Put variable declaration and initialisation together.
 * [x] Divide large expressions into digestable and self-explanatory ones. Use multiple variables if required.
 * [x] Put a blank line between two logically different chunks of code.
 * [x] Follow the [style guide](https://github.com/canonical/workshop/tree/main/docs/contributing.rst#error-messages) for new error messages.

## Docs

* [x] I have checked and added or updated relevant documentation.
[...]

PR #294, opened 2025-02-13
Switch to new path spec

lachypjones:
# Description
Implements the following:

1. Changes the `workshop` content path to respect `XDG_DATA_HOME`, falling back to `~/.local/share` if unset:
`$HOME/.local/share/workshop` &rarr; `$XDG_DATA_HOME/workshop/`

2. Moves project content into a unique directory per workshop: 
`.../project/<id>/<content-type>/<workshop>_<sdk>_<plug>.sdk` &rarr; `.../id/<id>/<workshop>/<content-type>`

3. In-line with the above, renames content paths appropriately:
`.../mount/<workshop>_<sdk>_<plug>` &rarr; `.../mount/<sdk>/<plug>`
`.../sdk/sketch/<workshop>/*` &rarr; `.../sdk/sketch/*`

# Testing changes
Querying mount paths now relies on systemd, this is required to determine the calling user's environment. As such, lingering has been enabled for the 'ubuntu' user during spread runs.

# Breaking changes
As the content directories for sketch and mount have changed, workshops will need to be removed in order to use the new paths. If using the Snap, this removal will happen automatically. 

# Self-review quick check
 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


akcano (approved):

cmd/workshop/info.go
```diff
@@ -60,14 +60,22 @@ $ workshop info`,
...
+//	/home/dmitry/.local/share/workshop/id/17942561/ws/mount/go_mod-cache
 //
 // becomes:
 //
-//	.../17942561/mount/albert_go_mod-cache.sdk
-func shortenDefaulPath(source string) string {
-	user, _ := user.Current()
-	defaultPathPrefix := filepath.Join(user.HomeDir, ".local", "share", "workshop", "project")
+//	.../17942561/mount/go_mod-cache
+func shortenDefaultPath(source string) string {
```
dmitry-lyfar:
```
       # this should come from the user not from env?
       home, err := ...
       // add error handling

       xdg, err := os.LookupEnv("XDG_DATA_HOME")
       // add error handling

	if xdg == "" {
		// If XDG_DATA_HOME is unset, fallback to $HOME/.local/share
		xdg = filepath.Join(home, ".local", "share")
	}
	defaultPathPrefix := filepath.Join(xdg, "workshop", "id")
```

internal/sdk/sdk.go
```diff
@@ -477,33 +479,68 @@ func SdkHookPath(sdkName, hookName string) string {
 	return filepath.Join(SdkHooksDir(sdkName), hookName)
 }
 
-func ProjectUserData(homedir, pid string) string {
-	return filepath.Join(homedir, ".local", "share", "workshop", "project", pid)
+func ProjectDir(usr *user.User, pid string) (string, error) {
```
dmitry-lyfar: This is not a `project directory`, the project directory is what sits under `/project`.

internal/sdk/sdk.go
```diff
@@ -477,33 +479,68 @@ func SdkHookPath(sdkName, hookName string) string {
...
+
+	dataDir := env["XDG_DATA_HOME"]
+	if dataDir != "" {
+		path = dataDir
+	}
+
+	return filepath.Join(path, "workshop", "id", pid), nil
+}
+
+func WorkshopDir(usr *user.User, pid, w string) (string, error) {
```
dmitry-lyfar: This is also not quite a `WorkshopDir`, see how this reads `sdk.WorkshopDir(...)`? Given the change it should rather be moved to `workshop.Something()`?

internal/sdk/sdk.go
```diff
@@ -477,33 +479,68 @@ func SdkHookPath(sdkName, hookName string) string {
...
+	path, err := WorkshopDir(usr, pid, w)
+	if err != nil {
+		return path, err
+	}
+	return filepath.Join(path, filepath.Join(toAppend...)), nil
 }
 
-func ProjectSketchSdkDir(homedir, pid string) string {
-	return filepath.Join(ProjectUserData(homedir, pid), "sdk", "sketch")
+func WorkshopMountDir(usr *user.User, pid, w string) (string, error) {
```
dmitry-lyfar: `wrapWorkshopDir` seems like a redundant func. Other similar funcs just call functions that form a higher-level dir. Alternatively, you can have a rootdir as an argument here.

internal/sdk/sdk.go
```diff
@@ -477,33 +479,68 @@ func SdkHookPath(sdkName, hookName string) string {
...
+	return wrapWorkshopDir(usr, pid, w, "mount")
 }
 
-func WorkshopSketchSdk(homedir, pid, wp string) string {
-	return filepath.Join(ProjectSketchSdkDir(homedir, pid), wp)
+func WorkshopSketchSdkDir(usr *user.User, pid, w string) (string, error) {
```
dmitry-lyfar: I think we should drop `Workshop` from the func name.

dmitry-lyfar: also, am not sure about the dependency on the user.User.

internal/systemd/user.go
```diff
@@ -29,3 +31,17 @@ func UserEnvironment(user *user.User) (map[string]string, error) {
 	rawEnv := strings.FieldsFunc(string(out), func(r rune) bool { return r == '\n' })
 	return osutil.ParseEnvironment(rawEnv)
 }
+
+func FakeUserEnvironment(env map[string]string) func() {
```
dmitry-lyfar: FakeUserEnvironment should accept a function as it fakes a function?

snap/local/hooks/post-refresh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh -e
```
dmitry-lyfar, line 1: Is there a way to link or call the remove hook from here to avoid duplicating logic?

internal/sdk/sdk.go
dmitry-lyfar: I think XDG needs to be provided as an argument (e.g. rootdir) and be separated from the logic that builds paths for workshops and SDKs data. So we will be able to change them independently and, say, not rely on `user.User`.

internal/interfaces/builtin/ssh_agent_test.go
```diff
@@ -67,10 +59,8 @@ slots:
 	connectedSlot := interfaces.NewConnectedSlot(slot, nil, nil)
 	deviceSpec := lxd_device.NewSpecification(&testuser, "consumer")
 
-	fake := testutil.FakeCommand(c, "sudo", `
-echo "SSH_AUTH_SOCK=/tmp/dir/ssh"
-exit 0`)
-	defer fake.Restore()
+	env := map[string]string{"SSH_AUTH_SOCK": "/tmp/dir/ssh"}
```
dmitry-lyfar: Why do we move to use FakeUserEnvironment from faking the command?

Also, you don't really need this it seems, we have `testutil.FakeCommand`?

I understand the point of faking a user environment via Get and Set env, i.e. where the tested code calls Get. We discussed that before. But here it only happens to read the environment as a result,  but it calls a whole different command. If that command changes in the code (i.e. to be something else than `sudo ...`), the tests would still continue to work fine as they rely on the env map.

internal/interfaces/builtin/mount_test.go
```diff
@@ -45,174 +50,125 @@ var _ = check.Suite(&mountSuite{
...
 func (s *mountSuite) TestMountInterfaceReadOnlyBool(c *check.C) {
-	plug := builtin.MockPlug(c, `name: consumer
-base: ubuntu@22.04
-plugs:
- mount:
-  workshop-target: /project/training
-  read-only: true
-`, s.projectId, "ws", "consumer", "mount")
-	c.Assert(interfaces.BeforePreparePlug(s.iface, plug), check.IsNil)
+	s.plug.Attrs["workshop-target"] = "/project/training"
```
dmitry-lyfar: Is there a connection between the PR and these changes to the tests? 

I find verbosity in tests useful as long as it localises your reading experience if one fails. You don't have to go back and forth between the test and its setup. E.g. see this https://github.com/canonical/workshop/pull/284/files#diff-bf636b900a734aa47cd1cb0e144d04e18e9a3e9986d542a7b6943b21eae682ee.

We would either need to redo all those ones or have a good reason to change these.

snap/local/hooks/post-refresh
lachypjones: I can make it a symlink. Note however that I avoid removing the images, and we need to cleanup the workshop state.

internal/interfaces/builtin/mount_test.go
lachypjones: No problem, I'll revert that change. It seemed like every test duplicated exactly the same stub, with the occasional extra attribute. 

I understand the advantage of verbosity, however I felt like knowing the base stub and then only having the exact attributes that were modified part of each test made what was been tested clearer.

internal/interfaces/builtin/ssh_agent_test.go
lachypjones: Because ssh-agent calls systemd.UserEnvironment, I thought it was more appropriate that we faked the specific function that needed to be faked.

internal/systemd/user.go
lachypjones: No problem.
I was thinking more along the lines of "it fakes an environment so it should take an environment". However I realise that FakeUserLookup for example still takes a function so there's good reason to keep the Fakes consistent.

internal/interfaces/builtin/ssh_agent_test.go
dmitry-lyfar: But if we call a linux command elsewhere, we fake that command's output. So, what I'm trying to understand how this one is different?

cmd/workshop/info.go
lachypjones: I was trying to be consistent with the behavior we had previously (ie. ignoring errors). As the function only modified the output, ignoring errors and falling back to the full path felt appropriate. But i'm happy to change.

Why pull the home dir from current user? It's client code so the client environment should be appropriate no?

internal/interfaces/builtin/ssh_agent_test.go
lachypjones: That's reasonable. I'm happy to go back, I only thought that it was slightly more clear as to what we were doing with an explicit faked function call.

cmd/workshop/info.go
```diff
@@ -154,20 +165,31 @@ func (c *CmdInfo) Run(cmd *cobra.Command, av []string) error {
 			if len(sk.Mounts) > 0 {
 				fmt.Fprintf(w, "    mounts:\n")
 				slices.SortFunc(sk.Mounts, func(a, b *client.Mount) int { return cmp.Compare(a.Plug.Name, b.Plug.Name) })
+				buf := &bytes.Buffer{}
 				for _, mount := range sk.Mounts {
+					hostSource, err := shortenDefaultPath(mount.HostSource)
+					if err != nil {
+						return fmt.Errorf("cannot build short mount path: %w", err)
```
dmitry-lyfar: This (path shortening) is our internal issue that should not go to the end user in a such a detailed message. We should `cannot list workshops: %v`.

cmd/workshop/info.go
```diff
@@ -154,20 +165,31 @@ func (c *CmdInfo) Run(cmd *cobra.Command, av []string) error {
 			if len(sk.Mounts) > 0 {
 				fmt.Fprintf(w, "    mounts:\n")
 				slices.SortFunc(sk.Mounts, func(a, b *client.Mount) int { return cmp.Compare(a.Plug.Name, b.Plug.Name) })
+				buf := &bytes.Buffer{}
```
dmitry-lyfar: Why do we need a buffer? tabWriter is buffered already.

docs/tutorial/index.rst
```diff
@@ -298,7 +298,7 @@ to see what went into your workshop:
          installed:  1.23.0  2024-08-15  (51)
          mounts:
            mod-cache:
-             host-source:      .../6b79e889/mount/dev_go_mod-cache.sdk
+             host-source:      .../6b79e889/dev/mount/go_mod-cache
```
dmitry-lyfar:
```suggestion
             host-source:      .../6b79e889/dev/mount/go/mod-cache
```

internal/overlord/workshopstate/request.go
```diff
@@ -322,7 +322,13 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	sketchdir := sdk.WorkshopSketchSdkCurrent(usr.HomeDir, pid, wp)
+
+	rootDir, err := workshop.UserDataRootDir(usr)
```
dmitry-lyfar: In most cases where UserDataRootDir is called, it is preceded by the user lookup. I would just combine these two in one and accept the user's name as an argument. That would make the client to not user `osutil.MaybeSudoUser`, but it should be fine as it is not really necessary.

dmitry-lyfar (commented):
See my suggestions in the comments + I have pushed a commit to demo the idea, feel free to reset this.

cmd/workshop/info.go
lachypjones: The reason I don't use tabWriter directly is because we don't want to write any output if we can't display an accurate list of workshops.
bytes.Buffer seemed like a better alternative to strings.Builder as it provides the io.Writer interface, and returns bytes, not a string. 

I can of course use a slice instead, the byte buffer just seemed more ergonomic.

cmd/workshop/info.go
```diff
@@ -60,18 +61,27 @@ $ workshop info`,
...
 //
-//	.../17942561/mount/albert_go_mod-cache.sdk
-func shortenDefaulPath(source string) string {
-	user, _ := user.Current()
-	defaultPathPrefix := filepath.Join(user.HomeDir, ".local", "share", "workshop", "project")
+//	.../17942561/mount/go_mod-cache
+func shortenDefaultPath(source string) (string, error) {
+	home, _ := os.LookupEnv("HOME")
+	if home == "" {
+		return "", fmt.Errorf("cannot determine user home directory")
```
dmitry-lyfar: I would suggest a different error message here, so we don't end up with:

```
cannot list workshops: cannot determine user home directory
```

something like:
```
cannot list: $HOME is not set
```

dmitry-lyfar: Note `cannot list workshops` - > `cannot list`, similar to `cannot refresh`.

snap/local/hooks/lxc-utils
```diff
@@ -0,0 +1,58 @@
+# Enforce local unix socket and the default project (if not specified).
```
dmitry-lyfar, line 1: This is nice.

cmd/workshop/info_test.go
```diff
@@ -168,7 +170,7 @@ var mockWorkshopWithMounts = `{"type":"sync","status-code":200,"status":"OK","re
                 "plug":"plug-name"
             }
         },{
-            "host-source":"%s/.local/share/workshop/project/17942561/mount/ws_go_mod-cache.sdk",
+            "host-source":"%s/workshop/id/17942561/ws/mount/go_mod-cache",
```
dmitry-lyfar:
```suggestion
            "host-source":"%s/workshop/id/17942561/ws/mount/go/mod-cache",
```

cmd/workshop/info_test.go
```diff
@@ -180,12 +182,70 @@ var mockWorkshopWithMounts = `{"type":"sync","status-code":200,"status":"OK","re
...
+project:  %s
+status:   ready
+notes:    -
+sdks:
+  go:
+    tracking:   latest/edge
+    installed:  1.8.0  2017-02-19  \(1\)
+    mounts:
+      plug-default:
+        host-source:      .../17942561/ws/mount/go_mod-cache
```
dmitry-lyfar:
```suggestion
        host-source:      .../17942561/ws/mount/go/mod-cache
```

cmd/workshop/info_test.go
```diff
@@ -180,12 +182,70 @@ var mockWorkshopWithMounts = `{"type":"sync","status-code":200,"status":"OK","re
     }]
 }}`
 
-func (m *workshopInfo) TestWorkshopInfoWithSdkMounts(c *check.C) {
+var mockWorkhsopWithMountsResponse = `name:     ws
```
dmitry-lyfar: That looks like `info` output, not response?

internal/daemon/api_test.go
```diff
@@ -91,22 +87,30 @@ func (s *apiSuite) SetUpTest(c *check.C) {
...
+	s.ctx = context.WithValue(ctx, workshop.ContextUser, s.user.Username)
 
 	_, _, err = s.b.CreateOrLoadProject(s.ctx, s.project.Path)
 	c.Assert(err, check.IsNil)
 
 	s.restoreSanitize = sdk.MockSanitizePlugsSlots(func(sdkInfo *sdk.Info) {})
 	s.secBackend = &ifacetest.TestSecurityBackend{BackendName: "api-suite"}
 	s.restoreSecBackend = ifacestate.MockSecurityBackends([]interfaces.SecurityBackend{s.secBackend})
+
+	xdgDir := c.MkDir()
```
dmitry-lyfar: Do we have to set it for the API tests? This logic (xdg from env vs. default path) is tested in the unit test, so here, we should rely on the code figuring that out and using the user's home dir.

internal/osutil/user_env.go
```diff
@@ -1,4 +1,4 @@
-package systemd
+package osutil
```
dmitry-lyfar: I think it's OK for this to live in user.go under osutil package.

internal/osutil/user_env_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (s *sdUserTestSuite) TestUserAndEnv(c *check.C) {
+	envIn := map[string]string{
+		"WORKSHOP_TEST_ENV_KEY":   "VALUE",
+		"WORKSHOP_TEST_ENV_EMPTY": "",
+	}
+
+	for k, v := range envIn {
+		// Set environment
+		cmd := exec.Command("systemctl", "--user", "set-environment", k+"="+v)
```
dmitry-lyfar: This should be fine, but if the test fails on assertion, the variables will remain unset? You can use c.Check to ensure that the environment will be cleaned up properly even if some checks have failed.

internal/overlord/workshopstate/handlers_test.go
```diff
@@ -81,17 +82,15 @@ func (s *workshopHandlers) SetUpTest(c *check.C) {
...
-		return u, nil
-	}, &workshop.LookupUsername)
+	s.user = &user.User{
+		HomeDir: c.MkDir(),
+	}
+	s.lookupUserRestore = osutil.FakeUserLookup(func(name string) (*user.User, error) {
+		return s.user, nil
+	})
+
+	s.sudoRestore = testutil.FakeCommand(c, "sudo", `
```
dmitry-lyfar: What do we fake `sudo` for here? I suppose that given that we moved UserAndEnv to a package with its own test, we can fake it similarly to how it was done in the API layer.

internal/overlord/workshopstate/manager_test.go
```diff
@@ -41,16 +43,12 @@ func (s *managerSuite) SetUpTest(c *check.C) {
...
-			Uid:      "1000",
-			Gid:      "1000",
-			HomeDir:  c.MkDir(),
-		}
-		return u, nil
-	}, &workshop.LookupUsername)
+	s.lookupUserRestore = osutil.FakeUserLookup(func(name string) (*user.User, error) {
+		return &user.User{HomeDir: c.MkDir()}, nil
+	})
+	s.sudoRestore = testutil.FakeCommand(c, "sudo", `
```
dmitry-lyfar: Same here, also these tests will likely be moved under the API layer tests cases in other branches which uses `Fake...`

cmd/workshop/info.go
```diff
@@ -87,6 +87,17 @@ func (c *CmdInfo) Run(cmd *cobra.Command, av []string) error {
 		return err
 	}
 
+	home, _ := os.LookupEnv("HOME")
+	if home == "" {
+		return fmt.Errorf("cannot show info: $HOME is not set")
```
jonathan-conder: I was going to say we should fall back to the `HomeDir` of `user.Current()` before returning an error. But that doesn't seem to comply with the XDG spec.

Since we're doing the spec, why not use the existing dependency https://github.com/adrg/xdg?

cmd/workshop/info.go
```diff
@@ -87,6 +87,17 @@ func (c *CmdInfo) Run(cmd *cobra.Command, av []string) error {
...
+	}
+
+	xdg, _ := os.LookupEnv("XDG_DATA_HOME")
```
jonathan-conder, line 99: I realise this is very late to be asking this, but have we considered `XDG_STATE_HOME`? The spec is vague on this but I think `XDG_DATA_HOME` is like `/usr/share` and `XDG_STATE_HOME` is like `/var/lib`.

cmd/workshop/info_test.go
```diff
@@ -180,12 +182,70 @@ var mockWorkshopWithMounts = `{"type":"sync","status-code":200,"status":"OK","re
...
 	cmd := &CmdInfo{root: &CmdRoot{}}
 	workshop := "ws"
 	n := 0
-	user, err := user.Current()
+	home := "/home/testuser"
+	xdg := filepath.Join(home, ".local", "share")
+	defer os.Setenv("XDG_DATA_HOME", os.Getenv("XDG_DATA_HOME"))
+	os.Setenv("XDG_DATA_HOME", "")
+	defer os.Setenv("HOME", os.Getenv("HOME"))
+	os.Setenv("HOME", home)
```
jonathan-conder, line 212: This part would have to change if we use the `xdg` package. I think that's a good reason to switch in itself - `Getenv`/`Setenv` can race with raw calls (e.g. in a C library) even if the variables are unrelated.

The cleanup also has a side effect of setting the variables to the empty string (if originally unset).

Actually maybe this test isn't necessary if we assume `xdg` has its own tests. If you do remove it, can probably move the info output string back inline.

docs/how-to/ros2/ros2-create-workshop.rst
```diff
@@ -154,17 +154,17 @@ although :command:`workshop refresh` rebuilds the workshop from scratch,
 including any SDK updates.
 
 The host-mapped contents of the workshop can actually be seen
-in |ws_markup|'s default content directory on the host
+in |ws_markup|'s default mount directory on the host
```
jonathan-conder, line 157: I think "content directory" is consistent throughout the docs, did you change all of them? I'm also not sure that "mount directory" is a better term, e.g. it could refer to the mount point inside the workshop.

internal/interfaces/builtin/desktop_test.go
```diff
@@ -68,13 +65,10 @@ slots:
...
-	deviceSpec := lxd_device.NewSpecification(&testuser, "consumer")
+	deviceSpec, err := lxd_device.NewSpecification(testuser.Name, "consumer")
+	c.Assert(err, check.IsNil)
 
-	fake := testutil.FakeCommand(c, "sudo", `
-echo "XDG_RUNTIME_DIR=/tmp"
-echo "WAYLAND_DISPLAY=wayland-1"
-exit 0`)
-	defer fake.Restore()
+	s.env["WAYLAND_DISPLAY"] = "wayland-1"
```
jonathan-conder: This approach of modifying the existing object feels a bit brittle to me. Probably OK because it's just a test, but I'd prefer just calling `FakeUserAndEnv` in each test separately. Or maybe implement as a callback rather than a static `map` object.

internal/interfaces/builtin/desktop_test.go
```diff
@@ -10,39 +10,36 @@ import (
...
+	restoreUserEnv func()
+	env            map[string]string
 }
 
 var _ = check.Suite(&desktopSuite{
 	iface: builtin.MustInterface("desktop"),
 })
 
 func (s *desktopSuite) SetUpTest(c *check.C) {
+	s.env = map[string]string{"XDG_RUNTIME_DIR": "/tmp"}
```
jonathan-conder: why not `c.Mkdir()`?

internal/interfaces/builtin/tunnel_test.go
```diff
@@ -949,6 +965,8 @@ slots:
 		Direction: workshop.WorkshopToHost,
 	}
 	c.Check(deviceSpec.Profile.Tunnels, check.DeepEquals, []workshop.Tunnel{{ProxyEntry: expectedEntry}})
+
+	testuser.Uid = ""
```
jonathan-conder: I'm much happier with this approach overall, but I think this line should be a `defer`. Or preferably just call `FakeUserAndEnv` again if that works.

internal/interfaces/lxd_device/backend.go
```diff
@@ -357,7 +351,7 @@ func (b *Backend) Setup(ctx context.Context, sdkInfo sdk.Ref, repo *interfaces.R
 	if !ok {
 		return fmt.Errorf("context key user not found")
 	}
-	user, err := workshop.LookupUsername(uname)
+	user, err := osutil.UserLookup(uname)
```
jonathan-conder: Is it possible to use `spec.User` here? If not I think it could use a comment explaining why we need to check again.

internal/osutil/user.go
```diff
@@ -167,6 +171,65 @@ func UserMaybeSudoUser() (*user.User, error) {
...
+	UserLookup = f
+	return func() { UserLookup = oldUserLookup }
+}
+
+func FakeUserLookupGroup(f func(name string) (*user.Group, error)) func() {
+	oldUserLookupGroup := UserLookupGroup
+	UserLookupGroup = f
+	return func() { UserLookupGroup = oldUserLookupGroup }
+}
+
```
jonathan-conder: I don't understand why these have moved out of `export_test`

internal/interfaces/lxd_device/tests/integration/backend_test.go
```diff
@@ -82,10 +82,25 @@ func defaultTestDevices() map[string]map[string]string {
...
+	}
+	f.env = make(map[string]string)
+
+	f.restoreUser = osutil.FakeUserLookup(func(name string) (*user.User, error) {
+		return f.usr, nil
+	})
+
+	f.restoreUserEnv = osutil.FakeUserAndEnv(func(name string) (*user.User, map[string]string, error) {
+		return f.usr, f.env, nil
+	})
```
jonathan-conder: I feel like `FakeUserAndEnv` should delegate to `FakeUserLookup` to avoid this. Alternatively make it `FakeUserEnv`, i.e. only change the env.

internal/interfaces/lxd_device/tests/integration/backend_test.go
```diff
@@ -272,14 +276,16 @@ func (f *backendDeviceSuite) TestSetupHostWorkshopMounts(c *check.C) {
 	b := lxd_device.Backend{}
 	cref := sdk.Ref{ProjectId: "42424242", Workshop: "test", Sdk: "consumer"}
 
+	f.env["XDG_RUNTIME_DIR"] = c.MkDir()
```
jonathan-conder: See earlier comment on overriding the env object.

internal/osutil/user_test.go
```diff
@@ -150,3 +152,40 @@ func (s *userSuite) TestNormalizeUidGid(c *check.C) {
...
+	}
+
+	envIn := map[string]string{
+		"WORKSHOP_TEST_ENV_KEY":   "VALUE",
+		"WORKSHOP_TEST_ENV_EMPTY": "",
+	}
+
+	for k, v := range envIn {
+		// Set environment
+		cmd := exec.Command("systemctl", "--user", "set-environment", k+"="+v)
```
jonathan-conder: If I understand correctly, this is modifying the real user's environment. That should be confined to spread tests or at least integration tests. Maybe here we should fake the `systemctl` command, then have a spread test to check the output format matches what we expect.

internal/workshop/workshop_dirs.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/canonical/workshop/internal/osutil"
+)
+
+func UserDataRootDir(name string) (string, error) {
+	usr, env, err := osutil.UserAndEnv(name)
+	if err != nil {
+		return "", err
+	}
+
+	path := filepath.Join(usr.HomeDir, ".local", "share")
```
jonathan-conder: Now we're going against the spec by using the real home and ignoring `$HOME`. I think we should look at `$HOME`, but also fall back to the real home if unset.

Also, about half the callers already have a `User` object. In that case we could do an error-free variant, i.e. if the environment lookup fails, treat that as "no environment," similar to "no $XDG_DATA_HOME" or "no $HOME."

Finally, slight preference to name this file `dirs.go`.

snap/local/hooks/remove
```diff
@@ -1,52 +1,20 @@
...
-# Storage pool custom volumes are shared among projects and removed last. These
-# belong to the default LXD project.
-storage_pool="workshop"
-volumes=$(run_lxc storage volume list "$storage_pool" -c nt type=custom -f compact | awk 'NR>1 {printf "%s ", $1}')
-echo "$volumes" | xargs -n1 -r | while read -r volume; do
-  run_lxc storage volume delete "$storage_pool" "$volume"
+  remove_instances "$p"
+  remove_profiles "$p"
+  remove_images "$p"
+  remove_project "$p"
```
jonathan-conder, line 16: previously `remove_project` was after everything else, is it safe to remove here?

jonathan-conder (commented):
Added my comments. The main issues I have are in the remove hook (deleting projects early) and the `user_test` which reconfigures the environment. Apologies if I'm re-litigating things you've already discussed.

internal/workshop/workshop_dirs.go
jonathan-conder: I think this one is outdated after Dmitry's changes

cmd/workshop/info_test.go
lachypjones: Originally I faked this, however it's now consistent with how other tests set temporary environments. 

As for the dependency, I hadn't realised we already pulled one in, happy to change if it seems like a better idea.

internal/interfaces/builtin/desktop_test.go
lachypjones: Nothing gets created in the directory by the desktop interface, it felt excessive to create a whole temp dir for this.

internal/interfaces/builtin/tunnel_test.go
lachypjones: I'm happy with a defer instead.

internal/osutil/user.go
lachypjones: Even though it's part of the primary package, because it has '_test', it's only exported to other tests in the same package. I think that answers your question?

cmd/workshop/info_test.go
jonathan-conder: Ok, given that other tests do this already it's fine with me.

For the second part, now I'm not so sure. I think we probably want the fallback to `User.HomeDir` on the backend, so doing the same on the client makes sense. That will require custom code (although maybe there's a function or two in `xdg` that can help).

cmd/workshop/info.go
dmitry-lyfar: @jonathan-conder agree on the fallback. As for the lib, I've been back and forth about depending on this lib (it only supports it for the current user). Also, that would become different from how server decides about XDG dirs.

internal/interfaces/builtin/desktop_test.go
jonathan-conder: Does `/tmp` need to exist?

snap/local/hooks/remove
lachypjones: Yes, storage volumes are not associated with a specific project, we remove all project-associated data here before removing the project.

internal/osutil/user.go
jonathan-conder: Ah yes I vaguely remember encountering that before. It's goes against my intuition so I have trouble remembering.

cmd/workshop/info.go
lachypjones: We had discussed this as a fallback before deciding that if $HOME is not set there are most likely larger issues and we should not continue. What would you like to do?

cmd/workshop/info.go
dmitry-lyfar: The reason it is `XDG_DATA_HOME`:

```
The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:

actions history (logs, history, recently used files, …)

current state of the application that can be reused on a restart (view, layout, open files, undo history, …)
```

We have mounts and sketch SDK files there for now and I think they are a bit different compared to the items listed. It's, ofc, subjective to an extent; given that users have their data there and it is not the application's state, I decided on `XDG_DATA_HOME` at that point.

cmd/workshop/info.go
jonathan-conder: Yeah I've changed my mind about the lib now. I think it's possible that HOME is unset within what systemd thinks the environment is (unless that is hard coded somewhere, which I wouldn't rely on)

cmd/workshop/info.go
jonathan-conder: I missed that part of the spec. Makes sense to me

docs/how-to/ros2/ros2-create-workshop.rst
lachypjones: My understanding was that 'content' (in this scenario) is a leftover relic from the content interface plug. I don't think 'content directory' is used anywhere else in the docs now, 'directory content' certainly is but I feel that's different. 

Thoughts?

internal/interfaces/builtin/desktop_test.go
lachypjones: No

jonathan-conder: seems fine then

docs/how-to/ros2/ros2-create-workshop.rst
jonathan-conder: You're right, it's just "content" used in a few other places. This change makes sense on its own.

cmd/workshop/info.go
lachypjones: No problem, fallback sounds good to me.

internal/workshop/workshop_dirs.go
```diff
@@ -2,23 +2,16 @@ package workshop
...
 
-func UserDataRootDir(name string) (string, error) {
-	usr, env, err := osutil.UserAndEnv(name)
-	if err != nil {
-		return "", err
-	}
-
-	path := filepath.Join(usr.HomeDir, ".local", "share")
+func UserDataRootDir(homedir string, env map[string]string) string {
+	path := filepath.Join(homedir, ".local", "share")
```
jonathan-conder, line 14: Just to reiterate on my outdated comment, this bit should check `$HOME` first

docs/how-to/ros2/ros2-create-workshop.rst
dmitry-lyfar: So yes, this is a mix up between the old interface name and the content itself (the case here, I believe). See line 156 where it discusses host-mapped contents of the workshop.

internal/osutil/user_test.go
lachypjones: I've removed for now. Tests to follow as a separate PR.

internal/workshop/workshop_dirs.go
lachypjones: I don't think it makes sense to do that here in the context of Dmitry's changes. I think it should be the caller's responsibility no?

jonathan-conder: the caller provides `env`, `homedir` is always the actual home dir

dmitry-lyfar: @jonathan-conder why do you want to check `$HOME` here first I'm failing to see? I suppose what you have in the `homedir` provided by the caller is more reliable. We can potentially fallback to checking the environment if empty.

I considered having only `env` as an argument here, but I think we should mandate the first argument from the caller.

internal/osutil/user_test.go
jonathan-conder: I think later on I saw something like this in one of the spread tests, might not need more

internal/workshop/workshop_dirs.go
jonathan-conder: We should respect user preferences. For example I used to override HOME when running steam so it didn't pollute my actual home dir.

jonathan-conder: Plus the XDG spec says to use `$HOME`. Falling back to the real dir is probably not compliant with the spec, but I think it's worth doing to avoid having unnecessary error paths.

jonathan-conder: Probably in future we should just look at what GLib does :sweat_smile: https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gutils.c#L1864
I think it's the same as what I'm suggesting

snap/local/hooks/remove
dmitry-lyfar: A possible reason for the project to be deleted after is this: https://github.com/canonical/lxd/issues/15179. This hook does not seem to suffer from the issue exactly because it shares volumes with the default project (I tested locally).

dmitry-lyfar (approved):

PR #302, opened 2025-02-17
Doc: Factor out a sketching how-to

akcano:


docs/how-to/build-sdks/sketch-sdk.rst
```diff
@@ -0,0 +1,185 @@
+.. _how_sketch:
+
+How to sketch an SDK with |ws_markup|
```
dmitry-lyfar: Given that a user may arrive here not only from the tutorial page (where sketch is introduced nicely):

I think the title and the intro should reflect the purpose of a sketch-sdk in the how-to guide. 

The purpose of sketch SDK is to let someone to experiment with their environment and preserve such modifications after they have created some core workshop (e.g. ROS 2, go, or Vitis AI environment).

One can install extra tools into the environment: linters, static code analysers, apt packages. Some may require running UI apps or use Git from within a workshop, the kind of functionality not provided by a framework or vendor SDK.

Can you rethink the structure and the guide to be talking in these terms? (Counterintuitively, `Craft an SDK` feels still relevant for a how-to guide, though).

Note, that you don't have to use `--no-install-recommends` in the examples starting from the next release, it will be preconfigured to do so (just a recent change made after this PR was pushed).

docs/how-to/build-sdks/sketch-sdk.rst
```diff
@@ -0,0 +1,256 @@
...
+Start sketching
+---------------
+
+Instead of manually installing tools using
+:command:`workshop shell` or :command:`workshop exec`,
+you can create a local SDK that automates these tasks with |ws_markup|.
+
+Running :command:`workshop sketch-sdk`
+opens a simplified version of an :ref:`SDK definition <exp_sdk_definition>`.
+This defines all SDK components in a single file named :file:`sketch.yaml`:
```
dmitry-lyfar:
```suggestion
This defines all SDK components in a single file named :file:`sdk.yaml`:
```

docs/how-to/build-sdks/sketch-sdk.rst
```diff
@@ -0,0 +1,256 @@
...
+
+
+.. note::
+
+   With |ws_markup|, you don't need to specify non-interactive flags like
+   :option:`!-y` or :option:`!--no-install-recommends` with :program:`apt-get`;
+   the environment handles this automatically.
+
+
+Once you save and exit :file:`sketch.yaml`,
```
dmitry-lyfar:
```suggestion
Once you save and exit :file:`sdk.yaml`,
```
This is called `sdk.yaml` to then be easily installed as regular local SDK. It could be a good idea to rename it to sketch.yaml when we introduce `--eject`

docs/how-to/build-sdks/sketch-sdk.rst
```diff
@@ -0,0 +1,256 @@
...
+.. code-block:: console
+
+   sketch:
+     tracking:   ~/.local/share/workshop/project/b5b0f128/sdk/sketch/dev
+     installed:  2025-02-24  (x1)
+
+
+The sketch SDK entry displays the last update time and its revision (:samp:`x1`).
+Instead of a channel, :samp:`tracking` lists the SDK definition path on the host,
+distinguishing the sketch SDK from other SDKs.
```
dmitry-lyfar: Just to note that this explanation may become obsolete relatively quickly. A path in `tracking` indicates that `sketch` SDK is an SDK that comes from a local source. That would be the same for other local SDKs that could be installed via (not yet existing) `workshop try`. So, sketch SDK is a regular-type SDK that comes from a local source. There'll be others like that which will not make `sketch` a special case in that regard.

dmitry-lyfar (approved):
I think this is much nicer.

Still not entirely convinced by the title and the place of this how-to in the navigational panel.

- The title suggests iterating on an SDK whilst to me it's about fine-tuning or extending your workshop where `sketch` SDK is a means.
- It is now under `Build an SDK`. It's logically correct (yes, we build an SDK here) but it is rather about `Use workshops` IMO.

Lmk know what you think.

akcano:
Here's my take at your suggestion, looks nice.

dmitry-lyfar (approved):

PR #263, opened 2025-01-16
Make sketch-sdk abort current refresh in progress if needed

jonathan-conder:
# Description

Previously this was a common workflow:
```console
$ workshop sketch-sdk
# add some buggy code to setup-base
2025-01-16T13:14:37+13:00 ERROR command exit code 1
error: cannot refresh; fix the errors reported,
then run "workshop refresh --continue ws".
To abort and revert, run "workshop refresh --abort ws"
$ workshop sketch-sdk
# fix bugs
error: cannot refresh "ws": waiting on error
$ workshop refresh --abort
"ws" refresh aborted
$ workshop sketch-sdk
# have to add some whitespace or something to trigger a refresh now
```

The `waiting on error` message should not appear anymore, `sketch-sdk` will abort automatically. The second issue (having to edit the file in some way to trigger a refresh) is still present, but should be much less common so I don't think it's really a problem anymore. Users can just call `workshop refresh ws/sketch` anyway.

I'm a little bit unhappy with the complexity of the test I added, but this is quite a long sequence of tasks so I don't know that there's a way to simplify it much further.

I think it was a good idea to handle this all client-side, e.g. the user will be able to see that a previous refresh was aborted. In order to enable that I needed a way to expose error types to the client, so I opted to replace `fmt.Sprintf()` with `fmt.Errorf().Error()` in the error responders. `Errorf()` will wrap errors formatted with `%w`, so we can inspect them using `errors.Is`. If this gets too complicated, we can switch to `errors.As` with a custom interface.

# Self-review quick check

 * [x] Make decisions that cost a lot to reverse explicit in the PR description.
 * [x] Avoid nested conditions.
 * [x] Delete dead code and redundant comments.
 * [x] Normalise symmetries by sticking to doing identical things identically. 
 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 // one way to handle multiple returns
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```
 * [x] Check that coupled code elements, files, and directories are adjacent. For example, test data is stored as close as possible to a test.
 * [x] Put variable declaration and initialisation together.
 * [x] Divide large expressions into digestable and self-explanatory ones. Use multiple variables if required.
 * [x] Put a blank line between two logically different chunks of code.
 * [x] Follow the [style guide](https://github.com/canonical/workshop/tree/main/docs/contributing.rst#error-messages) for new error messages.

## Docs

* [ ] I have checked and added or updated relevant documentation.
[...]

PR #307, opened 2025-02-20
Make sketch-sdk abort refresh in progress before status check

jonathan-conder:
# Description

Fixes a regression in the behaviour introduced by https://github.com/canonical/workshop/pull/263. The status check in https://github.com/canonical/workshop/pull/289 happens before `sketch` has a chance to abort the refresh.

It slipped through because I put the wrong status in one of the mocked responses in the unit test. That's fixed now, but I also added an end-to-end test to prevent further regressions.

Previously, the abort is triggered only after attempting a refresh. I've added a new `Waiting` status so that sketch can detect early on if another sketch (or `refresh --wait-on-error`) is paused, without relying on the `Notes` field. The docs now consistently refer to this state as `Waiting` or "paused" (and not "stopped").

I'm assuming if the abort succeeds, the workshop will be `Ready` again (to avoid calling `info` again).

I removed the special error handling for `wait-on-error`, but kept the more general error handling changes from https://github.com/canonical/workshop/pull/263

A separate but related change is that `exec` no longer sets the workshop status to `Pending`. This allows some operations which weren't previously possible during `workshop shell`. I've added a warning when connecting/disconnecting interfaces during a shell session (because the environment won't be updated).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

* [] I confirm the PR has no implications for documentation.

### DIFF
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -23,13 +23,13 @@ func (c *CmdRemove) Command() *cobra.Command {
 		Long: `
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off' or 'Pending'
+- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'
 - Stops the workshop if it's not already 'Stopped'
 - Deletes the workshop but preserves its definition
 
 Notes:
 
-- If any listed workshop is 'Off' or 'Pending', none are removed.
+- If any listed workshop is 'Off', 'Pending' or 'Waiting', none are removed.
 
 - To rebuild a removed workshop from scratch, use 'workshop launch'.
 
--- a/docs/reference/cli/workshop-remove.rst
+++ b/docs/reference/cli/workshop-remove.rst
@@ -18,13 +18,13 @@ Remove one or many workshops.
 
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off' or 'Pending'
+- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'.
 - Stops the workshop if it's not already 'Stopped'
 - Deletes the workshop but preserves its definition
 
 Notes:
 
-- If any listed workshop is 'Off' or 'Pending', none are removed.
+- If any listed workshop is 'Off', 'Pending' or 'Waiting', none are removed.
 
 - To rebuild a removed workshop from scratch, use 'workshop launch'.
 
--- a/docs/reference/workshop-status.rst
+++ b/docs/reference/workshop-status.rst
@@ -28,7 +28,7 @@ there is no container yet.
    stateDiagram-v2
        OFF --> READY: launch
        OFF --> ERROR: launch (on error)
-       OFF --> PENDING: launch --wait-on-error (on error)
+       OFF --> WAITING: launch --wait-on-error (on error)
 
 
 Ready
@@ -49,7 +49,7 @@ up and ready to do some work.
        READY --> READY: remount
        READY --> READY: refresh
        READY --> ERROR: refresh (on error)
-       READY --> PENDING: refresh --wait-on-error (on error)
+       READY --> WAITING: refresh --wait-on-error (on error)
 
 
 Stopped
@@ -68,23 +68,32 @@ but is still linked to the project directory.
        STOPPED --> STOPPED: remount
 
 
-Pending
+Waiting
 -------
 
-The workshop is being updated or changing its status;
-only a few commands will be accepted,
-and the container itself is non-operational.
+The workshop is paused
+in the middle of a change
+to allow for interactive debugging;
+only a few commands will be accepted.
 
 .. mermaid::
-   :alt: Pending state
-   :caption: Pending state
+   :alt: Waiting state
+   :caption: Waiting state
    :align: center
 
     stateDiagram-v2
-        PENDING --> OFF: launch --abort
-        PENDING --> READY: launch --continue
-        PENDING --> READY: refresh --abort
-        PENDING --> READY: refresh --continue
+        WAITING --> OFF: launch --abort
+        WAITING --> READY: launch --continue
+        WAITING --> READY: refresh --abort
+        WAITING --> READY: refresh --continue
+
+
+Pending
+-------
+
+An intermediate state
+while the workshop is being updated or changing its status;
+only a few commands will be accepted.
 
 
 Error
--- a/internal/overlord/conflict/conflict.go
+++ b/internal/overlord/conflict/conflict.go
@@ -3,6 +3,7 @@ package conflict
 import (
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/canonical/workshop/internal/overlord/state"
 	"github.com/canonical/workshop/internal/sdk"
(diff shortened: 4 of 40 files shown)


lachypjones (commented):
This reintroduces an error #289 fixed. 
If a workshop is pending due to --wait-on-error from a full refresh, sketch will abort that refresh.

jonathan-conder:
> If a workshop is pending due to --wait-on-error from a full refresh, sketch will abort that refresh.

Is there a way to distinguish that from a failed sketch?

internal/overlord/workshopstate/manager.go
```diff
@@ -68,18 +67,20 @@ func (w *WorkshopManager) CheckStatus(ctx context.Context, name, pId string, all
...
+			switch health.Code {
+			case "wait-on-error", "wait-on-sketch-error":
+				return errors.New("waiting on error")
+			default:
+				return errors.New("other changes in progress")
 			}
-			return fmt.Errorf("other changes in progress")
 		case healthstate.ErrorStatus:
-			return fmt.Errorf("workshop is unhealthy")
+			return errors.New("workshop is unhealthy")
```
akcano: I'd say the use of 'is' is rather inconsistent:
- is unhealthy
- is unknown

but:

- in progress
- not running

WDYT?

jonathan-conder: yeah let's remove the `is`s

cmd/workshop/remove.go
```diff
@@ -23,13 +23,13 @@ func (c *CmdRemove) Command() *cobra.Command {
 		Long: `
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off' or 'Pending'
+- Checks that the workshop isn't 'Off', 'Pending', or 'Waiting'
```
akcano:
```suggestion
- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'
```

cmd/workshop/remove.go
```diff
@@ -23,13 +23,13 @@ func (c *CmdRemove) Command() *cobra.Command {
...
 - Stops the workshop if it's not already 'Stopped'
 - Deletes the workshop but preserves its definition
 
 Notes:
 
-- If any listed workshop is 'Off' or 'Pending', none are removed.
+- If any listed workshop is 'Off', 'Pending', or 'Waiting', none are removed.
```
akcano:
```suggestion
- If any listed workshop is 'Off', 'Pending' or 'Waiting', none are removed.
```

akcano: We don't use the Oxford comma for historical reasons; this is subject to change in the nearest future as we migrate to en-US by the end of the cycle.

internal/overlord/conflict/conflict.go
```diff
@@ -107,14 +107,26 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
...
 func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreChange string) error {
 	for _, task := range st.Tasks() {
 		chg := task.Change()
 		if chg.IsReady() || chg.ID() == ignoreChange {
 			continue
 		}
 
+		var incidental bool
+		if err := chg.Get("incidental", &incidental); err != nil {
+			var noState *state.NoStateError
```
dmitry-lyfar: it's errors.Is in all other different NoStateError cases; that's not entirely correct usage of As / Is by the state package, but it's consistent.

jonathan-conder:
Let me know what you guys think of 4977252. If everyone's happy I'll go ahead and clean up the history

internal/overlord/conflict/conflict.go
```diff
@@ -108,40 +109,41 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
 
 // Iterates over the list of running tasks and returns a ChangeConflictError if
 // there is another major change running for the provided projectID / workshop pair.
-func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreChange string) error {
+// Changes can be filtered by ID or kind for fine-grained processing.
```
dmitry-lyfar: I'm not entirely sure what the returned list of filtered changes means, could you update the function's comment to reflect the changed returned value? It describes only the error for now.

Does the `filterChange` and `filterKinds` mean the id and kinds of changes that the conflict checker should be consider compatible to co-exist?

Also, do we use that `filterChange` (previously, `ignoreChange`) anywhere?

dmitry-lyfar:
> Let me know what you guys think of [4977252](https://github.com/canonical/workshop/commit/497725218d67fdc7650a30b4718e083aee97be54). If everyone's happy I'll go ahead and clean up the history

Left my comments, I think it's ok to squash it all.

internal/overlord/conflict/conflict.go
jonathan-conder: > I'm not entirely sure what the returned list of filtered changes means, could you update the function's comment to reflect the changed returned value? It describes only the error for now.

Yeah I can try to make that more clear.

> Does the `filterChange` and `filterKinds` mean the id and kinds of changes that the conflict checker should be consider compatible to co-exist?

Currently `ignoreChange` isn't used anywhere, so I'm not sure what the intent is. For `filterKinds` the idea is that some changes may or may not be compatible, so instead of deciding that here we return those changes and leave the rest up to the caller. Maybe it would be better to just return the list of running changes at let the caller decide whether to return an error.

> Also, do we use that `filterChange` (previously, `ignoreChange`) anywhere?

I don't think so

dmitry-lyfar: Perhaps, it can just accept kinds that are compatible as the purpose is to raise an error if there are incompatible changes running?

dmitry-lyfar: To duplicate my comment about `ignore`, we should remove it; sorry, was sending it via email response.

jonathan-conder: > Perhaps, it can just accept kinds that are compatible as the purpose is to raise an error if there are incompatible changes running?

That basically works. It needs to return something different if there's nothing pending vs only compatible changes pending (so that connect can know whether to raise a warning). Could just be a bool though

jonathan-conder: > To duplicate my comment about `ignore`, we should remove it; sorry, was sending it via email response.

I'll do that tomorrow

jonathan-conder:
I redid the exec-ignoring logic and cleaned up the history a bit. And rebased on main.

jonathan-conder:
Added a few more tests and suppressed `errcheck` complaints about `defer d.Overlord.Stop()` in the tests.

internal/overlord/ifacestate/request.go
```diff
@@ -19,17 +20,20 @@ func (e ErrAlreadyConnected) Error() string {
...
+	if err := healthstate.CheckWorkshopHealth(st, slotW, allowed); err != nil {
+		return nil, fmt.Errorf("cannot connect %q: %w", connRef.SlotRef.ShortRef(), err)
 	}
 
-	slotProject, slotWorkshop := connRef.SlotRef.ProjectId, connRef.SlotRef.Workshop
-	err = conflict.CheckChangeConflict(st, slotProject, slotWorkshop, "")
-	if err != nil {
-		return nil, err
+	if _, err := conflict.FindChangeByKind(st, plugW.Project.ProjectId, plugW.Name, "exec"); err == nil {
+		st.Warnf("some connections may not affect existing shell sessions in %q", plugW.Name)
```
dmitry-lyfar: Can we suggest something actionable, maybe: `active shell sessions exist in %q, you may need to restart them for connections to take effect`?

dmitry-lyfar (approved):

docs/reference/cli/workshop-remove.rst
```diff
@@ -18,13 +18,13 @@ Remove one or many workshops.
 
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off' or 'Pending'
+- Checks that the workshop isn't 'Off', 'Pending', or 'Waiting'.
```
akcano:
```suggestion
- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'.
```

no Oxford comma yet

docs/reference/cli/workshop-remove.rst
```diff
@@ -18,13 +18,13 @@ Remove one or many workshops.
...
 - Stops the workshop if it's not already 'Stopped'
 - Deletes the workshop but preserves its definition
 
 Notes:
 
-- If any listed workshop is 'Off' or 'Pending', none are removed.
+- If any listed workshop is 'Off', 'Pending', or 'Waiting', none are removed.
```
akcano:
```suggestion
- If any listed workshop is 'Off', 'Pending' or 'Waiting', none are removed.
```

docs/reference/workshop-status.rst
```diff
@@ -68,23 +68,32 @@ but is still linked to the project directory.
...
-        PENDING --> READY: launch --continue
-        PENDING --> READY: refresh --abort
-        PENDING --> READY: refresh --continue
+        WAITING --> OFF: launch --abort
+        WAITING --> READY: launch --continue
+        WAITING --> READY: refresh --abort
+        WAITING --> READY: refresh --continue
+
+
+Pending
```
akcano, line 91: I get it that 'Pending' doesn't warrant a diagram now; fine by me.

akcano (approved):
LGTM, minor notes.

internal/overlord/ifacestate/request.go
jonathan-conder: I'd hesitate to say `active shell sessions exist`, it could just be a long-running script for example. How about `active processes in %q may need restarting for new connections to take effect`

jonathan-conder: or `need to be restarted` or `need a restart`

docs/reference/workshop-status.rst
jonathan-conder: yeah, pending is represented by the arrows in most cases

dmitry-lyfar (approved):

PR #309, opened 2025-02-21
Doc: Add interfaces to workshop definition reference

jonathan-conder:
# Description

Most interfaces, especially `port-forward` (coming soon), behave differently in the system SDK. I think this behaviour should be documented alongside the workshop definition rather than the SDK definition. This should also make it easier to remember to update the workshop definition docs when adding a new interface.

There will probably be a fair bit of repetition between these two references, but I think the table-based format is a nice alternative to the example-based format in the SDK definition docs, and it doesn't hurt to have both. If it gets out of hand we could document the system-specific attributes in the workshop definition, and refer to the SDK definition for other attributes.

The docs are much easier to read when interfaces behave consistently, so I changed how the mount interface works.

The system SDK will now only have one mount slot, named "mount". This is the same as most other interfaces, and reinforces the idea that the system SDK is for system resources only (no workshop-source on a system SDK slot).

There's no need to prevent installation of slots with a host-source attribute at the policy level, as this is handled by the slot sanitizer, in line with other interfaces. This attribute is never attached to the slot itself, only to a ConnectedSlot.

I think the previous auto-connect policy would allow auto-connecting any slot named "mount," which doesn't make sense to me because this slot might not have the special host-source logic that the system SDK does. So auto-connect is now based on the slot SDK type (and explicit connections, as before).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/reference/definitions/workshop-definition.rst
+++ b/docs/reference/definitions/workshop-definition.rst
@@ -106,7 +106,8 @@ Each SDK is described with the following keys:
    * - :samp:`name` (required)
      - string
      - Name of an existing SDK
-       that is available from the SDK store.
+       that is available from the SDK store,
+       or :samp:`system`.
 
    * - :samp:`channel` (required)
      - string
@@ -120,6 +121,8 @@ Each SDK is described with the following keys:
        `snap-like format <https://snapcraft.io/docs/channels>`__
        of :samp:`<TRACK>/<RISK>`
        without the :samp:`<BRANCH>` part.
+
+       Not required for the :ref:`system SDK <ref_system_sdk>`.
 
    * - :samp:`plugs`
      - object
@@ -140,13 +143,31 @@ Each SDK is described with the following keys:
        and the relevant attributes (described below).
 
 
+.. _ref_system_sdk:
+
+System SDK
+~~~~~~~~~~
+
+.. @artefact system SDK
+
+The system SDK is built in to every workshop,
+and isn't available in the SDK store,
+so it doesn't require a :samp:`channel`.
+
+The system SDK declares slots for many interfaces.
+These represent resources provided by the host system.
+The workshop definition can define additional plugs and slots
+for all SDKs,
+including the system SDK.
+
+
 Camera interface
 ~~~~~~~~~~~~~~~~
 
 .. @artefact camera interface
 
 Camera interface plugs must be named :samp:`camera`
-and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+and can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They have no attributes.
 
 The only camera interface slot is :samp:`system:camera`.
@@ -158,7 +179,7 @@ Desktop interface
 .. @artefact desktop interface
 
 Desktop interface plugs must be named :samp:`desktop`
-and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+and can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They have no attributes.
 
 The only desktop interface slot is :samp:`system:desktop`.
@@ -170,7 +191,7 @@ GPU interface
 .. @artefact GPU interface
 
 GPU interface plugs must be named :samp:`gpu`
-and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+and can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They have no attributes.
 
 The only GPU interface slot is :samp:`system:gpu`.
@@ -181,7 +202,7 @@ Mount interface
 
 .. @artefact mount interface
 
-Mount interface plugs can't belong to the :ref:`system SDK <exp_system_sdk>`.
+Mount interface plugs can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They are described by the following attributes:
 
 .. @artefact mount interface attributes
@@ -205,7 +226,7 @@ They are described by the following attributes:
      - Whether the target directory should be read-only.
 
 
-The only mount interface slot in the :ref:`system SDK <exp_system_sdk>` is :samp:`system:mount`.
+The only mount interface slot in the :ref:`system SDK <ref_system_sdk>` is :samp:`system:mount`.
 It has a single dynamic attribute named :samp:`host-source`,
 which can be only configured at :ref:`remount <ref_workshop_remount>`.
 
@@ -238,7 +259,7 @@ SSH interface
 .. @artefact SSH interface
 
 SSH interface plugs must be named :samp:`ssh-agent`
-and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+and can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They have no attributes.
 
 The only SSH interface slot is :samp:`system:ssh-agent`.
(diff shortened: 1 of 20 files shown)


docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+
+.. @artefact camera interface
+
+Camera interface plugs must be named :samp:`camera`
+and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+They have no attributes.
+
+The only camera interface slot is :samp:`system:camera`.
+
+Desktop interface
```
akcano, line 176:
```suggestion


Desktop interface
```

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+
+.. @artefact desktop interface
+
+Desktop interface plugs must be named :samp:`desktop`
+and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+They have no attributes.
+
+The only desktop interface slot is :samp:`system:desktop`.
+
+GPU interface
```
akcano, line 188:
```suggestion


GPU interface
```

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+~~~~~~~~~~~~~
+
+.. @artefact GPU interface
+
+GPU interface plugs must be named :samp:`gpu`
+and can't belong to the :ref:`system SDK <exp_system_sdk>`.
+They have no attributes.
+
+The only GPU interface slot is :samp:`system:gpu`.
+
```
akcano, line 198:
```suggestion
The only GPU interface slot is :samp:`system:gpu`.


```

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+     - boolean
+     - Whether the target directory should be read-only.
 
-       The only interface with additional attributes is :samp:`mount`;
-       it requires the :samp:`workshop-source` property
-       to specify a path inside the workshop
-       for the slot's source directory;
+
+The only mount interface slot in the :ref:`system SDK <exp_system_sdk>` is :samp:`system:mount`.
+It has a dynamic attribute :samp:`host-source`,
```
akcano:
```suggestion
It has a single dynamic attribute named :samp:`host-source`,
```

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+which can be only configured using :ref:`remount <ref_workshop_remount>`.
```
akcano:
```suggestion
which can be only configured at :ref:`remount <ref_workshop_remount>`.
```

Commands are generally best referenced by their full names (e.g. `workshop remount`); here, we can nounify (sorry) it.

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+   * - :samp:`workshop-source` (required)
+     - string
+     - A path inside the workshop
+       to be used as the slot's source directory;
        :file:`/project` or :envvar:`$SDK`-based paths can be used;
        :envvar:`$SDK` expands into the SDK's installation path in the workshop.
 
-.. @artefact $SDK
+
+SSH interface
```
akcano:
```suggestion


SSH interface
```

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+       and the relevant attributes (described below).
+
+
+Camera interface
+~~~~~~~~~~~~~~~~
+
+.. @artefact camera interface
+
+Camera interface plugs must be named :samp:`camera`
+and can't belong to the :ref:`system SDK <exp_system_sdk>`.
```
akcano: Technically, the existing linking policy is to stay within the boundaries of a subgenre; inline links in reference docs should only point to other reference docs etc. The exception is the tutorial; other exceptions are possible if no relevant section exists, which is the case here. OTOH, this probably warrants such a section, so I'll make a note.

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +131,115 @@ Each SDK is described with the following keys:
...
+     - Value
+     - Description
+
+   * - :samp:`workshop-target` (required)
+     - string
+     - A path inside the workshop
+       to be used as the plug's target directory.
+
+   * - :samp:`read-only`
+     - boolean
```
akcano:
```suggestion
     - Boolean
```

Fun fact: I lived for a couple of years in the same borough where George Boole lived and died (Ballintemple, Cork).

akcano (changes requested):
Awesome work, a few technical fixes.

docs/reference/definitions/workshop-definition.rst
jonathan-conder: Haha no way! My parents visited Cork last year and stayed not far from there.

I had thought lowercase adjectives was a convention in maths but seems like it's just French influence: https://math.stackexchange.com/questions/4526/mathematical-concepts-named-after-mathematicians-that-have-become-acceptable-to

docs/reference/definitions/workshop-definition.rst
jonathan-conder: This has gone up to 3 lines of space now. I'll revert it but let me know if the 3 lines was intentional.

docs/reference/definitions/workshop-definition.rst
jonathan-conder: I think the section should just go in this same file, the workshop definition is where people are most likely to encounter the system SDK. I'll make an attempt but you can do something different if you prefer.

docs/reference/definitions/workshop-definition.rst
akcano: No, 2 lines max.

docs/reference/definitions/workshop-definition.rst
```diff
@@ -132,26 +134,136 @@ Each SDK is described with the following keys:
...
+SSH interface
+~~~~~~~~~~~~~
+
+.. @artefact SSH interface
+
+SSH interface plugs must be named :samp:`ssh-agent`
+and can't belong to the :ref:`system SDK <ref_system_sdk>`.
+They have no attributes.
+
+The only SSH interface slot is :samp:`system:gpu`.
```
dmitry-lyfar:
```suggestion
The only SSH interface slot is :samp:`system:ssh-agent`.
```

dmitry-lyfar (approved):
Minor typo, otherwise looks great. Please fix and merge.

docs/reference/definitions/workshop-definition.rst
jonathan-conder: nice catch!

dmitry-lyfar (approved):

PR #312, opened 2025-02-22
Pre-download LXD base images for spread runs

dmitry-lyfar:
# Description

This helps with (mainly) local spread runs that reuse existing VMs. The images must populated into a directory on the host and mounted to all the spread VMs via modifying their profiles:

```
lxc profile device add default mnt-image disk source=$image_dir path=/mnt
```

The `prepare` script for the test suites that uses LXD, would import the images before the run instead of letting every spread instance to download it from slow LXD servers multiple times.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/.github/workflows/spread.yaml
+++ b/.github/workflows/spread.yaml
@@ -27,11 +27,25 @@ jobs:
       - name: Install LXD
         uses: canonical/setup-lxd@main
         with:
-          channel: 5.21/stable
+          channel: 6/stable
       - name: Cleanup job workspace
         run: |
           rm -rf "${{ github.workspace }}"
           mkdir "${{ github.workspace }}"
+      - name: Pre-download LXD images
+        run: |
+          image_dir="$HOME/images"
+          mkdir -p "$image_dir"
+
+          for version in 22.04 24.04; do
+            if [ ! -f "$image_dir/ubuntu-$version.tar.gz" ]; then
+              lxc image copy "ubuntu:$version/amd64" local: --alias "$version/amd64"
+              lxc image export "$version/amd64" "$image_dir/ubuntu-$version.tar.gz"
+            fi
+          done
+
+          lxc project switch default
+          lxc profile device add default mnt-image disk source=$image_dir path=/mnt
       - name: Checkout Workshop
         uses: actions/checkout@v4
         with:
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -246,6 +246,17 @@ of ``Spread``:
 
    spread tests/<TestPathName>
 
+When running locally, you can accelerate the test runs by reusing instances 
+and local LXD base images. For more examples, see the ``spread`` GitHub workflow.
+
+.. code-block:: console
+
+   image_dir=$HOME/images
+   lxc image export <fingerprint> "$image_dir/ubuntu-22.04.tar.gz"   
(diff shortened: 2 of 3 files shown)


.github/workflows/spread.yaml
```diff
@@ -27,11 +27,25 @@ jobs:
...
+          mkdir -p "$image_dir"
+
+          for version in 22.04 24.04; do
+            if [ ! -f "$image_dir/ubuntu-$version.tar.gz" ]; then
+              lxc image copy "ubuntu:$version/amd64" local: --alias "$version/amd64"
+              lxc image export "$version/amd64" "$image_dir/ubuntu-$version.tar.gz"
+            fi
+          done
+
+          lxc project switch default
```
jonathan-conder, line 47: It would be nice to have a separate project for spread. But I think that will have to live in the spread fork and not here.

tests/lib/utils.sh
```diff
@@ -10,6 +10,21 @@ function setup_lxd() {
...
+    # This assumes images are mounted into the spread instance at /mnt.
+    image_dir="/mnt"
+    versions=("20.04" "22.04" "24.04")
+
+    for version in "${versions[@]}"; do
+        image_file="$image_dir/ubuntu-$version.tar.gz"
+        image_root="$image_dir/ubuntu-$version.tar.gz.root"
+        image_alias="workshop-ubuntu@$version-amd64"
+
+        if [ -f "$image_file" ] && ! lxc image list | grep -q "$image_alias"; then
```
jonathan-conder:
```suggestion
        if [ -f "$image_file" ] && ! lxc image info "$image_alias" &>/dev/null; then
```

tests/lib/utils.sh
```diff
@@ -23,9 +38,19 @@ function prepare_environment() {
...
     snap install --classic --channel=1.23/stable go
     snap install yq
+
     # The unattended upgrades hold locks on reusable instances
     # and can break a spread run.
-    apt-get remove -y unattended-upgrades
+    systemctl stop unattended-upgrades.service || true
+    systemctl disable unattended-upgrades.service || true
+    systemctl disable apt-daily.timer || true
+    systemctl disable apt-daily-upgrade.timer || true
```
jonathan-conder, line 50: is this required for the apt test? I think users are likely to have these services enabled, so we should keep them enabled for the tests too

jonathan-conder (approved):

tests/lib/utils.sh
jonathan-conder: For future reference, this is to prevent the prepare script from failing (e.g. when reusing an existing spread instance). Since workshops don't currently interact with apt/dpkg on the host, it shouldn't have implications for the tests.

docs/contributing.rst
```diff
@@ -246,6 +246,17 @@ of ``Spread``:
 
    spread tests/<TestPathName>
 
+When running locally, you can accelerate the test runs by reusing instances 
+and local LXD base images. For more examples, see the `spread` GitHub workflow.
```
```suggestion
and local LXD base images. For more examples, see the ``spread`` GitHub workflow.
```

akcano (changes requested):

PR #313, opened 2025-02-24
Introduce a resolver for workshop refresh

dmitry-lyfar:
# Description

`workshop refresh` will soon support partial refreshes by recovering from a snapshot containing installation side-effects for previously installed SDKs that have not received updates.

This change introduces a two-stage process for the refresh operation. First, it compares the previous workshop definition with the new one to identify differences and creates a refresh plan. Then, it uses this plan to generate the necessary tasks.

There will be no changes to the existing behavior of `refresh`—except when the workshop definition remains unchanged and no new revisions are available for its SDKs. In such cases, the refresh operation will be a no-op.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


akcano (approved):

internal/daemon/api_workshops_test.go
```diff
@@ -1211,7 +1268,84 @@ func (s *apiSuite) TestWorkshopConnectionsPlugIsBoundTo(c *check.C) {
...
+	slots       []string
+}
+
+func (s *apiSuite) ensureWorskhops(c *check.C, want []expectedWorkshop) {
+	got, err := s.b.ProjectWorkshops(s.ctx)
+	c.Assert(err, check.IsNil)
+	c.Assert(got, check.HasLen, len(want), check.Commentf("expected %d workshops, got %d", len(want), len(got)))
+
+	for idx, w := range got {
+		wantws := want[idx]
```
jonathan-conder, line 1286: how do we know the lists are in the same order?

internal/daemon/api_workshops_test.go
```diff
@@ -1211,7 +1268,84 @@ func (s *apiSuite) TestWorkshopConnectionsPlugIsBoundTo(c *check.C) {
...
+		for _, conn := range wantws.connections {
+			connref, err := interfaces.ParseConnRef(conn)
+			c.Assert(err, check.IsNil)
+
+			_, err = repo.Connection(connref)
+			c.Assert(err, check.IsNil)
+		}
+
+		for _, plug := range wantws.plugs {
+			ref := strings.SplitN(plug, ":", -1)
```
jonathan-conder, line 1305:
```suggestion
			ref := strings.Split(plug, ":")
```
same below, and maybe elsewhere - haven't checked yet

internal/daemon/api_workshops_test.go
```diff
@@ -1249,283 +1383,542 @@ func (s *apiSuite) TestRefreshWorkshopSuccess(c *check.C) {
...
+
+	s.ensureWorskhops(c, want)
 }
 
-func (s *apiSuite) TestRefreshWorkshopRestoresPreviousWorkshopIfFailed(c *check.C) {
+func (s *apiSuite) TestRefreshRemoveSdk(c *check.C) {
 	s.daemon(c)
 	s.d.Overlord().Loop()
-	defer s.d.Overlord().Stop()
+	defer func() { c.Check(s.d.Overlord().Stop(), check.IsNil) }()
```
jonathan-conder: We should be consistent about this

internal/overlord/workshopstate/request.go
```diff
@@ -228,31 +192,38 @@ func constructWorkshop(st *state.State, file *workshop.File, project workshop.Pr
...
+	full := slices.Clone(sdks)
+	full = slices.Insert(full, 0, sdk.Setup{Name: sdk.System.String(), Revision: sdk.Revision{N: -1}})
 
-	// run a quick check health script (for every SDK, if present)
-	checkHealth := checkHealthHooks(st, file)
-	checkHealth.WaitAll(launch)
-	launch.AddAll(checkHealth)
+	connect := autoconnectSdks(st, full)
+	connect.WaitAll(install)
+	connect.WaitAll(system)
```
jonathan-conder, line 214: install already depends on system, so this line isn't needed. But no harm in keeping it I guess

internal/overlord/workshopstate/request.go
```diff
@@ -262,60 +233,392 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
...
+	sketchFound, err := maybeSketch(ctx, w.Project.ProjectId, w.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	if w.Base != file.Base {
+		plan.refreshAnyway = true
+	}
+
+	if !slices.Equal(w.File.Connections, file.Connections) {
```
jonathan-conder, line 451: this should be insensitive to sorting, right?

internal/overlord/workshopstate/request.go
```diff
@@ -262,60 +233,392 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
...
+			if _, exist := w.Sdks[info.Name]; exist {
+				plan.refresh = append(plan.refresh, sdk.Setup{Name: info.Name, Channel: info.Channel, Revision: info.Revision})
+			}
+		}
+
+		if sketchFound {
+			if installed, exist := w.Sdks[sdk.Sketch]; exist {
+				plan.refresh = append(plan.refresh, installed)
+			} else {
+				plan.refresh = append(plan.refresh, sdk.Setup{Name: sdk.Sketch, Revision: sdk.Revision{N: -1}})
```
jonathan-conder: should this be plan.install?

internal/overlord/workshopstate/request.go
```diff
@@ -262,60 +233,392 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
...
+
+	if plan.RecreateWorkshop() {
+		conn := []sdk.Setup{{Name: sdk.System.String(), Revision: sdk.Revision{N: -1}}}
+		conn = append(conn, plan.InstallOrRefresh()...)
+		connect := autoconnectSdks(st, conn)
+		addTaskSet(connect)
+	} else {
+		connect := autoconnectSdks(st, plan.InstallOrRefresh())
+		addTaskSet(connect)
+	}
```
jonathan-conder, line 564: these branches share most of the logic and could be simplified, but I guess the second branch will soon be different so might not be worth it.

internal/overlord/workshopstate/request.go
```diff
@@ -262,60 +233,392 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
...
+		removeStash.JoinLane(cleanupLane)
+		removeStash.WaitFor(lastRefreshTsk)
+
+		if refresh.MaybeEdge(EdgeRefreshCleanup) == nil {
+			refresh.MarkEdge(removeStash, EdgeRefreshCleanup)
+		}
+
+		refresh.AddTask(removeStash)
+
+	}
```
```suggestion
	}
```

jonathan-conder (commented):
Looks good overall. My biggest concern is about comparing lists that might not be sorted (maybe they get sorted at some point that I'm not aware of). A few minor fixes as well

internal/overlord/workshopstate/request.go
dmitry-lyfar: `install` can be empty and that will make `connect` unintentionally concurrent.

internal/daemon/api_workshops_test.go
dmitry-lyfar: as discussed: we  do not intend to check the installation order here. It is a subject to a separate check / set of cases. This has to check the desired state after running `refresh` commands.

internal/overlord/workshopstate/request.go
dmitry-lyfar: as discussed: sorting appears to be important to preserve (e.g. subsequent submounts may rely on that, although, unlikely. in general, more strict rules and simplicity in this logic will be prioritised for now).

internal/overlord/workshopstate/request.go
dmitry-lyfar: True. Interesting that installation of a `sketch` SDK is not going to reach this as it is performed as part of the partial refresh. I will move it to the install list, as refresh will support partial refresh anyway.

jonathan-conder (approved):

PR #325, opened 2025-03-03
Doc: Explain scripts, bump version

akcano:
The sketch how-to presents a use case for scripts in the how-to section; the only part missing was the explanation.

### DIFF
--- a/docs/explanation/workshops/workshops.rst
+++ b/docs/explanation/workshops/workshops.rst
@@ -87,11 +87,11 @@ A simple definition might look like this:
 .. code-block:: yaml
    :caption: workshop.yaml
 
-   name: golang
+   name: dev
    base: ubuntu@22.04
    sdks:
      - name: go
-       channel: latest/stable
+       channel: jammy/stable
 
 
 .. @artefact SDK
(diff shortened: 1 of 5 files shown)


docs/explanation/workshops/workshops.rst
```diff
@@ -174,6 +174,65 @@ Also, both connections established here
...
+To run these scripts, you use the :command:`workshop run` command:
+
+.. code-block:: console
+
+   $ workshop run lint
+
+
+When you thus invoke a script, it's injected into the workshop
+and executed non-interactively
+(unless you change the mode with the :option:`!--interactive` flag).
```
jonathan-conder: I don't think this is true, it should work the same as exec (interactive is decided based on `stdin`/`stdout`)

dmitry-lyfar: Yes, `--interactive` is "enforcing" interactivity and it is optional. The mode is decided by the `run` client. So, if you `workshop run`  from a terminal, that would mean your terminal's in & out ptys will be associated with virtual ptys on the server (LXD). Same as `exec`.

dmitry-lyfar (approved):
Looks good except for that part re. interactivity.

dmitry-lyfar:
@akcano given we are on this page could we update this:
```
name: golang
base: ubuntu@22.04
sdks:
  - name: go
    channel: latest/stable
```

to

```
name: dev
base: ubuntu@22.04
sdks:
  - name: go
    channel: jammy/stable
```
?

dmitry-lyfar:
@akcano I can merge it, but it needs auto-squashing in your branch. It looks good to me.

docs/explanation/workshops/workshops.rst
```diff
@@ -91,7 +91,7 @@ A simple definition might look like this:
    base: ubuntu@22.04
    sdks:
      - name: go
-       channel: latest/stable
+       channel: jammy/stable
```
dmitry-lyfar, line 94: This workshop's name is `golang`, but later on in this doc `dev` is used.

PR #329, opened 2025-03-07
Fix wayland socket issue in spread

lachypjones:
# Description
Fixes: https://github.com/canonical/workshop/actions/runs/13665540077/job/38207709073

This doesn't fix the root cause (lxd trying to bind the socket before systemd has created the runtime directory) however in practice this does not seem to be an issue. 

We decided to only introduce the additional complexity required to symlink the socket if we encounter this issue in the wild. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/tests/main/interface-desktop/task.yaml
+++ b/tests/main/interface-desktop/task.yaml
@@ -51,6 +51,7 @@ execute: |
   EOF
 
   echo "Connect the desktop interface for an x11 session"
+  sudo -u ubuntu XDG_RUNTIME_DIR=/run/user/1000 systemctl --user unset-environment WAYLAND_DISPLAY
   sudo -u ubuntu touch /tmp/.Xauthority
   sudo -u ubuntu XDG_RUNTIME_DIR=/run/user/1000 systemctl --user set-environment DISPLAY=:0 XAUTHORITY=/tmp/.Xauthority
   workshop_exec connect ws-desktop/test-sdk-desktop:desktop :desktop


dmitry-lyfar:
@jonathan-conder I don not understand the fix, but I do understand the issue of the race between LXD and systemd. Can you explain this fix?

jonathan-conder:
> I do not understand the fix, but I do understand the issue of the race between LXD and systemd. Can you explain this fix?

It doesn't make sense to me either. By unsetting `WAYLAND_DISPLAY`, the connect will only try to proxy X11, and not wayland. That should be safe(r) because the X11 socket is in `/tmp` not `/run/user/1000`. But it seems like the issue could still occur further up, when testing the wayland socket.

jonathan-conder:
Looking at the error more closely, it happens when you do `workshop start` and the interface is already connected. Maybe `workshop launch` already waits for `/run/user/1000` so it's safe to connect at that point.

jonathan-conder:
Just consolidating my earlier comments:
- `WAYLAND_DISPLAY` can be an absolute path so we don't need to mess around with symlinks.
- GTK only added support in `4.11.1`. But all this means is that it no longer checks that `XDG_RUNTIME_DIR` is in the environment.
- The actual support is part of libwayland-client 1.15 (from 2018): https://lists.freedesktop.org/archives/wayland-devel/2018-April/037767.html. Focal has 1.18.

PR #333, opened 2025-03-11
Remove today from tests for workshop changes

jonathan-conder:
# Description

If these tests run at the start of a new day, the output can contain "yesterday" instead of "today." Rather than allowing yesterday we should just accept anything which looks like a human-readable date.

I checked that `timeutil.Human` is only used by `changes` and `tasks`, and only in the tests for `changes`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/tests/main/changes/task.yaml
+++ b/tests/main/changes/task.yaml
@@ -5,6 +5,6 @@ execute: |
   workshop_exec launch
 
   echo "The launch change must present in the list"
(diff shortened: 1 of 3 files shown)


dmitry-lyfar (approved):
One less possibility for a flaky run!

PR #335, opened 2025-03-11
Shell completion for shell, run and scripts

jonathan-conder:
# Description

Adds shell completion for shell, run and scripts. The run implementation requires an ugly workaround for https://github.com/spf13/cobra/issues/1877. I also found a slightly better workaround here: https://github.com/spf13/cobra/pull/2259.

Fixes a few other issues:
- Some commands were completing multiple workshop names despite only accepting one.
- Completion was previously able to print the "new warnings" message.
- Completion didn't work if the command had a relative `--project` path.

Errors during completion can be debugged more easily by setting `BASH_COMP_DEBUG_FILE=/tmp/err.log`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -247,7 +247,8 @@ $ workshop shell nimble
 
 The name is optional if the project has only one workshop:
 $ workshop shell`,
-		RunE: c.Run,
+		RunE:              c.Run,
+		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Waiting"}),
 	}
 
 	return cmd
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -367,7 +367,7 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 		return err
 	}
 
-	project, err := cli.Project(root.project)
+	project, err := cli.Project(root.project())
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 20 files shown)


dmitry-lyfar (approved):

lachypjones (approved):

cmd/workshop/exec.go
```diff
@@ -604,37 +641,42 @@ $ workshop scripts nimble
...
 }
 
 func (c *CmdScripts) Run(cmd *cobra.Command, av []string) error {
+	scripts, err := c.scripts(av)
+	if err != nil || len(scripts) == 0 {
+		return err
+	}
+
+	encoder := yaml.NewEncoder(Stdout)
+	encoder.SetIndent(2)
```
dmitry-lyfar, line 658: Thanks, that made me think we shall use the same indent size where save the workshop definition in `user.workshop.file`

dmitry-lyfar (approved):
Tested on a workshop with zero, one and many scripts. All works well. In the zero scripts case it auto completes the name and stops.

cmd/workshop/exec.go
jonathan-conder: I think, after the tunnel interface, `workshop info` may no longer be valid YAML, e.g. if there's an abstract Unix socket or an IPv6 address in square brackets. Might be time to use an encoder there also? The indentation will be tricky to get right though.

PR #338, opened 2025-03-13
Improve refresh performance with snapshots

dmitry-lyfar:
# Description

This PR extends the refresh resolver to restore a workshop from a snapshot if it's available to speed up refresh changes. The resolver makes a decision about the exact snapshot that the workshop needs to be restored from based on the updates available for the SDKs that constitute the workshop, changes to the connections or plugs & slots in the workshop definitions. Briefly speaking, if a workshop has been defined as:

```
name: dev
base: ubuntu@22.04
sdks:
  - name: ros2
    channel: latest/stable
  - name: nav2
    channel: latest/stable
```

and `nav2` SDK was the only one that had an update, the resolver would restore the workshop to the snapshot created after installing `ros2` SDK and install a new revision of `nav2` on top of that.

If a workshop's base was changed, the workshop will be rebuilt from scratch by using LXD's `RebuildInstance` API and providing it with a required image.

If the `sketch` SDK exists, it will be refreshed every time as there is no mechanism to detect updates to the local SDKs.

## Implementation notes

The behaviour of stashing has changed from moving the instance with `MigrateInstance` to copying the instance with `CopyInstance` to the stash area (with all the snapshots). This allows rebuilding the workshop in-place and restoring if something went wrong after restoring from an earlier snapshot.

## Out of scope

The `refresh` CLI will be extended with `--rebuild` and `--restore` options to rebuild the entire workshop from scratch or restore it to the last up-to-date snapshot respectively. These options is not part of the PR.

## Breaking changes

* The storage pool is created with `volume.zfs.remove_snapshots: "true"` which allows for removing previous snapshots on restore automatically.
* `system` SDK is tracked as one of the workshop's installed SDKs in `user.workshop.sdks` without any special treatment in operations like `link` or `unlink` SDKs. 

## Known issues

* https://github.com/canonical/lxd/issues/15078

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -225,9 +225,6 @@ func tunnels(conns *connectionsJSON) (map[string]*TunnelInfo, error) {
 		}
 
 		sk := conn.Plug.Sdk
-		if sdk.IsSystem(sk) {
-			sk = conn.Slot.Sdk
-		}
 		info, ok := tunnels[sk]
 		if !ok {
 			info = &TunnelInfo{}
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -119,19 +119,21 @@ func workshopFileToInfo(pid string, name string, path string) *WorkshopFileInfo 
 	return &ws
 }
 
-func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health healthstate.HealthState) *WorkshopInfo {
+func workshopToInfo(ctx context.Context, w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
 	var info WorkshopInfo
 	info.Name = w.Name
 	info.ProjectId = w.Project.ProjectId
 	info.Base = w.Base
 
+	sdks, err := w.SdkInfosByInstallOrder(ctx)
+	if err != nil {
+		return nil, err
+	}
+
 	mnts := w.Mounts(sdks)
 
-	for _, sk := range w.Sdks {
-		sdkInfo := sdks[sk.Name]
-		if sdkInfo == nil {
-			sdkInfo = &sdk.Info{}
-		}
+	for _, sk := range sdks {
+		sdkInfo := sk
 
 		var healthInfo *HealthCheckInfo
 		if sdkHealth, ok := health.SdkHealth[sk.Name]; ok {
@@ -151,7 +153,7 @@ func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health heal
 			Channel:     sk.Channel,
 			Revision:    sk.Revision.String(),
 			BuildTime:   sdkInfo.BuildTime,
-			InstallTime: sk.InstallTime,
+			InstallTime: w.Sdks[sk.Name].InstallTime,
 			Health:      healthInfo,
 			Mounts:      mntInfos,
 		})
@@ -162,7 +164,7 @@ func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health heal
 	}
 	info.Status = health.Status.String()
 
-	return &info
+	return &info, nil
 }
 
 func mountInfos(sk sdk.Ref, mnts []workshop.Mount) []*Mount {
@@ -381,10 +383,14 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
 
 	info := Workshops{}
 	info.Workshops = make([]*WorkshopInfo, 0, len(workshops))
+	ctx := context.WithValue(r.Context(), workshop.ContextProjectId, projectId)
 	for _, w := range workshops {
 		health := healthstate.WorkshopHealth(state, w)
 		if ignoreStatus || health.Status == status {
-			wi := workshopToInfo(w, nil, health)
+			wi, err := workshopToInfo(ctx, w, health)
+			if err != nil {
+				return statusBadRequest("%w", err)
+			}
 			info.Workshops = append(info.Workshops, wi)
 		}
 	}
@@ -590,11 +596,11 @@ func v1GetProjectWorkshop(c *Command, r *http.Request, _ *userState) Response {
 	health := healthstate.WorkshopHealth(state, w)
 
 	ctx := context.WithValue(r.Context(), workshop.ContextProjectId, projectId)
-	sdks, err := w.SdkInfos(ctx)
+
+	info, err := workshopToInfo(ctx, w, health)
 	if err != nil {
 		return statusBadRequest("%w", err)
 	}
-	info := workshopToInfo(w, sdks, health)
 
 	ts, err := tunnels(conns)
 	if err != nil {
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -465,7 +465,7 @@ func refresh(ctx context.Context, st *state.State, mgr *workshopstate.WorkshopMa
 		if sk != sdk.Sketch {
 			return change, taskset, fmt.Errorf(`partial refresh is supported only for "sketch" SDK`)
 		}
-		taskset, err = mgr.RefreshLocalSdk(ctx, pid, wp, sk)
+		taskset, err = mgr.RefreshMany(ctx, pid, []string{wp})
 	} else {
 		change = newWorkshopChange(st, "refresh", user, pid, reqData.Action, reqData.Names)
 		taskset, err = mgr.RefreshMany(ctx, pid, reqData.Names)
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -121,7 +121,7 @@ func workshopFileToInfo(pid string, name string, path string) *WorkshopFileInfo 
 
 // Returns essential workshop properties and its SDKs health statuses (if
 // available).
-func workshopToInfo(w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
+func workshopToInfo(w *workshop.Workshop, health healthstate.HealthState) *WorkshopInfo {
 	var info WorkshopInfo
 	info.Name = w.Name
 	info.ProjectId = w.Project.ProjectId
@@ -152,7 +152,7 @@ func workshopToInfo(w *workshop.Workshop, health healthstate.HealthState) (*Work
 		info.Notes = append(info.Notes, health.Code)
 	}
 	info.Status = health.Status.String()
-	return &info, nil
+	return &info
 }
 
 // Returns essential workshop properties, SDK health statuses (if available) and
@@ -421,10 +421,7 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
 	for _, w := range workshops {
 		health := healthstate.WorkshopHealth(state, w)
 		if ignoreStatus || health.Status == status {
-			wi, err := workshopToInfo(w, health)
-			if err != nil {
-				return statusBadRequest("%w", err)
-			}
+			wi := workshopToInfo(w, health)
 			info.Workshops = append(info.Workshops, wi)
 		}
 	}
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"path/filepath"
 	"slices"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -162,25 +161,6 @@ func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health heal
 		info.Notes = append(info.Notes, health.Code)
 	}
 	info.Status = health.Status.String()
-
-	// Sort the SDKs in installation order.
-	orderMap := make(map[string]int)
-	for i, v := range w.File.Sdks {
-		// Ignore the order of the system SDK from the definion (if exists).
-		if sdk.IsSystem(v.Name) {
-			continue
-		}
-		orderMap[v.Name] = i
-	}
-	sort.Slice(info.Sdks, func(i, j int) bool {
-		// system SDK is always first. Note: it may or may not present in the
-		// original file in an arbitrary order, but it will always be in the
-		// list of installed SDKs.
-		if sdk.IsSystem(info.Sdks[i].Name) || sdk.IsSystem(info.Sdks[j].Name) {
-			return true
-		}
-		return orderMap[info.Sdks[i].Name] < orderMap[info.Sdks[j].Name]
-	})
 
 	return &info
 }
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -133,8 +133,6 @@ func workshopToInfo(ctx context.Context, w *workshop.Workshop, health healthstat
 	mnts := w.Mounts(sdks)
 
 	for _, sk := range sdks {
-		sdkInfo := sk
-
 		var healthInfo *HealthCheckInfo
 		if sdkHealth, ok := health.SdkHealth[sk.Name]; ok {
 			healthInfo = &HealthCheckInfo{
@@ -149,10 +147,10 @@ func workshopToInfo(ctx context.Context, w *workshop.Workshop, health healthstat
 
 		info.Sdks = append(info.Sdks, &SdkInfo{
 			Name:        sk.Name,
-			Version:     sdkInfo.Version,
+			Version:     sk.Version,
 			Channel:     sk.Channel,
 			Revision:    sk.Revision.String(),
-			BuildTime:   sdkInfo.BuildTime,
+			BuildTime:   sk.BuildTime,
 			InstallTime: w.Sdks[sk.Name].InstallTime,
 			Health:      healthInfo,
 			Mounts:      mntInfos,
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"path/filepath"
 	"slices"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -162,6 +163,25 @@ func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health heal
 	}
 	info.Status = health.Status.String()
 
+	// Sort the SDKs in installation order.
+	orderMap := make(map[string]int)
+	for i, v := range w.File.Sdks {
+		// Ignore the order of the system SDK from the definion (if exists).
+		if sdk.IsSystem(v.Name) {
+			continue
+		}
+		orderMap[v.Name] = i
+	}
+	sort.Slice(info.Sdks, func(i, j int) bool {
+		// system SDK is always first. Note: it may or may not present in the
+		// original file in an arbitrary order, but it will always be in the
+		// list of installed SDKs.
+		if sdk.IsSystem(info.Sdks[i].Name) || sdk.IsSystem(info.Sdks[j].Name) {
+			return true
+		}
+		return orderMap[info.Sdks[i].Name] < orderMap[info.Sdks[j].Name]
+	})
+
 	return &info
 }
 
@@ -223,7 +243,7 @@ func tunnels(conns *connectionsJSON) (map[string]*TunnelInfo, error) {
 		}
 
 		sk := conn.Plug.Sdk
-		if sk == sdk.System.String() {
+		if sdk.IsSystem(sk) {
 			sk = conn.Slot.Sdk
 		}
 		info, ok := tunnels[sk]
@@ -389,11 +409,10 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
 		}
 	}
 
-	// If the client queried everything available,
-	// we add workshop files to the response.
-	// Some of these may only exist as files, not instances.
-	// The "available" query is a best-effort version of "all":
-	// if something is wrong with the files we still return the instances.
+	// If the client queried everything available, we add workshop files to the
+	// response. Some of these may only exist as files, not instances. The
+	// "available" query is a best-effort version of "all": if something is
+	// wrong with the files we still return the instances.
 	if ignoreStatus {
 		files, err := wrkmgr.WorkshopFiles(r.Context(), projectId)
 		var fileErr *workshopstate.WorkshopFileError
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -119,7 +119,47 @@ func workshopFileToInfo(pid string, name string, path string) *WorkshopFileInfo 
 	return &ws
 }
 
-func workshopToInfo(ctx context.Context, w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
+// Returns essential workshop properties and its SDKs health statuses (if
+// available).
+func workshopToInfo(w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
+	var info WorkshopInfo
+	info.Name = w.Name
+	info.ProjectId = w.Project.ProjectId
+	info.Base = w.Base
+
+	sdkSetups := w.SdksByInstallOrder()
+
+	for _, sk := range sdkSetups {
+		var healthInfo *HealthCheckInfo
+		if sdkHealth, ok := health.SdkHealth[sk.Name]; ok {
+			healthInfo = &HealthCheckInfo{
+				Timestamp: sdkHealth.Timestamp,
+				Message:   sdkHealth.Message,
+				Code:      sdkHealth.Code,
+			}
+		}
+
+		info.Sdks = append(info.Sdks, &SdkInfo{
+			Name:        sk.Name,
+			Channel:     sk.Channel,
+			Revision:    sk.Revision.String(),
+			InstallTime: sk.InstallTime,
+			Health:      healthInfo,
+		})
+	}
+
+	if len(health.Code) > 0 {
+		info.Notes = append(info.Notes, health.Code)
+	}
+	info.Status = health.Status.String()
+	return &info, nil
+}
+
+// Returns essential workshop properties, SDK health statuses (if available) and
+// information about the installed SDKs. This function reads from the workshop's
+// filesystem to obtain certain fields, it has to be used with possible latency
+// changes in mind.
+func workshopToInfoFull(ctx context.Context, w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
 	var info WorkshopInfo
 	info.Name = w.Name
 	info.ProjectId = w.Project.ProjectId
@@ -378,11 +418,10 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
 
 	info := Workshops{}
 	info.Workshops = make([]*WorkshopInfo, 0, len(workshops))
-	ctx := context.WithValue(r.Context(), workshop.ContextProjectId, projectId)
 	for _, w := range workshops {
 		health := healthstate.WorkshopHealth(state, w)
 		if ignoreStatus || health.Status == status {
-			wi, err := workshopToInfo(ctx, w, health)
+			wi, err := workshopToInfo(w, health)
 			if err != nil {
 				return statusBadRequest("%w", err)
 			}
@@ -592,7 +631,7 @@ func v1GetProjectWorkshop(c *Command, r *http.Request, _ *userState) Response {
 
 	ctx := context.WithValue(r.Context(), workshop.ContextProjectId, projectId)
 
-	info, err := workshopToInfo(ctx, w, health)
+	info, err := workshopToInfoFull(ctx, w, health)
 	if err != nil {
 		return statusBadRequest("%w", err)
 	}
--- a/internal/daemon/api_workshops_test.go
+++ b/internal/daemon/api_workshops_test.go
@@ -418,8 +418,9 @@ func (s *apiSuite) TestGetWorkshops(c *check.C) {
 
 	_, err = rsp.MarshalJSON()
 	c.Assert(err, check.IsNil)
-	// for DeepEqual to work correctly
-	install1 := s.installTime
+
+	install1 := time.Date(2023, 04, 25, 1, 2, 3, 0, time.UTC)
+	build1 := time.Date(2020, 4, 22, 19, 12, 7, 903032000, time.UTC)
 	info := rsp.Result.(Workshops)
 
 	c.Check(info.Workshops, testutil.DeepUnsortedMatches, []*WorkshopInfo{{
@@ -436,11 +437,12 @@ func (s *apiSuite) TestGetWorkshops(c *check.C) {
 			{
 				Name:        "test-sdk",
 				Channel:     "latest/stable",
+				Version:     "0.1.2",
 				Revision:    "1",
 				InstallTime: &install1,
+				BuildTime:   &build1,
 			},
 		},
-		Notes: nil,
 	}, {
 		Name:      "basic",
 		Base:      "ubuntu@22.04",
--- a/internal/daemon/api_workshops_test.go
+++ b/internal/daemon/api_workshops_test.go
@@ -143,6 +143,17 @@ sdks:
 base: ubuntu@22.04
 `
 
+	manysdks_extended = `name: manysdks
+base: ubuntu@22.04
+sdks:
+  - name: test-sdk-3
+    channel: latest/stable
+  - name: test-sdk-2
+    channel: latest/stable
+  - name: test-sdk
+    channel: latest/stable
+`
+
 	workshoptunnels = `name: tunnels
 base: ubuntu@22.04
 sdks:
@@ -358,6 +369,16 @@ slots:
     workshop-source: /mnt
     interface: mount
 `
+
+	testsdk3 = `
+name: test-sdk-3
+title: title
+base: ubuntu@20.04
+version: '20200401.3f3a63f'
+summary: summary
+description: SDK
+sdkcraft-started-at: '2020-05-03T22:05:35.811829Z'
+`
 )
 
 type testSdk struct {
@@ -368,6 +389,7 @@ type testSdk struct {
 var apiSuiteSdks = map[string]testSdk{
 	"test-sdk":   {s: sdk.Setup{Name: "test-sdk", Revision: sdk.R(1)}, meta: testsdk},
 	"test-sdk-2": {s: sdk.Setup{Name: "test-sdk-2", Revision: sdk.R(1)}, meta: testsdk2},
+	"test-sdk-3": {s: sdk.Setup{Name: "test-sdk-3", Revision: sdk.R(1)}, meta: testsdk3},
 }
 
 func (s *apiSuite) launchWorkshop(c *check.C, name, yaml string, sdks map[string]testSdk) {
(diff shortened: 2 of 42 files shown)


internal/daemon/api_workshops.go
```diff
@@ -119,19 +119,21 @@ func workshopFileToInfo(pid string, name string, path string) *WorkshopFileInfo
...
+
 	mnts := w.Mounts(sdks)
 
-	for _, sk := range w.Sdks {
-		sdkInfo := sdks[sk.Name]
-		if sdkInfo == nil {
-			sdkInfo = &sdk.Info{}
-		}
+	for _, sk := range sdks {
+		sdkInfo := sk
```
jonathan-conder:
```suggestion
	for _, sk := range sdks {
```
If anything I'd maybe put `sdkSetup := w.Sdks[sk.Name]` instead

internal/daemon/api_workshops.go
```diff
@@ -223,9 +225,6 @@ func tunnels(conns *connectionsJSON) (map[string]*TunnelInfo, error) {
 		}
 
 		sk := conn.Plug.Sdk
-		if sk == sdk.System.String() {
-			sk = conn.Slot.Sdk
-		}
```
jonathan-conder, line 228: needs further refactoring. For example `if sk == conn.Plug.Sdk` below is always true now. I can address in a follow-up PR though.

internal/daemon/api_workshops.go
```diff
@@ -381,19 +380,22 @@ func v1GetProjectWorkshops(c *Command, r *http.Request, _ *userState) Response {
...
 	info.Workshops = make([]*WorkshopInfo, 0, len(workshops))
+	ctx := context.WithValue(r.Context(), workshop.ContextProjectId, projectId)
 	for _, w := range workshops {
 		health := healthstate.WorkshopHealth(state, w)
 		if ignoreStatus || health.Status == status {
-			wi := workshopToInfo(w, nil, health)
+			wi, err := workshopToInfo(ctx, w, health)
+			if err != nil {
+				return statusBadRequest("%w", err)
+			}
```
jonathan-conder: Including the SDKs in `workshop list` should reduce the number of API calls needed for `workshop info` (without a workshop name)

internal/overlord/hookstate/handlers.go
```diff
@@ -49,35 +49,43 @@ func (h *HookManager) doRunHook(task *state.Task, tomb *tomb.Tomb) error {
...
 		return h.executeHook(ctx, task, w, prj.ProjectId, &hook)
+	case SetupBase:
+		if err = h.executeHook(ctx, task, w, prj.ProjectId, &hook); err != nil {
+			return err
+		}
+		if err = h.backend.Snapshot(ctx, w, workshop.SnapshotId(w, hook.Sdk)); err != nil {
+			return err
+		}
+		return nil
+
```
jonathan-conder:
```suggestion
		return h.backend.Snapshot(ctx, w, workshop.SnapshotId(w, hook.Sdk))
```

internal/overlord/sdkstate/handlers.go
```diff
@@ -180,7 +180,7 @@ func (m *SdkManager) doInstallSdk(task *state.Task, tomb *tomb.Tomb) error {
 	return m.backend.AttachVolume(ctx, w, sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision.String()), sdkPath, true)
 }
 
-func (m *SdkManager) undoInstallSdk(task *state.Task, tomb *tomb.Tomb) error {
+func (m *SdkManager) doUnistallSdk(task *state.Task, tomb *tomb.Tomb) error {
```
```suggestion
func (m *SdkManager) doUninstallSdk(task *state.Task, tomb *tomb.Tomb) error {
```

internal/overlord/sdkstate/handlers.go
```diff
@@ -225,9 +225,9 @@ func (m *SdkManager) doLinkSdk(task *state.Task, tomb *tomb.Tomb) error {
 
 	st := task.State()
 	rev.Add(func() {
-		if err := wp.UnlinkSdk(ctx, setup.Name); err != nil {
+		if reverr := wp.UnlinkSdk(ctx, setup.Name); reverr != nil {
```
jonathan-conder, line 228: Renaming this doesn't seem necessary as far as I can tell.

internal/overlord/sdkstate/handlers.go
```diff
@@ -254,28 +254,32 @@ func (m *SdkManager) doLinkSdk(task *state.Task, tomb *tomb.Tomb) error {
...
+	if err = wp.UnlinkSdk(ctx, setup.Name); err != nil {
 		return err
 	}
 
-	return wp.UnlinkSdk(ctx, sdkSetup.Name)
+	if err = m.repo.RemoveSdk(project.ProjectId, w, setup.Name); err != nil {
+		return err
+	}
+
+	return nil
```
```suggestion
	return m.repo.RemoveSdk(project.ProjectId, w, setup.Name)
```

internal/overlord/workshopstate/handlers.go
```diff
@@ -80,27 +80,43 @@ func (m *WorkshopManager) doCreateWorkshop(task *state.Task, tomb *tomb.Tomb) er
...
 	rev.Add(func() {
 		cleanupCtx, cancel := context.WithTimeout(cleanupCtx, 30*time.Second)
 		defer cancel()
-		if err1 := m.backend.RemoveWorkshop(cleanupCtx, w); err1 != nil {
-			logger.Noticef("On doCreateWorkshop: cannot remove %q workshop on cleanup: %v", w, err1)
+
+		// This may fail if the first workshop launch has failed for some
+		// reason; It is safe to ignore the error in that case.
+		if reverr := m.backend.RemoveWorkshop(cleanupCtx, w); reverr != nil {
+			logger.Noticef("On doCreateWorkshop: cannot remove %q workshop on cleanup: %v", w, reverr)
```
jonathan-conder: I don't understand why this has moved before launch/restore. Are those not transactional?

internal/overlord/workshopstate/request.go
```diff
@@ -20,20 +20,23 @@ import (
...
 	// mark the tasks that denote irreversible clean up logic for refresh (e.g.
 	// removing state storage and the old workshop copy)
-	EdgeRefreshCleanup = state.TaskSetEdge("refresh-cleanup")
+	EdgeRefreshFirstCleanupTask = state.TaskSetEdge("refresh-cleanup")
 )
 
-var checkHealthTimeout = 5 * time.Second
+var (
+	checkHealthTimeout = 5 * time.Second
+)
```
jonathan-conder: doesn't seem necessary

internal/overlord/workshopstate/request.go
```diff
@@ -77,153 +80,212 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
...
 
-	install := installSdks(st, file.Name, sdks, retrieve)
-	install.WaitAll(system)
+	retrieve, rmap := retrieveSdks(st, sdks)
+	addTaskSet(retrieve)
 
-	full := slices.Clone(sdks)
-	full = slices.Insert(full, 0, sdk.Setup{Name: sdk.System.String(), Revision: sdk.Revision{N: -1}})
+	createAptCache := st.NewTask("create-apt-cache", fmt.Sprintf("Create apt cache for %q", file.Name))
+	addTaskSet(state.NewTaskSet(createAptCache))
```
jonathan-conder, line 273: Previously the apt cache was created in parallel with retrieving the SDKs. Could keep that behaviour by sneaking the task into `retrieve`.

internal/overlord/workshopstate/request.go
```diff
@@ -274,7 +336,14 @@ func (w *WorkshopManager) RefreshMany(ctx context.Context, projectId string, nam
 
 	taskset := make([]*state.TaskSet, 0, len(refreshReqs))
 	for _, req := range refreshReqs {
-		tasks, err := refresh(ctx, w.state, sto, req.w, req.file)
+		w.state.Unlock()
+		plan, err := resolveRefresh(ctx, sto, req.w, req.file)
+		w.state.Lock()
```
jonathan-conder: I looked into this yesterday. If we want to prevent concurrent refreshes, it's not safe to unlock the state until there's at least one task associated with the `refresh` change (empty changes don't show up in `CheckChangeConflict`).

It's not clear from the diff, but the status check is just above here.

internal/overlord/workshopstate/request.go
```diff
@@ -77,153 +80,212 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
...
 
-	var prevSetup *state.Task
-	setupHook := state.NewTaskSet()
+		if prevInstall != nil {
+			ts.WaitAll(prevInstall)
+		}
+		prevInstall = ts
+
+		all.AddAll(ts)
+	}
```
jonathan-conder: Seems common enough that we could create `TaskSequence` and `TaskSetSequence` abstractions.

internal/overlord/workshopstate/manager.go
```diff
@@ -26,7 +26,7 @@ func New(st *state.State, runner *state.TaskRunner) *WorkshopManager {
 	st.Unlock()
 
 	runner.AddHandler("download-base", OnDo(manager.doDownloadBase), nil)
-	runner.AddHandler("create-workshop", OnDo(manager.doCreateWorkshop), manager.undoCreateWorkshop)
+	runner.AddHandler("create-workshop", OnDo(manager.doConstructWorkshop), manager.undoCreateWorkshop)
```
jonathan-conder:
```suggestion
	runner.AddHandler("create-workshop", OnDo(manager.doConstructWorkshop), manager.undoConstructWorkshop)
```

internal/overlord/workshopstate/request.go
```diff
@@ -77,153 +80,212 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
...
+		construct.AddTask(t)
+		if prev != nil {
+			t.WaitFor(prev)
+		}
+		prev = t
+	}
+
+	base := st.NewTask("download-base", fmt.Sprintf("Download %q base image", file.Base))
+	base.Set("workshop-base", file.Base)
+	addTask(base)
```
jonathan-conder: Do we need to download the base when restoring from a snapshot?

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
-		}
-
-		idx = slices.IndexFunc(p.refresh, contains)
-		if idx != -1 {
-			ordered = append(ordered, p.refresh[idx])
-			continue
+			idx := slices.IndexFunc(setup, contains)
+			if idx != -1 {
+				ordered = append(ordered, setup[idx])
+			}
```
```suggestion
				ordered = append(ordered, setup[idx])
				break
			}
```

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
 
-	installOrder  []string
-	refreshAnyway bool
+	sdkSnapshot  string
+	installOrder []string
 }
 
-func (p refreshPlan) InstallOrRefresh() []sdk.Setup {
+func (p refreshPlan) ordered(setups ...[]sdk.Setup) []sdk.Setup {
 	ordered := []sdk.Setup{}
```
```suggestion
	ordered := make([]sdk.Setup, 0, len(p.installOrder))
```

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
-		// If the SDK is not in the list of the existing SDKs -- it is being
-		// installed, not refreshed.
-		if installed, exist := w.Sdks[info.Name]; exist {
-			plan.refreshAnyway = maybeRefresh(installed, info.Info)
-			if plan.refreshAnyway {
+	// Determine if a workshop can be partially updated.
+	lastIntact := 0
+	for ni, s := range candidates {
+		// Do we have this SDK in the same order as in the running workshop?
+		if ni < len(installed) && installed[ni].Name == s.Name {
```
jonathan-conder: If this condition is `false`, I think we need to `break`.

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
+				sdk.Setup{Name: sdk.Sketch, Revision: sdk.Revision{N: -1}})
 		}
 	}
 
+	if fullRefresh {
+		plan.refresh = append(plan.refresh, plan.intact...)
+		plan.remove = append(plan.remove, plan.intact...)
+		plan.intact = plan.intact[:0]
+		plan.sdkSnapshot = ""
+	}
```
jonathan-conder: I think this is equivalent to skipping the loop `for ni, s := range candidates {`. Doing that would be a bit simpler, and also (I think) mean that `plan.refresh` is in installation order. The `w.Base` check can be moved up, but unsure about the plugs/slots check.

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
+	return p.ordered(p.install, p.refresh)
 }
 
-func (p refreshPlan) RefreshOrRemove() []sdk.Setup {
-	ordered := []sdk.Setup{}
-
-	for _, s := range p.installOrder {
-		contains := func(sp sdk.Setup) bool { return s == sp.Name }
+func (p refreshPlan) IntactOrRemove() []sdk.Setup {
+	ordered := p.ordered(p.intact)
```
jonathan-conder: `p.intact` is already in installation order. Based on usage I think `intact + remove` should be sorted in reverse order from `w.SdksByInstallOrder()`.

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
+func (p refreshPlan) Remove() []sdk.Setup {
+	return p.remove
 }
 
 func resolveRefresh(ctx context.Context, sto sdk.Store, w *workshop.Workshop, file *workshop.File) (*refreshPlan, error) {
+	fullRefresh := false
+
 	plan := &refreshPlan{
 		install:      make([]sdk.Setup, 0),
 		refresh:      make([]sdk.Setup, 0),
```
jonathan-conder: Maybe `intact` should be here for consistency.

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
-	}
-	return ordered
+func (p refreshPlan) Refresh() []sdk.Setup {
+	return p.ordered(p.refresh)
 }
 
-func (p refreshPlan) RecreateWorkshop() bool {
-	return p.refreshAnyway || len(p.install) != 0 || len(p.refresh) != 0 || len(p.remove) != 0
+func (p refreshPlan) Remove() []sdk.Setup {
+	return p.remove
```
jonathan-conder: Can also be sorted in reverse install order.

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +406,163 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
-		plan.refreshAnyway = true
+	if len(plan.refresh) == 0 && len(plan.remove) == 0 && len(plan.install) == 0 {
+		plugsSlotsChanged := func(s1, s2 workshop.SdkRecord) bool {
+			return reflect.DeepEqual(s1.Plugs, s2.Plugs) &&
+				reflect.DeepEqual(s1.Slots, s2.Slots)
+		}
+		if !slices.EqualFunc(w.File.Sdks, file.Sdks, plugsSlotsChanged) {
+			fullRefresh = true
+		}
 	}
```
jonathan-conder: I can't make sense of this part. If you change one plug/slot it requires a full refresh, but only if none of the SDKs have updates?

internal/overlord/hookstate/handlers.go
```diff
@@ -49,35 +49,43 @@ func (h *HookManager) doRunHook(task *state.Task, tomb *tomb.Tomb) error {
...
-			err = fs.MkdirAll(hook.Environment["SDK_STATE_DIR"], 0755)
-			fs.Close()
-			if err != nil {
-				return fmt.Errorf("cannot run hook \"save-state\" for %q SDK: %v", hook.Sdk, err)
-			}
+		fs, err := h.backend.WorkshopFs(ctx, w)
+		if err != nil {
+			return fmt.Errorf("cannot run hook \"save-state\" for %q SDK: %v", hook.Sdk, err)
 		}
+		err = fs.MkdirAll(hook.Environment["SDK_STATE_DIR"], 0755)
```
jonathan-conder, line 56: Not really related, but the save/restore hooks do a fair bit of work before checking whether the hook exists. It looks easy enough to move the check further up.

internal/overlord/workshopstate/request.go
```diff
@@ -513,107 +577,98 @@ func refresh(ctx context.Context, st *state.State, sto sdk.Store, w *workshop.Wo
...
+	addTaskSet(saveState)
+
+	disconnect := disconnectSdks(st, plan.IntactOrRemove())
+	addTaskSet(disconnect)
+
+	// Unlink and remove SDKs from interfaces repository. If refresh fails, the
+	// SDKs will be linked back and returned to the repository. This is the
+	// reason unlink has to happen before stashing. As the workshop, that will
+	// be reconstructed after stashing, will link and add their SDKs to the
+	// repository in `installSdks` step.
```
```suggestion
	// Unlink and remove SDKs from interfaces repository. If refresh fails, the
	// SDKs will be linked back and returned to the repository. This is the
	// reason unlink has to happen before stashing. The workshop that will
	// be reconstructed after stashing will link and add the SDKs to the
	// repository in `installSdks` step.
```
I don't see where the `installSdks` step comes in, I think the undo step of `unlinkSdks` is the thing which re-links them.

internal/overlord/sdkstate/manager.go
```diff
@@ -21,9 +21,12 @@ func New(s *state.State, runner *state.TaskRunner, repo *interfaces.Repository)
...
 	runner.AddHandler("retrieve-sdk", OnDo(manager.doRetrieveSdk), nil)
-	runner.AddHandler("install-sdk", OnDo(manager.doInstallSdk), manager.undoInstallSdk)
-	runner.AddHandler("install-local-sdk", OnDo(manager.doInstallLocalSdk), manager.undoInstallLocalSdk)
-	runner.AddHandler("link-sdk", OnDo(manager.doLinkSdk), manager.undoLinkSdk)
+	runner.AddHandler("install-sdk", OnDo(manager.doInstallSdk), manager.doUnistallSdk)
+	runner.AddHandler("install-local-sdk", OnDo(manager.doInstallLocalSdk), manager.doUninstallLocalSdk)
+	runner.AddHandler("link-sdk", OnDo(manager.doLinkSdk), manager.doUnlinkSdk)
+	runner.AddHandler("unlink-sdk", OnDo(manager.doUnlinkSdk), manager.doLinkSdk)
+	runner.AddHandler("remove-sdk", OnDo(manager.doUnistallSdk), nil)
+	runner.AddHandler("remove-local-sdk", OnDo(manager.doUninstallLocalSdk), manager.doInstallLocalSdk)
```
jonathan-conder: Doesn't seem to be used currently.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -257,13 +263,51 @@ func (s *Backend) LaunchWorkshop(ctx context.Context, file *workshop.File) error
...
+		op, err := conn.UpdateInstance(rebuilt.Name, rebuilt.Writable(), etag)
+		if err != nil {
+			return err
+		}
+		if err = op.Wait(); err != nil {
+			return err
+		}
+		return nil
+	} else {
+		op, err := conn.CreateInstance(req)
```
jonathan-conder: `req` is only used in this branch, can be defined here. There's no need for `else` either.

internal/workshop/backend.go
```diff
@@ -165,8 +167,11 @@ type Backend interface {
 	// Returns a list of workshops for the project in context.
 	ProjectWorkshops(ctx context.Context) ([]*Workshop, error)
 
-	// Launch a barebone workshop instance using the base provided.
-	LaunchWorkshop(ctx context.Context, file *File) error
+	// Launch a barebone workshop instance. If the workshop exists, wipe out its
+	// rootfs and rebuild it from the workshop file's base image. The
+	// configuration and devices of the rebuilt workshop will be reset to the
+	// default one.
```
jonathan-conder, line 173: I don't think this is true, the implementation copies over the old config and devices.

internal/overlord/workshopstate/request.go
```diff
@@ -513,107 +577,98 @@ func refresh(ctx context.Context, st *state.State, sto sdk.Store, w *workshop.Wo
...
-		addTaskSet(system)
-	}
-
-	install := installSdks(st, file.Name, plan.InstallOrRefresh(), retrieve)
+		addTaskSet(state.NewTaskSet(stateStorage))
+	}
+
+	// Call save-state hooks for the SDKs that are installed and will not be
+	// removed after this refresh.
+	saveState := runHooks(st, w.Project.ProjectId, file.Name, plan.Refresh(), 0, hookstate.SaveState)
```
jonathan-conder, line 616: It's a little bit misleading because `plan.Refresh()` has the new revisions, not the installed ones. But `runHooks` only looks at the names so it's OK.

internal/overlord/workshopstate/request.go
```diff
@@ -633,75 +688,25 @@ func (w *WorkshopManager) RefreshLocalSdk(ctx context.Context, pid string, wpn s
 		return nil, fmt.Errorf("cannot refresh %q: %w", wpn, err)
 	}
 
-	ts, err := w.refreshLocalSdk(wp, sdkn)
```
jonathan-conder, line 636: As discussed, `RefreshLocalSdk` no longer uses the SDK name. The main difference from a normal refresh is it loads the workshop file from LXD rather than from disk.

Eventually I think it should merge with a normal refresh, but avoid querying the store if the channel is unmodified. My assumption is that interfaces can be reconnected even when restoring from a later snapshot.

internal/overlord/workshopstate/request.go
```diff
@@ -633,75 +688,25 @@ func (w *WorkshopManager) RefreshLocalSdk(ctx context.Context, pid string, wpn s
...
-		task.Set("workshop", wp.Name)
-		task.Set("project", wp.Project)
+	ts, err := refresh(ctx, w.state, plan, wp, file)
+	if err != nil {
+		return nil, err
 	}
-
-	return ts, nil
+	taskset = append(taskset, ts)
+	return taskset, nil
```
```suggestion
	return ts, nil
```
No need for a slice anymore

internal/workshop/backend.go
```diff
@@ -202,6 +207,11 @@ type Backend interface {
 	Exec(ctx context.Context, name string, args *Execution) (ExecContext, error)
 }
 
+type Snapshot interface {
+	Snapshot(ctx context.Context, workshop, snapid string) error
+	Restore(ctx context.Context, workshop, snapid string, file *File) error
```
jonathan-conder, line 212:
```suggestion
	SnapshotWorkshop(ctx context.Context, workshop, snapid string) error
	RestoreWorkshop(ctx context.Context, workshop, snapid string, file *File) error
```

I think this is more consistent with the other `Backend` methods.

internal/workshop/fakebackend/backend.go
```diff
@@ -335,6 +372,16 @@ func DoExecDefault(ctx context.Context, name string, args *workshop.Execution) (
 }
 
 func (s *FakeWorkshopBackend) RemoveWorkshopStash(ctx context.Context, name string) error {
+	_, projectId, err := s.userProject(ctx)
+	if err != nil {
+		return err
+	}
+
+	wp := s.StashedWorkshops[projectId][workshop.StashNamePrefix+name]
+	if wp == nil {
```
```suggestion
	if s.StashedWorkshops[projectId][workshop.StashNamePrefix+name] == nil {
```

internal/workshop/fakebackend/backend.go
```diff
@@ -503,6 +557,35 @@ func (b *FakeWorkshopBackend) Download(ctx context.Context, base string, report
...
+	if wp, ok := s.Workshops[prj.ProjectId][name]; !ok {
+		return workshop.ErrWorkshopNotLaunched
+	} else {
+		wp.File = file
+	}
+
+	s.RestoreCalls = append(s.RestoreCalls, RestoreCall{Workshop: name, Snapid: snapid, File: file})
+	if s.SnapshotCallback != nil {
+		return s.SnapshotCallback(ctx, name, snapid)
+	}
```
```suggestion
	if s.RestoreCallback != nil {
		return s.RestoreCallback(ctx, name, snapid)
	}
```

internal/workshop/lxd/lxd_backend_stash.go
```diff
@@ -83,33 +84,55 @@ func (s *Backend) UnstashWorkshop(ctx context.Context, name string) error {
...
-	// the new instance creation; hence, the prefix.
-	if op, err := conn.MigrateInstance(instanceFrom, api.InstancePost{
-		Name:      instanceTo,
-		Project:   target,
-		Migration: true,
-		Profiles:  []string{"default"},
-	}); err != nil {
+
+	dest := conn
+	dest = dest.UseProject(targetProject)
```
jonathan-conder:
```suggestion
	dest := conn.UseProject(targetProject)
```

internal/workshop/lxd/lxd_backend_stash.go
```diff
@@ -83,33 +84,55 @@ func (s *Backend) UnstashWorkshop(ctx context.Context, name string) error {
...
+		return err
+	}
+	if err = op.Wait(); err != nil {
+		return err
+	}
+
+	rev := revert.New()
+	defer rev.Fail()
+
+	rev.Add(func() {
```
jonathan-conder: There's no need for `revert` here since you never call `rev.Success()`.

internal/workshop/workshop.go
```diff
@@ -79,36 +82,29 @@ func (w *Workshop) LinkSdk(ctx context.Context, s sdk.Setup) error {
...
-	// We do not track the system SDK in the Sdks field; this is only for the
-	// user-defined SDKs as the system SDK is a special case (e.g. cannot be
-	// removed from the workshop).
-	if s.Name != sdk.System.String() {
-		now := InstallTimeNow()
-		s.InstallTime = &now
-
-		sdksetup, exist := w.Sdks[s.Name]
-		if exist {
-			sdksetup.RevisionSequence = append(sdksetup.RevisionSequence, sdksetup.Revision)
```
jonathan-conder, line 91: `RevisionSequence` appears to be unused apart from tests now

internal/workshop/workshop.go
```diff
@@ -294,23 +268,80 @@ func (w *Workshop) SdkInfo(ctx context.Context, sdkName string) (*sdk.Info, erro
...
+		return orderMap[infos[i].Name] < orderMap[infos[j].Name]
+	})
+
+	// The workshop definition which defines the install order may or may not
+	// contain the system SDK declared in an arbitrary place. Therefore, we have
+	// to make sure that the system SDK goes first.
+	infos = ensureFirst(infos, func(i *sdk.Info) bool { return sdk.IsSystem(i.Name) })
+
+	infos = ensureLast(infos, func(i *sdk.Info) bool { return sdk.IsSketch(i.Name) })
+
```
jonathan-conder: might be easier to put `orderMap["system"] = -1` and `orderMap["sketch"] = len(w.File.Sdks)`

internal/workshop/workshop.go
```diff
@@ -409,3 +440,7 @@ func (w *Workshop) InstallLocalSdk(ctx context.Context, name string, rev string,
 	reverter.Success()
 	return nil
 }
+
+func SnapshotId(w, sk string) string {
+	return strings.Join([]string{w, sk}, "-")
```
jonathan-conder: An SDK `foo-foo/bar` would clash with `foo/foo-bar`. It looks like the only banned characters are `/?&+`: https://github.com/canonical/lxd/blob/186dbf22220e82c0b456239833f16f1aff8816e5/shared/validate/validate.go#L523. Maybe also anything which can't appear in an instance name.

But also the instance name seems to become part of the snapshot name, so maybe the SDK name is sufficient? https://github.com/canonical/lxd/blob/186dbf22220e82c0b456239833f16f1aff8816e5/lxd/instance/drivers/driver_common.go#L707

tests/main/refresh/task.yaml
```diff
@@ -38,3 +38,45 @@ execute: |
...
+  pid=$(cat .workshop.lock)
+  expected_output=$(cat <<EOF
+  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-system
+  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-basic
+  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-mount
+  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-gpu
+  EOF
+  )
+  output=$(zfs list -t snapshot -s creation -o name | grep ws-refresh-snapshots | sed 's|^.*workshop/containers/workshop.ubuntu_||')
+  diff <(echo "$expected_output") <(echo "$output") > /dev/null
```
jonathan-conder:
```suggestion
  expected_output() {
    cat <<EOF
  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-system
  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-basic
  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-mount
  ws-refresh-snapshots-$pid@snapshot-ws-refresh-snapshots-test-sdk-gpu
  EOF
  }
  output() {
    zfs list -t snapshot -s creation -o name | grep ws-refresh-snapshots | sed 's|^.*workshop/containers/workshop.ubuntu_||'
   }
  diff <(expected_output) <(output) > /dev/null
```

internal/daemon/api_workshops_test.go
```diff
@@ -819,13 +855,17 @@ func (s *apiSuite) runActionTest(c *check.C, buffers []*bytes.Buffer, expected [
...
-			st.Lock()
-			if expected[num].ChangeErr != "" {
-				c.Check(change.Err(), check.ErrorMatches, expected[num].ChangeErr, check.Commentf("case: %v", num))
+
+			err := change.Err()
+			if err != nil {
+				c.Check(err, check.ErrorMatches, expected[num].ChangeErr, check.Commentf("case: %v", num))
 			} else {
-				c.Assert(change.Err(), check.IsNil)
+				c.Assert(expected[num].ChangeErr, check.Equals, "")
```
```suggestion
				c.Check(expected[num].ChangeErr, check.Equals, "")
```
Otherwise the state might not unlock

internal/workshop/workshop.go
```diff
@@ -294,23 +268,80 @@ func (w *Workshop) SdkInfo(ctx context.Context, sdkName string) (*sdk.Info, erro
...
+	// contain the system SDK declared in an arbitrary place. Therefore, we have
+	// to make sure that the system SDK goes first.
+	sdks = ensureFirst(sdks, func(i sdk.Setup) bool { return sdk.IsSystem(i.Name) })
+
+	sdks = ensureLast(sdks, func(i sdk.Setup) bool { return sdk.IsSketch(i.Name) })
+	return sdks
+}
+
 // Mounts returns a map of active mounts,
 // given a map of SDK info as returned by SdkInfos.
```
jonathan-conder:
```suggestion
// given a map of SDK info as returned by SdkInfosByInstallOrder.
```

jonathan-conder (changes requested):
Looks like a nice step forward. Think I found a couple of bugs but happy to be proven wrong

internal/daemon/api_workshops.go
jonathan-conder: However it also slows down `workshop list` quite a bit. For example:
```bash
mkdir -p .workshop
for i in {0..9}; do
	cat >".workshop/w${i}.yaml" <<EOF
name: w${i}
base: ubuntu@22.04
sdks:
  - name: system
    plugs:
      auto-connect-by-name:
        interface: tunnel
        endpoint: 1111${i}
      http:
        interface: tunnel
        endpoint: 2222${i}
  - name: test-sdk-basic
    channel: latest/stable
    plugs:
      jobs:
        interface: mount
        workshop-target: /mnt/wp-plug
  - name: test-sdk-camera
    channel: latest/stable
  - name: test-sdk-desktop
    channel: latest/stable
  - name: test-sdk-gpu
    channel: latest/stable
  - name: test-sdk-mount
    channel: latest/stable
  - name: test-sdk-ssh-agent
    channel: latest/stable
  - name: test-sdk-tunnel
    channel: latest/stable
connections:
  - plug: test-sdk-basic:jobs
    slot: test-sdk-mount:jobs
EOF
	workshop launch "w${i}"
	workshop connect "w${i}/test-sdk-camera:camera"
	workshop connect "w${i}/test-sdk-desktop:desktop"
	workshop connect "w${i}/test-sdk-ssh-agent:ssh-agent"
done
```

With this setup `workshop list` takes around `1.2s`. Removing the SDKs drops it to `0.3s`.

internal/overlord/workshopstate/handlers.go
dmitry-lyfar: This is moved because rebuilding can leave an existing instance still running, it needs to be deleted. I moved the clean up logic up to be executed on failure regardless as I found it simpler than trying to differentiate what failed (launch or rebuild) and decide whether you need to remove the workshop or it does not exist there. The clean up logic is something I want to be as quick and simple as possible.

jonathan-conder: yeah this makes more sense now that I've seen the whole PR.

Just noticed the log message should say `doConstructWorkshop` though

internal/workshop/backend.go
dmitry-lyfar: It copies over the default config and devices to the new instance's configuration, thus, overwriting the old devices and configuration set by the workshop for things like `user.workshop.sdks`. It just cannot set those directly because there is no way of setting an individual instance's config key, for example, except for using `UpdateInstance` that accepts `InstancePut`. I hope it will be clearer in the updated version of the implementation.

internal/workshop/lxd/lxd_backend_stash.go
dmitry-lyfar: It looks like we don't need this workaround anymore as we disconnect all the workshop's SDKs and remove profiles that were created by them *before* we attempt to stash. It used to be not true.

internal/overlord/workshopstate/handlers.go
jonathan-conder: Actually I think it would be nicer to move this into both `LaunchOrRebuildWorkshop` and `Restore`. Since both can leave the instance in an intermediate state, there's no way for clients to recover from an error except by removing the workshop if it still exists.

dmitry-lyfar: It's not that we don't have the guarantee. The workshop must be removed if the logic down the line failed (even though the workshop was successfully created). So, even if `Restore` does additional work to clean that up, it won't rid us of the logic here.

Also, I think it **is** transactional.

jonathan-conder: Right, I shouldn've said we should move the revert into `LaunchOrRebuildWorkshop` and `Restore`, and also move it after a successful call to those functions.

For me at least, if we follow this pattern consistently it reduces cognitive overhead. On the other hand it makes the code more complex. And future backends will have to implement their methods more carefully. So I'll leave the decision to you.

What I mean by not transactional, is that in `Restore` for example, it's possible to restore the snapshot but fail to restore the instance config.

internal/workshop/backend.go
jonathan-conder: It's clear at the moment, I just saw the config copy in `Restore`, and confirmation bias kicked in when I looked at the implementation of `LaunchOrRebuildWorkshop` to check.

internal/overlord/workshopstate/request.go
dmitry-lyfar: This is actually a bug!
```
SDK-1
SDK-2 

SDK-3  
SDK-2  # this snapshot will be chosen
SDK-1
```

internal/overlord/workshopstate/request.go
jonathan-conder: for this one I just meant it can be one line, like it was before

internal/daemon/api_workshops.go
dmitry-lyfar: I will leave this one as is for now.

internal/overlord/sdkstate/handlers.go
dmitry-lyfar: It's not if `err` is defined in the scope of `if`. I tend to using different names for errors in a revert logic (or defers per se) to avoid silly mistakes when it's `:=` instead of `=` that won't be caught by the linter / compiler.

internal/overlord/workshopstate/request.go
dmitry-lyfar: I will make it as part of the loop, no need to check it here anymore.

internal/overlord/workshopstate/request.go
dmitry-lyfar: Decided to keep it serial for now.

internal/overlord/workshopstate/request.go
dmitry-lyfar: Must be addressed in a separate PR.

internal/overlord/hookstate/handlers.go
dmitry-lyfar: Probably, not so easy anymore. The snapshot logic wants to be executed regardless of whether the hook exists or not and right after the `setup-base`. So, it'll take some restructure of this code to accommodate it and the benefit gained is not so obvious (esp. in this scope).

internal/workshop/backend.go
dmitry-lyfar: Decided to leave as is. The names should be shorten for the interface where a workshop is an obvious object of the operation.

internal/workshop/workshop.go
dmitry-lyfar: '.' was selected as the separator.

internal/daemon/api_workshops.go
```diff
@@ -119,20 +119,60 @@ func workshopFileToInfo(pid string, name string, path string) *WorkshopFileInfo
 	return &ws
 }
 
-func workshopToInfo(w *workshop.Workshop, sdks map[string]*sdk.Info, health healthstate.HealthState) *WorkshopInfo {
+// Returns essential workshop properties and its SDKs health statuses (if
+// available).
+func workshopToInfo(w *workshop.Workshop, health healthstate.HealthState) (*WorkshopInfo, error) {
```
jonathan-conder: I think the `error` return is unused

internal/overlord/workshopstate/request.go
```diff
@@ -337,168 +422,172 @@ func maybeSketch(ctx context.Context, pid, wp string) (bool, error) {
...
+}
+
+func definitionChanged(old, new *workshop.File, sdkName string) bool {
+	byName := func(s workshop.SdkRecord) bool { return s.Name == sdkName }
+	oldidx := slices.IndexFunc(old.Sdks, byName)
+	newidx := slices.IndexFunc(new.Sdks, byName)
+
+	if oldidx == -1 || newidx == -1 {
+		return false
+	}
```
jonathan-conder: Seems like this won't cause a refresh if you add or remove the system SDK from the workshop definition.

jonathan-conder (commented):
Just a couple of things this time

internal/daemon/api_workshops.go
dmitry-lyfar: That's one of my "on the fence" issues (there are some similar examples in the repo.go too) is whether you return an error from a function that cannot technically raise it but you want to remain consistent with other same-family functions.

jonathan-conder: In this case the function parameters don't match IIRC, but fair point otherwise.

jonathan-conder (approved):
Think that's everything!

PR #340, opened 2025-03-14
Detect if LXD is installed when getting ConnectionError

jy5275:
# Description

Current error message for LXD not installed is hard for user to understand and figure out the problem.

Workshopd should check if LXD socket is available and report a user-friendly message if LXD socket does not exist.

https://warthogs.atlassian.net/jira/software/c/projects/WSP/boards/1645?selectedIssue=WSP-570

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


jy5275:
@dmitry-lyfar  I think it's difficult to achieve the goal.

There's a simple way to check if LXD is installed, by calling `lxd.ConnectLXDUnixWithContext`.

But as you mentioned before, workshop client shouldn't be backend aware. Maybe we'll provide other backend selections besides LXD in the future. Therefore, client code can't talk to LXD socket directly if we stick to this rule.

The workshop client only knows that it cannot connect to workshop socket. Many possible reasons may lead to that:
- LXD not installed
- Zfs module not loaded to kernel (on WSL)
- socket file was mistakenly removed
- ...

But checking the error logs by `snap logs workshop` requires root permission. Normal workshop command wouldn't be able to do that. 

So I think the only thing we can do is to enrich the ConnectionError message to instruct the user to check error logs manually. Do you think we can do something more?

dmitry-lyfar:
See:

```
func (s *Backend) LxdClient(ctx context.Context) (lxd.InstanceServer, error) {
	user, ok := ctx.Value(workshop.ContextUser).(string)
	if !ok {
		return nil, fmt.Errorf("context key %s not found", workshop.ContextUser)
	}

	if srv, err := lxd.ConnectLXDUnixWithContext(ctx, LxdSock, nil); err != nil {
		return nil, err
	} else {
		if err = InitLxdProject(srv, user); err != nil {
			return nil, err
		}
		return srv.UseProject(LxdProjectName(user)), nil
	}
}
```

This is where the backend initiates the connection and where we need to understand if the socket exists (I think if you uninstall LXD and try to run this code, it will reply with a stat error about socket not existing). This is what we should intercept here and assume that LXD may not be installed in the error message. It will then make its way to the client. Client should not concern itself about the LXD backend.

However, the client error message should suggest a proper command to view workshop logs on a ConnectionError, e.g. `sudo snap logs workshop`.  Or, better, suggest a snap-agnostic option: `journalctl -u <name>.service`. You need to figure our a proper `<name>` part.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -83,6 +83,12 @@ func ImageAlias(name string) string {
 	return fmt.Sprintf("workshop-%s-%s", name, runtime.GOARCH)
 }
 
+func ErrorWithInstallLXDPrompt(err error) error {
+	return fmt.Errorf("cannot connect to LXD: %w, maybe LXD isn't installed?\n"+
+		"Install and initialize LXD by:\n  sudo snap install lxd\n  lxd init --auto\n"+
+		"Then restart workshop daemon by:\n  sudo snap start workshop.workshopd", err)
```
akcano:
```suggestion
		"Install and initialize LXD:\n  sudo snap install lxd\n  lxd init --auto\n"+
		"Then restart the workshop daemon:\n  sudo snap start workshop.workshopd", err)
```

akcano (approved):
One minor change.

internal/interfaces/lxd_device/backend.go
```diff
@@ -50,7 +50,7 @@ func lxdClient(ctx context.Context) (lxd.InstanceServer, error) {
 
 	srv, err := lxd.ConnectLXDUnixWithContext(ctx, LxdSock, nil)
 	if err != nil {
```
dmitry-lyfar, line 52: We should differentiate between errors that are due to LXD is not installed and LXD service is not available. Remove LXD and you'll get:

```
error: cannot create or load project at "/home/dmitry/work/workshop/refresh-revisions": Get "http://unix.socket/1.0": dial unix /var/snap/lxd/common/lxd/unix.socket: connect: no such file or directory
```

Stop the LXD service and you'll get:

```
error: cannot create or load project at "/home/dmitry/work/workshop/refresh-revisions": Get "http://unix.socket/1.0": dial unix /var/snap/lxd/common/lxd/unix.socket: connect: connection refused
```

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: @jy5275 I don't think we should recommend restarting the workshop daemon here. We simply cannot connect  to the LXD daemon, restarting would not make it different. See my comment above about error messages.

client/client.go
```diff
@@ -224,7 +224,7 @@ type ConnectionError struct {
 }
 
 func (e ConnectionError) Error() string {
-	return fmt.Sprintf("cannot communicate with server: %v", e.error)
+	return fmt.Sprintf("cannot communicate with server: %v\nUse `journalctl -u snap.workshop.workshopd.service` to find out more", e.error)
```
dmitry-lyfar:
```suggestion
	return fmt.Sprintf("cannot communicate with workshopd: %v\nRun `journalctl -u snap.workshop.workshopd.service` to find out more", e.error)
```

internal/interfaces/lxd_device/backend.go
dmitry-lyfar: So, unwrap the error received and check whether it's `connection refused` (the daemon is likely not started) OR `no entry` (no unix socket, e.g. is likely not installed):

```
		var syscall *os.SyscallError
		if errors.As(err, &syscall) {
			/// ... differentiate between conn refused and no entry here.
		}
```

internal/workshop/lxd/lxd_backend.go
```diff
@@ -83,6 +83,12 @@ func ImageAlias(name string) string {
 	return fmt.Sprintf("workshop-%s-%s", name, runtime.GOARCH)
 }
 
+func ErrorWithInstallLXDPrompt(err error) error {
```
dmitry-lyfar, line 88: This function does not have to be exported.

dmitry-lyfar (changes requested):

internal/workshop/lxd/lxd_backend.go
jy5275: It has to be exported because it's used in another package `lxd_device` https://github.com/canonical/workshop/blob/lxd-detect-msg/internal/interfaces/lxd_device/backend.go#L53

internal/workshop/lxd/lxd_backend.go
```diff
@@ -84,6 +85,21 @@ func ImageAlias(name string) string {
...
+	switch {
+	case errors.Is(err, unix.ECONNREFUSED):
+		return fmt.Errorf("cannot connect to LXD: %w, maybe LXD daemon isn't running?\n"+
+			"Start the LXD daemon:\n  sudo snap start lxd.daemon\n"+
```
```suggestion
			"Start the LXD daemon:\n  sudo snap start lxd\n"+
```
see `snap start` help.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -84,6 +85,21 @@ func ImageAlias(name string) string {
...
+			"Then restart the workshop daemon:\n  sudo snap start workshop.workshopd", err)
```
dmitry-lyfar: You are suggesting to restart, so `snap restart workshop`

internal/workshop/lxd/lxd_backend.go
```diff
@@ -84,6 +85,21 @@ func ImageAlias(name string) string {
...
+	case errors.Is(err, os.ErrNotExist):
+		return fmt.Errorf("cannot connect to LXD: %w, maybe LXD isn't installed?\n"+
+			"Install and initialize LXD:\n  sudo snap install lxd\n  lxd init --auto\n"+
```
dmitry-lyfar: Can we have the same prompt added to Backend.New, instead of just returning ` errors.New("LXD not initialized")` ? cc @jonathan-conder

internal/workshop/lxd/lxd_backend.go
```diff
@@ -84,6 +85,21 @@ func ImageAlias(name string) string {
...
+			"Then restart the workshop daemon:\n  sudo snap start workshop.workshopd", err)
```
dmitry-lyfar: `snap start workshop` or `snap restart workshop`.

jy5275:
@dmitry-lyfar I've committed all your suggested changes

tests/main/refresh/health.exp
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/expect -f
```
dmitry-lyfar: Why do we have this file added here?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -121,7 +137,7 @@ func New() (*Backend, error) {
 	// and non-Workshop LXD containers won't have network access
 	// without further manual configuration.
 	if len(networks) == 0 {
-		return nil, errors.New("LXD not initialized")
+		return nil, errors.New("LXD not initialized\nInitialize LXD:\n  lxd init --auto")
```
```suggestion
		return nil, errors.New("LXD not initialized. Run 'lxd init --auto'")
```

client/client.go
```diff
@@ -224,7 +224,7 @@ type ConnectionError struct {
 }
 
 func (e ConnectionError) Error() string {
-	return fmt.Sprintf("cannot communicate with server: %v", e.error)
+	return fmt.Sprintf("cannot communicate with workshopd: %v\nRun `journalctl -u snap.workshop.workshopd.service` to find out more", e.error)
```
dmitry-lyfar: What the actual message is going to look like for the user?

I've added you to `sdk-drafts` repository. Please add all the error messages that are possible from this PR as they are going to look like for the end user (error-messages/errors.md). See the format of the existing file to stick to. I think there are some inconsistencies with other error messages.

dmitry-lyfar: If you check the actual error messages that the user will see, you'll notice repetitions and that they are hard to read.

jy5275: The actual message is as the screenshot here https://warthogs.atlassian.net/jira/software/c/projects/WSP/boards/1645?selectedIssue=WSP-570

I don't see the "repetitions". Do you mean the caller functions of `ConnectionError` will add similar message when wrapping that error?

tests/main/refresh/health.exp
jy5275: This is a mistake. This file is for another PR (--no-wait option for refresh). 
I removed it and checked other changes of that PR are not mistakenly introduced.

client/client.go
jy5275: I've update all LXD related error messages to be consistent with other error messages.

Added them here https://github.com/dmitry-lyfar/sdk-drafts/blob/main/error-messages/errors.md

dmitry-lyfar (approved):

PR #341, opened 2025-03-14
Doc: Workshop internals reference

akcano:


docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+(:command:`exec`, :command:`run`, :command:`shell`)
+run login shells as :samp:`uid=1000` and :samp:`gid=1000`,
+unless you explicitly change the IDs.
+This ensures files created inside the container map back to the host user.
+
+By default, :envvar:`$XDG_RUNTIME_DIR` is set to :samp:`/run/user/1000/`,
+and a user-scoped session bus address is created at :samp:`/run/user/1000/bus`.
+Using other IDs can break sessions.
+
+Finally, it's worth mentioning
```
dmitry-lyfar: This has changed: if a user has an active session, it won't prevent `refresh` or any other change from running. E.g., connecting an interface will work but raise a warning if you have an active session as some interfaces affect the environment and may require the existing sessions to restart to take effect.

docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+tailored to a specific project.
+
+
+How workshops are updated
+-------------------------
+
+.. @artefact workshop refresh
+
+|ws_markup| has a refresh manager that tracks all updates in a workshop
+(changes to the base image, adding or removing SDKs,updates to the definition)
```
dmitry-lyfar:
```suggestion
(changes to the base image, adding or removing SDKs, updates to the definition)
```

docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+and builds a refresh plan to decide
+if a simple refresh is enough or if a stash is needed:
+
+Minor changes can be updated in place with a simple rollback mechanism.
```
dmitry-lyfar: We only have a snapshot mechanism that will or will not be used depending on the changes (i.e. there is no rollback mechanism in the sense described here). E.g. if the base has been changed, the workshop will be rebuilt, otherwise, it is always some kind of restoration from a snapshot.

There is an exception from the above: scripts. These do not require any refresh.

docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+
+ZFS storage and pool size
+-------------------------
+
+|ws_markup| stores its containers and data on a ZFS pool by default.
+This approach consolidates container images, :program:`apt` caches, SDKs
+and other workshop content under a single system.
+
+If you need more space or different performance,
+you can grow or tune the ZFS pool by means default to ZFS itself.
```
dmitry-lyfar: This needs to be done as documented here: https://documentation.ubuntu.com/lxd/en/latest/howto/storage_pools/#resize-a-storage-pool. The ZFS pools that are managed by LXD are not recommended to be changed by using zfs tools directly.

docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+Day-to-day usage usually requires no manual interaction with pool management.
+
+Also, |ws_markup| makes sure the storage pool is at least 5GiB.
+If it's smaller, it can run out of space if under 14GiB of space is available.
```
dmitry-lyfar: I think there is a dot in the middle of the sentence, not sure how you planned to structure it.

docs/reference/workshop-internals.rst
```diff
@@ -0,0 +1,177 @@
...
+Interfaces
+----------
+
+.. @artefact interface
+
+As a reminder, |ws_markup| enforces resource access with plugs and slots.
+A plug requests something (for example, GPU or certain sockets),
+and a slot provides it.
+
+However, only the user mapped inside the workshop
```
dmitry-lyfar: Hmm, maybe I'm missing something. That looks like the other way around, only the `workshop` user can access host resources enabled by interfaces from within the workshop. Other users that may exist in the workshop will not have access. As for other **host** users have access to the host resources per definition.

dmitry-lyfar (changes requested):

docs/reference/workshop-internals.rst
dmitry-lyfar: The ZFS pool's name is `workshop`.

dmitry-lyfar (approved):
Looks good! There is a fixup comment, so I'll wait for it to be auto-squashed before merging.

PR #345, opened 2025-03-20
Doc: Expose system SDK

akcano:


docs/reference/definition-files/workshop-definition.rst
```diff
@@ -150,15 +152,30 @@ System SDK
...
+
+The system SDK declares slots for interfaces that only support a single slot,
+because they expose host-based resources that are singular by nature:
+:samp:`system:camera`, :samp:`system:desktop`, :samp:`system:gpu`,
+:samp:`system:mount`, and :samp:`system:ssh-agent`.
+No other SDKs can declare slots for these interfaces, except for :samp:`mount`.
+The :samp:`system:mount` slot is still unique
+because it's the only one that provides access to the host file system,
+whereas slots under regular SDKs only expose locations in the workshop.
+However, regular SDKs can also declare slots for the mount interface.
```
dmitry-lyfar:
```suggestion
```

dmitry-lyfar: That appears to be explained in the previous two sentences.

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -150,15 +152,30 @@ System SDK
...
+
+Additional plugs and slots for interfaces like :samp:`tunnel` or :samp:`mount`
+are subject to extra conditions and restrictions
```
dmitry-lyfar: I would say they are not `auto-connectable` during `launch` or `refresh`. That's probably the only restriction that matters here.

docs/tutorial/index.rst
```diff
@@ -665,6 +679,94 @@ to a new location on the host:
...
+by adding a list item to the :samp:`connections` list;
+
+Refresh the workshop to enable the tunnel;
+the tunnel interface validates and enables the connection.
+Then start the server as root at the default port 80 (the slot):
+
+.. code-block:: console
+
+   $ workshop refresh
+   $ workshop exec --uid 0 -- /home/workshop/go/bin/caddy file-server
```
dmitry-lyfar: No need to use the full path in the latest versions of the Go SDK as it adds ../go/bin to `.profile` instead of `bashrc`

docs/tutorial/index.rst
```diff
@@ -665,6 +679,94 @@ to a new location on the host:
...
+to connect to a server running in the workshop.
+
+.. @artefact tunnel interface
+
+For a quick demo, let's install `Caddy <https://caddyserver.com/>`_
+to serve files over HTTP:
+
+.. code-block:: console
+
+   $ workshop exec -- go install github.com/caddyserver/caddy/v2/cmd/caddy@latest
```
dmitry-lyfar, line 697: We need a tutorial test update as well. I've just realised there was no `noble/stable` in the store, only `noble/edge` for the `go` SDK (updated).

docs/tutorial/index.rst
dmitry-lyfar: This is not going to work as you've refreshed the workshop. `go` SDK does not have a plug for `~/go/bin` for now. I will update it and upload shortly, you should be able to do that here and in the tests for `noble/stable`.

docs/tutorial/index.rst
```diff
@@ -665,6 +679,94 @@ to a new location on the host:
...
+           interface: tunnel
+           endpoint: localhost:80
+     - name: system
+       plugs:
+         server:
+           interface: tunnel
+           endpoint: 8080
+   connections:
+     - plug: system:server
+       slot: go:server
```
dmitry-lyfar:
```
name: goenv
base: ubuntu@24.04
sdks:
  - name: go
    channel: noble/stable
    slots:
      caddy:
        interface: tunnel
        endpoint: localhost:80
  - name: system
    plugs:
      caddy:
        interface: tunnel
        endpoint: 8080
```

1. The plug / slot name should reflect the service name.
2. If the plug and the slot name are equal AND the service is exposed workshop -> host, it will be connected automatically, so no need for `connections`.

dmitry-lyfar (changes requested):
That looks nice, some issues with the tutorial/interfaces.

docs/tutorial/index.rst
dmitry-lyfar: The SDK is updated.

akcano: It does work, although the workaround for `go/bin` sure looks nasty and will be improved (thanks for updating the SDK):

```
$ workshop exec --uid 0 -- /home/workshop/go/bin/caddy file-server
2025/03/25 09:29:50.311 WARN    admin   admin endpoint disabled
2025/03/25 09:29:50.312 WARN    http.auto_https server is listening only on the HTTP port, so no automatic HTTPS will be applied to this server       {"server_name": "static", "http_port": 80}
2025/03/25 09:29:50.312 INFO    tls.cache.maintenance   started background certificate maintenance      {"cache": "0xc000734b80"}
2025/03/25 09:29:50.312 WARN    http    HTTP/2 skipped because it requires TLS  {"network": "tcp", "addr": ":80"}
2025/03/25 09:29:50.312 WARN    http    HTTP/3 skipped because it requires TLS  {"network": "tcp", "addr": ":80"}
2025/03/25 09:29:50.312 INFO    http.log        server running  {"name": "static", "protocols": ["h1", "h2", "h3"]}
2025/03/25 09:29:50.312 INFO    Caddy serving static files on :80
2025/03/25 09:29:50.334 INFO    tls     cleaning storage unit   {"storage": "FileStorage:/root/.local/share/caddy"}
2025/03/25 09:29:50.340 INFO    tls     finished cleaning storage units
```

docs/tutorial/index.rst
akcano: I wanted to update the tutorial test after we've settled on the command sequence in the tutorial.

docs/tutorial/index.rst
dmitry-lyfar: `go/bin` as a plug of the SDK should be OK-ish to mount on the host in this particular SDK and this is why it works now with the updated SDK. Previously, refresh would rebuild the entire workshop and, hence, remove `caddy`. Ideally, `caddy` should be part of an SDK which would preserve it over refreshes (a sketch SDK or a project-specific SDK).

dmitry-lyfar: We should also preserve `GOCACHE='/home/dmitry/.cache/go-build'` in the `go` SDK I suppose, just a note, not relevant to this tutorial.

tests/documentation/tutorial/task.yaml
```diff
@@ -55,6 +56,30 @@ execute: |
   workshop_exec remount nimble/test-sdk-go:mod-cache ~/mod
   workshop_exec info
 
+  # Caddy
+  workshop_exec exec -- go install github.com/caddyserver/caddy/v2/cmd/caddy@latest
+  cat <<EOF > Caddyfile
+  :8080 {
```
dmitry-lyfar:
```suggestion
  :8082 {
```

dmitry-lyfar (approved):
Looks good, I believe the tests fail because the ports are mixed up.

Also, what do you think of using the actual store for the documentation tests? We can merge this one as is, but I don't see a good reason for crafting a separate fake SDK in the e2e tests now.

akcano:
Using the real store would make the tests much more relevant, although there was a concern of stability AFAIR. If that's out of question, I'm all for it.

PR #351, opened 2025-03-23
Migrate camera interface to unix-hotplug devices

jonathan-conder:
# Description

Now that LXD 6.3 has been in `latest/stable` for a few days, this should be safe to merge. It bumps the minimum required LXD version to 6.3.

The new feature this uses is the `subsystem` config for `unix-hotplug` devices. This allows us to support `media` devices and more than a fixed number of `video4linux` devices. The ownership of these devices will match the host system.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


tests/lib/utils.sh
```diff
@@ -1,8 +1,7 @@
 #!/bin/bash
 
 function setup_lxd() {
-    snap install --classic lxd
-    snap refresh --channel=5.21/stable lxd
+    snap install --classic --channel=6/stable lxd
```
dmitry-lyfar: `snap install` was a no-op in most cases there afaik -- the image has already had LXD installed (5.0.1, hence the refresh).

dmitry-lyfar: So, your change would likely be a no-op too?

jonathan-conder: it says `snap "lxd" is already installed, see 'snap help refresh'`. So yes

jonathan-conder: I'll follow https://github.com/canonical/setup-lxd/blob/main/action.yml

dmitry-lyfar:
Let's also fix that found issue with snapd here so that everything else can rebase. I'm happy to merge otherwise.

dmitry-lyfar (approved):

jonathan-conder:
If I understand correctly, that's fixed here: https://github.com/canonical/workshop/pull/351/commits/54d83dedbcd167a4881a751c04493f4816c20343. But I need to rebase on main to resolve the conflict

docs/explanation/interfaces/camera-interface.rst
```diff
@@ -65,10 +65,11 @@ can be invoked manually after the workshop has started:
...
 
 Establishing a connection means
-that all currently connected USB-based cameras
+that all existing :samp:`video4linux` and :samp:`media` devices
 will be made available inside the workshop.
-New cameras can be added
-by disconnecting and then reconnecting the camera interface.
+While the connection is active,
+new devices on the host will be made available inside the workshop;
+unplugged devices will also be removed from the workshop.
```
akcano, line 72:
```suggestion
adding new devices on the host will also make them available inside the workshop,
whereas unplugged devices will also be removed from the workshop.
```

akcano (commented):
LGTM, one minor fix.

docs/explanation/interfaces/camera-interface.rst
jonathan-conder: Thanks, I'll address in a follow up. Dmitry was keen to merge this one to fix the actions

PR #355, opened 2025-03-25
Track revisions for sketch SDK and bind-mount them in the workshop

jonathan-conder:
# Description

In preparation for other types of local SDKs, revisions for sketch SDKs are now tracked on the host instead of inside the workshop. These revisions are immutable (without user misbehaviour) and are mounted read-only inside the workshop, similar to store SDKs. All SDKs are mounted at `/var/lib/workshop/sdk/<NAME>` with no revision numbers or `current` symlink.

If the user runs `sketch-sdk` and makes no changes to the YAML, the subsequent refresh will trigger, but won't do much work. Same with manual refreshes while a sketch SDK is installed.

Editing the sketch SDK is now more robust to potential concurrent edits, e.g. we don't assume the user's editor saves files atomically.

The sketch SDK template now includes `apt-get update` - when testing I forgot to put this in there a couple of times, and I suspect our users will too. It also doesn't have a `base` since this can be inferred from the workshop. Going forward, only store SDKs are likely to require a specific `base`.

The system SDK is now treated more like a store SDK, mounted from a volume (which is populated from the metadata embedded in `workshopd`). This will become important when we start allowing users to upgrade the workshop snap without removing their existing workshops. There's currently no cleanup of old system SDK "downloads", because it's not possible to recover old versions. All store SDKs could benefit from improved lifecycle management, so we can address this issue when we get to that.

Snapshots are now taken after each SDK is installed, so restoring from a snapshot doesn't require SDKs to be unmounted. Rebuilt workshops will also lose their SDKs, to be more consistent with newly launched workshops. They will retain the `/project` mount, but we plan to mount this later on and it won't be preserved during a rebuild. See https://github.com/canonical/workshop/pull/364 for more details.

Launch and refresh share more behaviour now, e.g. refresh only queries at the relevant projects/workshops and launch unlocks the state while reading workshop definition files and querying the store. I'll do some tests soon to confirm that this doesn't have a negative impact on latency. Launch will delete the sketch SDK if it finds one (left over from a snap reinstall).

Workshop will read local SDK info from the host filesystem instead of the workshop. I don't think this is necessary but it's more in line with how store SDKs work.

The tasks for installing SDKs are split into `install-sdk` and `register-sdk`. The former mounts the SDK and adds it to the instance config. The task is automatically undone when removing, restoring or rebuilding a workshop. On the other hand `register-sdk` updates the interface repo and needs to be undone before removing, restoring or rebuilding.

This breaks quite a few things, e.g. `workshop remove` will probably struggle to remove existing workshops after updating the daemon.

The documentation probably needs another pass, I'll take a look soon.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"os"
@@ -126,8 +125,6 @@ var (
 )
 
 func stashSketch(sketchdir, stashdir string) (*revert.Reverter, error) {
-	reverter := revert.New()
-
 	recs, err := os.ReadDir(sketchdir)
 	if err != nil && !osutil.IsDirNotExist(err) {
 		return nil, err
@@ -137,20 +134,43 @@ func stashSketch(sketchdir, stashdir string) (*revert.Reverter, error) {
 		return nil, fmt.Errorf(`cannot stash: the 'sketch' SDK doesn't exist`)
 	}
 
-	if err := os.MkdirAll(stashdir, 0755); err != nil {
+	// Ensure stashdir exists but is empty.
+	if err := clearStash(stashdir); err != nil {
 		return nil, err
 	}
+
+	reverter := revert.New()
+	defer reverter.Fail()
 
 	if err := osutil.Exchange(sketchdir, stashdir); err != nil {
 		return nil, err
 	}
 	reverter.Add(func() { _ = osutil.Exchange(stashdir, sketchdir) })
 
-	if err := os.RemoveAll(sketchdir); err != nil {
+	if err := os.Remove(sketchdir); err != nil {
 		return nil, err
 	}
 	reverter.Add(func() { _ = os.MkdirAll(sketchdir, 0755) })
-	return reverter, nil
+
+	clone := reverter.Clone()
+	reverter.Success()
+	return clone, nil
+}
+
+func clearStash(stashdir string) error {
+	temp, err := os.MkdirTemp(filepath.Dir(stashdir), "stash-*")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(temp)
+	if err := os.Chmod(temp, 0755); err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(stashdir, 0755); err != nil {
+		return err
+	}
+	return osutil.Exchange(stashdir, temp)
 }
 
 func restoreSketch(sketchdir, stashdir string) error {
@@ -185,7 +205,20 @@ func removeSketch(sketchdir string) error {
 		return fmt.Errorf(`cannot remove: the 'sketch' SDK doesn't exist`)
 	}
 
-	return os.RemoveAll(sketchdir)
+	temp, err := os.MkdirTemp(filepath.Dir(sketchdir), "sketch-*")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(temp)
+	if err := os.Chmod(temp, 0755); err != nil {
+		return err
+	}
+
+	if err := osutil.Exchange(sketchdir, temp); err != nil {
+		return err
+	}
+
+	return os.Remove(sketchdir)
 }
 
 func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
@@ -254,17 +287,17 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
 		cmdrefresh := &CmdRefresh{root: c.root}
 		cmdrefresh.WaitOnError = true
 
-		storedir := workshop.SketchSdkStash(userDataDir, p.Id, wp.Name)
+		stashdir := workshop.SketchSdkStash(userDataDir, p.Id, wp.Name)
 
-		if err = restoreSketch(sketchdir, storedir); err != nil {
+		if err = restoreSketch(sketchdir, stashdir); err != nil {
 			return err
 		}
 
-		// Run refresh with the stored sketch SDK. We do not revert dirs exchange
+		// Run refresh with the stashed sketch SDK. We do not revert dirs exchange
 		// on a failed refresh here as it is run with the content from "stored"
 		// and with --wait-on-error. Hence, there is always a possibility to
-		// workshop refresh --abort and workshop sketch-sdk --restore to restore the
-		// original sketch content.
+		// workshop refresh --abort and workshop sketch-sdk --stash to restore the
+		// original stash content.
 		return cmdrefresh.Run(cmd, []string{fmt.Sprintf("%s/sketch", wp.Name)})
 	}
 
@@ -277,38 +310,8 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
 		return cmdrefresh.Run(cmd, []string{wp.Name})
 	}
 
-	metafile := filepath.Join(sketchdir, "meta", "sdk.yaml")
-
-	// Format sketch SDK template header.
-	boilerplate := fmt.Sprintf(sketchTemplate, wp.Path)
-
-	if osutil.FileExists(metafile) {
-		old, err := os.ReadFile(metafile)
-		if err != nil {
-			return err
-		}
-
-		new, err := runTextEditor(metafile, []byte{})
-		if err != nil {
-			return err
-		}
-
-		if bytes.Equal(old, new) {
-			return nil
-		}
-
-		if err = writeSketchSdk(sketchdir, new); err != nil {
-			return err
-		}
-	} else {
-		res, err := runTextEditor("", []byte(boilerplate))
-		if err != nil {
-			return err
-		}
-
-		if err = writeSketchSdk(sketchdir, res); err != nil {
-			return err
-		}
+	if err = editSketchSdk(sketchdir, wp.Path); err != nil {
+		return err
 	}
 
 	cmdrefresh := &CmdRefresh{root: c.root}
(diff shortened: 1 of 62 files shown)


akcano:
Please note the changes to the tutorial intro'd in https://github.com/canonical/workshop/pull/365

docs/tutorial/index.rst
```diff
@@ -369,29 +369,54 @@ pass the ID of the change to the :ref:`tasks <ref_workshop_tasks>` command:
...
      ID   Status  Spawn               Ready               Summary
-     133  Done    today at 11:32 GMT  today at 11:32 GMT  Create new "dev" workshop
-     134  Done    today at 11:32 GMT  today at 11:32 GMT  Mount project directory "hello-workshop"
-     135  Done    today at 11:32 GMT  today at 11:32 GMT  Start "dev" workshop
-     136  Done    today at 11:32 GMT  today at 11:32 GMT  Retrieve "go" SDK from channel "latest/stable"
-     137  Done    today at 11:32 GMT  today at 11:32 GMT  Install "go" SDK
-     138  Done    today at 11:32 GMT  today at 11:33 GMT  Link "go" SDK
-     139  Done    today at 11:33 GMT  today at 11:33 GMT  Run hook "setup-base" for "go" SDK
-     140  Done    today at 11:33 GMT  today at 11:33 GMT  Auto-connect interfaces of "go" SDK
+     132  Done    today at 11:32 GMT  today at 11:32 GMT  Retrieve "system" SDK
```
dmitry-lyfar, line 334: @akcano is there a way to shrink this growing monster in the documentation (esp. in the tutorial)?

jonathan-conder: I defaulted to just putting the actual output in there. But we can probably get away with removing some of it

dmitry-lyfar: Yes, it's interesting whether there are possible representations that one can expand if required, but, at the same time, see the gist.

dmitry-lyfar: It's not that bad rendered on my screen, though.

akcano: Apart from dropping some lines, we have the `dropdown` directive:

https://sphinx-design.readthedocs.io/en/latest/dropdowns.html

For an example in our docs, see [here](https://github.com/canonical/workshop/blob/f90184283f0910cb0ff904f29e991465887343e2/docs/reference/definition-files/sdk-definition.rst#L144).

jonathan-conder: I moved the full example out of the tutorial and into the debugging HOWTO, let me know what you think

jonathan-conder:
Docs are now updated. I cherry-picked the tutorial changes, will rebase when we're closer to merging.

When updating the tasks output I noticed that `check-health` only runs for the SDKs that were refreshed. I think it makes more sense to run it for all SDKs even if they weren't refreshed.

Fixed a couple of bugs I noticed with local SDKs. This won't affect sketch because the `~/.local/share/...` directory already exists. But it will affect project-defined SDKs.

I hard-coded a local `go` SDK to trigger a setup-base failure for the tasks output, which seems to be working. Also tested the `sketch-sdk` scenarios from `sdk-drafts`.

cmd/workshop/sketch.go
```diff
@@ -185,7 +205,20 @@ func removeSketch(sketchdir string) error {
 		return fmt.Errorf(`cannot remove: the 'sketch' SDK doesn't exist`)
 	}
 
-	return os.RemoveAll(sketchdir)
+	temp, err := os.MkdirTemp(filepath.Dir(sketchdir), "sketch-*")
```
dmitry-lyfar, line 208: It looks like this logic can be merged with `clearStash`? We don't have to have separate prefixes for these temp dirs.

internal/overlord/ifacestate/ifacemgr_test.go
```diff
@@ -103,8 +94,9 @@ func (s *interfaceManagerSuite) mockSdk(c *check.C, name, sdkYaml string, rev sd
 	s.state.Lock()
 	be := s.o.WorkshopBackend()
 	s.state.Unlock()
-	err = be.ImportVolume(s.ctx, sdk.VolumeName(name, rev), vfs)
-	c.Assert(err, check.IsNil)
+	if err := be.ImportVolume(s.ctx, sdk.VolumeName(name, rev), vfs); err != nil {
+		c.Assert(err, testutil.ErrorIs, workshop.ErrVolumeAlreadyExists)
```
dmitry-lyfar, line 98: Why can we have `already exists` in those cases and is it safe to ignore them here?

akcano (approved):

internal/sdk/local.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	for _, revision := range revisions {
+		dir := filepath.Join(target, revision.String())
+		// FIXME: should drop privileges for this. There's a (small) chance that
+		// unauthorised users can brute force the source directory contents.
+		if osutil.DirsAreEqual(source, dir) {
+			// Extend revision lifetime.
+			if err := os.Chtimes(dir, time.Time{}, time.Now()); err != nil {
+				logger.Noticef("TODO")
```
dmitry-lyfar: TODO.

internal/sdk/local.go
```diff
@@ -0,0 +1,124 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/canonical/workshop/internal/logger"
+	"github.com/canonical/workshop/internal/osutil"
+)
+
+// CommitRevision copies the source SDK into a subdirectory of the target, reusing an existing
+// subdirectory if possible. The subdirectory name is a local revision number, e.g. x123 for -123.
+// If no local revision matches the source, the new revision is one less than the lowest existing
+// revision, e.g. -124. Before creating a new revision, all but the two most recently installed
```
dmitry-lyfar:
```suggestion
// revision, e.g. -124. After creating a new revision, all but the two most recently installed
```

internal/sdk/local.go
```diff
@@ -0,0 +1,124 @@
...
+		return
+	}
+
+	times := make([]revTime, 0, len(revisions))
+	for _, revision := range revisions {
+		var modTime time.Time
+		info, err := os.Stat(filepath.Join(path, revision.String()))
+		if err != nil {
+			// Broken revision, use zero modTime so it is removed first.
+			logger.Noticef("TODO")
```
dmitry-lyfar: TODO

internal/sdk/local.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		times = append(times, revTime{revision, modTime})
+	}
+
+	slices.SortFunc(times, func(a, b revTime) int { return a.modTime.Compare(b.modTime) })
+	times = times[:remove]
+	for _, time := range times {
+		err := os.RemoveAll(filepath.Join(path, time.revision.String()))
+		if err != nil {
+			logger.Noticef("TODO")
```
dmitry-lyfar: TODO

internal/overlord/workshopstate/request.go
```diff
@@ -55,59 +55,97 @@ func (w *WorkshopManager) loadProject(ctx context.Context, id string) (workshop.
...
+
+func (w *WorkshopManager) resolveWorkshops(ctx context.Context, project workshop.Project, names []string, action string) ([]workshopReq, error) {
+	sto := sdk.StoreService(w.state)
+	reqs := make([]workshopReq, 0, len(names))
+
+	// Not an error, the state is locked; unlock it to let other requests to be
+	// processed while we are getting the store info sorted.
+	// This code can be concurrent with other changes,
+	// so we avoid interacting with local SDKs.
+	w.state.Unlock()
```
dmitry-lyfar, line 124: Why cannot `resolveWorkshops` also call `resolveLocalSdks` before the state is unlocked here? Thus, you will keep the resolution logic under the `resolveWorkshops` and keep local SDKs as a property of `workshopReq`. In the client code it will (pseudo):

```
		sdks := ordered(req.installOrder, req.storeSdks, req.localSdks)

		tasks := launch(w.state, req.file, sdks, project)
		taskset = append(taskset, tasks)
```

internal/overlord/workshopstate/request.go
```diff
@@ -144,34 +251,38 @@ func retrieveSdks(st *state.State, sdks []sdk.Setup) (*state.TaskSet, map[string
...
+			retrieveTask = retrieveTasks[setup.Name]
+			install = sdkstate.Install(st, setup.Name, retrieveTask)
 		}
+		addTask(install)
+
+		register := sdkstate.Register(st, setup.Name, retrieveTask)
+		addTask(register)
 
-		addTaskSet(install)
+		hook := hookstate.Hook(st, pid, w, setup.Name, 0, hookstate.SetupBase)
```
dmitry-lyfar, line 290: Perhaps, we should comment why these `setup-base` run in that particular order contrary to the state or health hooks.

internal/sdk/local.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return revisions, nil
+}
+
+func nextRevision(source, target string, revisions []Revision) (Revision, bool) {
+	if len(revisions) == 0 {
+		return R(-1), false
+	}
+
+	for _, revision := range revisions {
```
dmitry-lyfar, line 76: Interesting. I'm maybe missing something but why do we have to iterate over all the existing revisions and compare? Could we just compare with the recent revision and, if a difference found, create a new one + remove too old ones?

internal/workshop/device.go
```diff
@@ -97,3 +105,34 @@ func NewSdkProfile(sdkName string) SdkProfile {
 		Mounts: make(map[string]Mount),
 	}
 }
+
+func defaultDevices(pid, w string) ([]Mount, []ProxyEntry) {
```
dmitry-lyfar, line 109: Nice.

internal/overlord/ifacestate/ifacemgr_test.go
jonathan-conder: The sdkstate handler (`doRetrieveSdk`) already ignores it. But these tests never reused the same SDK until the system SDK started using volumes.

cmd/workshop/sketch.go
jonathan-conder: I think it could technically be merged, but the logic is different - clearStash is creating an empty directory, removing the existing contents safely. Whereas this code is just removing the existing contents.

internal/sdk/local.go
jonathan-conder: I should have caught this in my self-review :facepalm:

internal/overlord/workshopstate/request.go
jonathan-conder: At the moment we only check for conflicting changes after the state is locked again. I think we have to do it that way, although maybe we could also check before unlocking the state. Given that a refresh may be in progress, we don't know which revisions are currently in use (there could be more than one), so it's not safe to delete them.

internal/overlord/workshopstate/request.go
jonathan-conder: I was thinking the snapshotting should be a separate task, that would make it easier to explain I think

internal/sdk/local.go
jonathan-conder: That's something we talked about a while back. At the moment there's no benefit in checking all old revisions. But the extra cost is likely small. In future, it may allow us to restore a snapshot instead of rebuilding from scratch, which could save a lot of time.

internal/workshop/device.go
jonathan-conder: my thinking was there are 2 kinds of devices, those needed for LXD (root device, network) and those needed for any backend

jonathan-conder: but this might not scale very well since we have lots of different device structs at this level

internal/overlord/workshopstate/request.go
jonathan-conder: Come to think of it, I don't think checking the workshop's existence is enough to prevent concurrent launches. Because the workshop is created asynchronously.

```console
$ workshop launch & workshop launch
[1] 109030
- Create apt cache for "dev" (The storage volume is still in use)
- Create new "dev" workshop (Failed rebuilding instance from image: Error deleting storage volume: Failed to run: zfs destroy -r workshop/containers/workshop.jono_dev-703b7873: exit status 1 (cannot destroy 'workshop/containers/workshop.jono_dev-703b7873': dataset is busy))
"dev" launch aborted
```

docs/tutorial/index.rst
dmitry-lyfar: I think it's better. Maybe we should leave one more task there to demonstrate different kinds as there are only retrieve ones.

dmitry-lyfar (approved):
There is a tutorial conflict, but everything looks pretty good. I've tested various workflows with sketch SDK (create, stash, restore, remove) and the behaviour, as well as the directory structure and mounts configuration, works correct.

jonathan-conder:
> There is a tutorial conflict

I've rebased on main to resolve the conflict and added one more line to the tutorial `workshop tasks`. Will squash in the morning, I guess after our meeting

PR #364, opened 2025-04-04
Run setup-base hook and snapshot right after installing SDK

jonathan-conder:
# Description

Previously, Workshop made an effort to remove existing SDKs after
restoring from a snapshot. However, the removed SDKs are based on the
current refresh. Future refreshes won't know about these SDKs, but
volumes are still mounted in the snapshot.

Example:
```console
$ cat <<EOF >workshop.yaml
name: dev
base: ubuntu@22.04
sdks:
  - name: test-sdk-basic
    channel: latest/stable
EOF
$ workshop launch
"dev" launched
$ cat <<EOF >workshop.yaml
name: dev
base: ubuntu@22.04
EOF
$ workshop refresh
"dev" refreshed
$ workshop exec -- ls -l /var/lib/workshop/sdk/test-sdk-basic
total 1
lrwxrwxrwx 1 root root 40 Apr  4 04:25 current -> /var/lib/workshop/sdk/test-sdk-basic/696
$ workshop refresh
$ workshop exec -- ls -l /var/lib/workshop/sdk/test-sdk-basic
total 1
drwx--x--x 4 root root  4 Apr  4 04:24 696
lrwxrwxrwx 1 root root 40 Apr  4 04:25 current -> /var/lib/workshop/sdk/test-sdk-basic/696
```

This PR restores earlier behaviour, where setup-base is run after
installing the associated SDK, and before other SDKs are mounted.

Other SDKs no longer need to be removed from the workshop after
restoring a snapshot. This means `FakeWorkshopBackend.Restore` needs to
be a bit smarter at restoring snapshots. Rather than copying the
WorkshopFs on each snapshot, it's enough for now to just remove the SDKs
that were installed after the current snapshot.

Also, rebuilt workshops will no longer include any SDK mounts, in line with freshly launched workshops.

This might break things when refreshing a workshop from an old-style snapshot.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/daemon/api_workshops_test.go
+++ b/internal/daemon/api_workshops_test.go
@@ -17,7 +17,6 @@ import (
 	"gopkg.in/check.v1"
 	"gopkg.in/yaml.v3"
 
-	"github.com/canonical/workshop/internal/dirs"
 	"github.com/canonical/workshop/internal/interfaces"
 	"github.com/canonical/workshop/internal/osutil"
 	"github.com/canonical/workshop/internal/overlord/conflict"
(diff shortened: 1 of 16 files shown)


jonathan-conder:
I don't think there's much point reviewing this one separately, but I'm leaving it open for now. GitHub will mark it as merged when we merge https://github.com/canonical/workshop/pull/355

PR #365, opened 2025-04-07
Doc: Refactor tutorial with less links

akcano:


docs/tutorial/index.rst
```diff
@@ -27,11 +27,15 @@ refer to the
 :ref:`explanation <exp_index>` section.
 For comprehensive details, explore the
 :ref:`reference <ref_index>` section.
-Finally,
-if you're looking for advanced practical steps,
+If you're looking for advanced practical steps,
 see the
 :ref:`how-to guides <how_index>`.
 
+.. note::
```
dmitry-lyfar: After re-reading the tutorial, I think we should also drop everything that goes after 

```
sudo snap install --dangerous --classic ./workshop_0.1.13_amd64.snap
```
and before `Install shell completions`. Having a note about debugging workshops is enough in most cases. Workshop must be up and running after that `snap install` line OR be very informative about what's not working (e.g. LXD is not available). I'm fairly confident in both and would not occupy the precious tutorial space with double-checking anymore. Otherwise, they should go to the how-to.

docs/tutorial/index.rst
```diff
@@ -27,11 +27,15 @@ refer to the
...
+
+   If anything goes wrong in your tutorial journey, see this guide:
```
dmitry-lyfar: This note does not seem local enough if we compare it to the `Troubleshooting Workshop` section which is around the place where potential problems may appear. Perhaps, this note should be presented when one has already launched a workshop (or about to).

docs/tutorial/index.rst
```diff
@@ -630,8 +624,7 @@ use :ref:`connections <ref_workshop_connections>`:
      mount      dev/go:mod-cache  dev/system:mount  -
 
 
-This lists a :ref:`mount interface <exp_mount_interface>` plug
-named :samp:`dev/go:mod-cache`.
+This lists a mount interface plug named :samp:`dev/go:mod-cache`.
 As seen in the :command:`workshop info` output,
 it was automatically connected at :ref:`launch <tut_launch>`
 to the :samp:`dev/system:mount` slot,
```
dmitry-lyfar, line 589: ```and abbreviated here as :samp:`:mount` by convention.``` <- this is no longer true, we do not abbreviate system slots in the `connections` output anymore (but we do allow them in `connect`, for example).

docs/tutorial/index.rst
```diff
@@ -813,7 +806,7 @@ We're at the end of our tutorial;
 the only thing left is the cleanup.
 
 If you no longer need your workshop,
-:ref:`remove <ref_workshop_remove>` it:
+remove it:
```
dmitry-lyfar, line 767: The following description says: ```...any other content that was stored outside the workshop, e.g. via the [mount interface](https://canonical-workshop--365.com.readthedocs.build/en/365/tutorial/#tut-interfaces).```. `workshop remove` _will_ remove the content if it was stored in the default mount locations. If a plug was `remount` then that host directory is considered to be non-default and will not be removed.

dmitry-lyfar (changes requested):
That reads much cleaner! I would definitely get rid of the links in the `See also` section. The next steps appear to be enough (they refer both the reference and explanation sections).

dmitry-lyfar (approved):

jonathan-conder:
The fixup commit for this one got merged, maybe we should have an action or something to prevent that?

PR #367, opened 2025-04-09
Rework some SDK-related code

jonathan-conder:
# Description

Fixes some minor issues I noticed while working on https://github.com/canonical/workshop/pull/355 but aren't directly related to it:
- Removes unused constant `dirs.SdkDir`
- Removes outdated test `TestUndoLinkSdkRestorePreviousRev` (was only passing by accident)
- Fixes a typo in the `fakestore` tests
- Removed repeated `string` in the signature of `ImportVolume()`
- Adds a workaround for https://github.com/canonical/lxd/issues/14507 (fixes permissions of the apt cache and mounted SDKs)
- Moves string conversion from the call sites into `sdk.VolumeName()`
- Adds unit tests for `revert` (makes `go test ./... -check.v` work)
- Makes checking for the system, sketch and local/store SDKs a bit more expressive
- Removes some unnecessary SDK mocks from the API tests, in favour of the fake store
- Ensures changes in API tests which use `wait-on-error` fail for the expected reason
- Simplifies GCS store downloads using `revert`

The only behaviour change should be the volume permissions (now `0755` instead of `0711`).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/fakestore/store.go
```diff
@@ -250,7 +251,7 @@ func storeSdkInfoImpl(ctx context.Context, name, channel string) (storeSdk, erro
 	sSdk.Name = name
 	sSdk.Channel = channel
 	// A simple modulo to keep revision numbers in a readable form for testing
-	sSdk.Revision = sdk.Revision{N: int(atr.Generation % 1000)}
+	sSdk.Revision = sdk.Revision{N: int(atr.Generation%1000) + 1}
```
dmitry-lyfar, line 254: Nice catch.

internal/workshop/backend.go
```diff
@@ -90,7 +90,7 @@ type VolumeManager interface {
 	CreateVolume(ctx context.Context, name string) error
 
 	// Import a tarball into the volume. The tarball must be a valid tarball filepath.
-	ImportVolume(ctx context.Context, name string, tarball string) error
+	ImportVolume(ctx context.Context, name, tarball string) error
```
dmitry-lyfar: The logic I tend to stick to: if the variables of the same kind, e.g. names, they should be enumerated with comma and the type is provided once, otherwise, like in this case, `name` and `tarball` are of different kind (name and path) and their types described separately. Not something I insist on as a rule, though. I was considering this approach as something to support readability.

dmitry-lyfar (approved):

internal/workshop/backend.go
jonathan-conder: I'm not familiar with the conventions, let's stick with that and I'll just remove this commit when I rebase

internal/fakestore/store.go
jonathan-conder: I got lucky that this triggered in CI given the 1/1000 chance (https://github.com/canonical/workshop/actions/runs/14349544543/job/40225750601)

PR #373, opened 2025-04-14
Waiting workshops can be removed

jy5275:
# Description
https://warthogs.atlassian.net/jira/software/c/projects/WSP/boards/1645?selectedIssue=WSP-579

If a workshop is in Waiting state, `workshop remove` shall remove it.

The abort is done in backend. If workshop X is waiting on a failed refresh, `workshop remove A` clears all the middle state and starts the background aborting process before returns. A new context key "abort-background" was introduced to mark the changes that are running background abort.

The "abort-background" mark has the following effects:
1. The change skips all undo handlers when aborting. Because :
  - undo handlers are to clear some states of half-way refresh - for "abort-background" changes, this has already done in the do handlers of `workshop remove`. So it's not necessary.
  - If the user launch X immediately after remove X, the async abort probably hasn't finished yet. The remove undo handlers may clear the results of launch do handlers - that's not expected. So we have to skip the undo handlers here.

2. The change is ignored by `CheckChangeConflict`. Because the undo process dones't run the undo handlers, it has no effect to future changes thus shouldn't be considered as "conflict"

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -68,32 +68,6 @@ func (c *CmdRemove) Run(cmd *cobra.Command, av []string) error {
 		av = []string{name}
 	}
 
-	// TODO(yan.jiang): remove this after backend abort is ready
-	// // Workshops should be removed even if yml file formats are wrong
-	// // So use cli.List() instead of cli.Workshop() here
-	// wps, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
-	// if err != nil {
-	// 	return err
-	// }
-
-	// // convert list to map for faster lookup
-	// wpsMap := map[string]*client.WorkshopInfo{}
-	// for _, wp := range wps {
-	// 	wpsMap[wp.Name] = wp
-	// }
-
-	// for _, name := range av {
-	// 	if wp, ok := wpsMap[name]; !ok {
-	// 		return fmt.Errorf("workshop %s not found", name)
-	// 	} else if wp.Status == "Waiting" {
-	// 		fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
-	// 		cmdabort := &CmdRefresh{root: c.root, Abort: true}
-	// 		if err = cmdabort.RunRefresh(cli, project, []string{wp.Name}); err != nil {
-	// 			return err
-	// 		}
-	// 	}
-	// }
-
 	changeId, err := cli.Remove(project.Id, av)
 	if err != nil {
 		return err
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -68,6 +68,20 @@ func (c *CmdRemove) Run(cmd *cobra.Command, av []string) error {
 		av = []string{name}
 	}
 
+	for _, name := range av {
+		wp, err := cli.Workshop(project.Id, name)
+		if err != nil {
+			return err
+		}
+		if wp.Status == "Waiting" {
+			fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
+			cmdabort := &CmdRefresh{root: c.root, Abort: true}
+			if err = cmdabort.RunRefresh(cli, project, []string{wp.Name}); err != nil {
+				return err
+			}
+		}
+	}
+
 	changeId, err := cli.Remove(project.Id, av)
 	if err != nil {
 		return err
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/canonical/x-go/strutil"
 	"github.com/spf13/cobra"
+
+	"github.com/canonical/workshop/client"
 )
 
 type CmdRemove struct {
@@ -68,12 +70,23 @@ func (c *CmdRemove) Run(cmd *cobra.Command, av []string) error {
 		av = []string{name}
 	}
 
+	// Workshops should be removed even if yml file formats are wrong
+	// So use cli.List() instead of cli.Workshop() here
+	wps, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
+	if err != nil {
+		return err
+	}
+
+	// convert list to map for faster lookup
+	wpsMap := map[string]*client.WorkshopInfo{}
+	for _, wp := range wps {
+		wpsMap[wp.Name] = wp
+	}
+
 	for _, name := range av {
-		wp, err := cli.Workshop(project.Id, name)
-		if err != nil {
-			return err
-		}
-		if wp.Status == "Waiting" {
+		if wp, ok := wpsMap[name]; !ok {
+			return fmt.Errorf("workshop %s not found", name)
+		} else if wp.Status == "Waiting" {
 			fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
 			cmdabort := &CmdRefresh{root: c.root, Abort: true}
 			if err = cmdabort.RunRefresh(cli, project, []string{wp.Name}); err != nil {
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -5,8 +5,6 @@ import (
 
 	"github.com/canonical/x-go/strutil"
 	"github.com/spf13/cobra"
-
-	"github.com/canonical/workshop/client"
 )
 
 type CmdRemove struct {
@@ -70,30 +68,31 @@ func (c *CmdRemove) Run(cmd *cobra.Command, av []string) error {
 		av = []string{name}
 	}
 
-	// Workshops should be removed even if yml file formats are wrong
-	// So use cli.List() instead of cli.Workshop() here
-	wps, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
-	if err != nil {
-		return err
-	}
+	// TODO(yan.jiang): remove this after backend abort is ready
+	// // Workshops should be removed even if yml file formats are wrong
+	// // So use cli.List() instead of cli.Workshop() here
+	// wps, _, err := cli.List(&client.ListOptions{ProjectId: project.Id})
+	// if err != nil {
+	// 	return err
+	// }
 
-	// convert list to map for faster lookup
-	wpsMap := map[string]*client.WorkshopInfo{}
-	for _, wp := range wps {
-		wpsMap[wp.Name] = wp
-	}
+	// // convert list to map for faster lookup
+	// wpsMap := map[string]*client.WorkshopInfo{}
+	// for _, wp := range wps {
+	// 	wpsMap[wp.Name] = wp
+	// }
 
-	for _, name := range av {
-		if wp, ok := wpsMap[name]; !ok {
-			return fmt.Errorf("workshop %s not found", name)
-		} else if wp.Status == "Waiting" {
-			fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
-			cmdabort := &CmdRefresh{root: c.root, Abort: true}
-			if err = cmdabort.RunRefresh(cli, project, []string{wp.Name}); err != nil {
-				return err
-			}
-		}
-	}
+	// for _, name := range av {
+	// 	if wp, ok := wpsMap[name]; !ok {
+	// 		return fmt.Errorf("workshop %s not found", name)
+	// 	} else if wp.Status == "Waiting" {
+	// 		fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
+	// 		cmdabort := &CmdRefresh{root: c.root, Abort: true}
+	// 		if err = cmdabort.RunRefresh(cli, project, []string{wp.Name}); err != nil {
+	// 			return err
+	// 		}
+	// 	}
+	// }
 
 	changeId, err := cli.Remove(project.Id, av)
 	if err != nil {
--- a/cmd/workshop/remove.go
+++ b/cmd/workshop/remove.go
@@ -19,13 +19,13 @@ func (c *CmdRemove) Command() *cobra.Command {
 		Long: `
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'
+- Checks that the workshop isn't 'Off' or 'Pending'
 - Stops the workshop if it's not already 'Stopped'
 - Deletes the workshop but preserves its definition
 
 Notes:
 
-- If any listed workshop is 'Off', 'Pending' or 'Waiting', none are removed.
+- If any listed workshop is 'Off' or 'Pending', none are removed.
 
 - To rebuild a removed workshop from scratch, use 'workshop launch'.
 
--- a/internal/overlord/conflict/conflict.go
+++ b/internal/overlord/conflict/conflict.go
@@ -110,7 +110,7 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
 // Iterates over the list of running tasks and returns either nil or
 // a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
-func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
+func FindRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
 	for _, task := range st.Tasks() {
 		chg := task.Change()
 		if chg.IsReady() || slices.Contains(ignoreKinds, chg.Kind()) {
@@ -132,7 +132,7 @@ func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds 
 // there is a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
 func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKinds []string) error {
-	chg, err := findRunningChange(st, projectId, workshop, ignoreKinds)
+	chg, err := FindRunningChange(st, projectId, workshop, ignoreKinds)
 	if err != nil {
 		return err
 	}
@@ -147,6 +147,25 @@ func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKind
 	}
 }
 
+func AbortIfWaitingBeforeRemove(st *state.State, workshop, projectId string) error {
+	chg, err := FindRunningChange(st, projectId, workshop, []string{"exec"})
+	if err != nil {
+		return err
+	}
+	if chg == nil || chg.Status() != state.WaitStatus {
+		return nil
+	}
+	for _, tsk := range chg.Tasks() {
+		if tsk.Status() == state.WaitStatus {
+			tsk.SetStatus(state.DoStatus)
+			tsk.Logf("Aborting %q for workshop %q...", chg.Kind(), workshop)
+		}
+	}
+
+	chg.Abort()
+	return nil
+}
+
 // Attempt to resume the change associated with the Resume/Launch operation
 // for the given workshop. Depending on the mode the change will either be
 // turned into Doing (Continue mode) or Abort (Abort mode).
@@ -156,7 +175,7 @@ func ResumeAfterWait(st *state.State,
 		return nil, fmt.Errorf("cannot resume: only abort or continue can be used to resume the operation")
 	}
 
-	chg, err := findRunningChange(st, projectId, workshop, []string{"exec"})
+	chg, err := FindRunningChange(st, projectId, workshop, []string{"exec"})
 	if err != nil {
 		return nil, err
 	}
--- a/internal/overlord/conflict/conflict.go
+++ b/internal/overlord/conflict/conflict.go
@@ -110,10 +110,20 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
 // Iterates over the list of running tasks and returns either nil or
 // a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
+// Ignore discard-background refresh changes.
 func FindRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
 	for _, task := range st.Tasks() {
 		chg := task.Change()
 		if chg.IsReady() || slices.Contains(ignoreKinds, chg.Kind()) {
+			continue
+		}
+
+		var discardBackground bool
+		err := chg.Get("discard-background", &discardBackground)
+		if err != nil && !errors.Is(err, state.ErrNoState) {
+			return nil, err
+		}
+		if discardBackground {
 			continue
 		}
 
@@ -131,6 +141,7 @@ func FindRunningChange(st *state.State, projectId, workshop string, ignoreKinds 
 // Iterates over the list of running tasks and returns a ChangeConflictError if
 // there is a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
+// Ignore discard-background refresh changes.
 func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKinds []string) error {
 	chg, err := FindRunningChange(st, projectId, workshop, ignoreKinds)
 	if err != nil {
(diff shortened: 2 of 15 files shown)


jonathan-conder:
We only did the status check on the client side for `sketch-sdk` because it's a two-stage process, and refresh is only the second part. For `remove` I think it makes more sense to do in the backend, unless you have a good reason not to.

If the workshop is waiting on a refresh error, there's no need to restore the original workshop before removing it. But there are a few tasks we do want to undo, e.g. create-state-storage and register-sdk. If you haven't tried already, I suggest seeing what happens if you remove the workshop (and the stash) and only then abort the waiting change (allowing the undo handlers to fail if they have nothing to do).

For launch, I'm pretty sure you can just abort the change and not worry about removing the workshop afterwards.

jy5275:
Hi @jonathan-conder I've move the abort to backend as you suggested. 

When `workshop remove`, all tasks of the pending refresh change are directly set to `Undone` (undo handlers won't run), including those tasks which doesn't have an undo handler. I'm not sure if that would cause some other effects.

jonathan-conder:
> When `workshop remove`, all tasks of the pending refresh change are directly set to `Undone` (undo handlers won't run), including those tasks which doesn't have an undo handler. I'm not sure if that would cause some other effects.

Oh right, sorry I forgot to follow up on this after getting back home. Dmitry and I discussed it a bit more, he mentioned that we should try to avoid directly setting tasks to a final state. That's because, in the current design, the task runner is always responsible for setting the final state, which makes things easier to reason about.

I suggest we take the following approach: when removing, if there's a waiting change, we mark it in some way before aborting it. Possibly by modifying the `wait-setup` attribute, or introducing a new attribute. The conflict system should ignore changes which have been marked in this way, so the user can re-launch right away. We also need to make sure the undo handlers don't run, otherwise e.g. a workshop could be removed in the middle of the next `launch`. I think this can be done by wrapping the undo handlers in something similar to `OnDo`, e.g. `OnUndo`.

jy5275:
> when removing, if there's a waiting change, we mark it in some way before aborting it. Possibly by modifying the wait-setup attribute, or introducing a new attribute. The conflict system should ignore changes which have been marked in this way, so the user can re-launch right away.

Did you mean that the changes should still be aborted in background, and the purpose of this new attribute is to make the new launch not affected by unfinished abort ?

I agree that the task runner should be the only one to set a change's final state. Could you see if I understand it correctly:

```
runner.AddHandler("create-state-storage", OnDo(...), **OnUndo**(manager.doRemoveStateStorage))

...

func OnUndo(handler state.HandlerFunc) state.HandlerFunc {
	return func(task *state.Task, tomb *tomb.Tomb) error {
		if change.Get("wait-setup") != "background-abort" {
			return handler(task, tomb)     // return err
		}
		return nil        // else: do nothing
	}
}
```

jonathan-conder:
> Did you mean that the changes should still be aborted in background, and the purpose of this new attribute is to make the new launch not affected by unfinished abort ?

It serves 2 purposes:
- prevent the undo handlers from running
- prevent the change from conflicting with new changes

> I agree that the task runner should be the only one to set a change's final state. Could you see if I understand it correctly:

Yeah, that's pretty much it. Note that `wait-setup` is an existing attribute and I don't think it's a string. But the gist of it looks good to me.

jy5275:
Hi @jonathan-conder I've modified the implementation as our discussion last week. Could you review it when you have time

internal/overlord/conflict/conflict.go
```diff
@@ -147,6 +156,30 @@ func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKind
 	}
 }
 
+func BackgroundAbortWaitingRefresh(st *state.State, workshop, projectId string) error {
+	chg, err := FindRunningChange(st, projectId, workshop, []string{"remove"})
```
jonathan-conder: I suspect we don't need to exclude `remove` from this search. But I think we should still exclude `exec`. Otherwise it might find the `exec` change first and return `nil`.

internal/overlord/conflict/conflict.go
```diff
@@ -110,12 +110,20 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
...
-func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
+// Ignore abort-background refresh changes.
+func FindRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
 	for _, task := range st.Tasks() {
 		chg := task.Change()
 		if chg.IsReady() || slices.Contains(ignoreKinds, chg.Kind()) {
 			continue
 		}
+		var abortBackground bool
+		if err := chg.Get("abort-background", &abortBackground); err != nil && !errors.Is(err, state.ErrNoState) {
```
jonathan-conder: I think I saw you use "discard" in an earlier version of this. I think that's a better description than "abort background."

internal/overlord/conflict/conflict.go
```diff
@@ -147,6 +156,30 @@ func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKind
 	}
 }
 
+func BackgroundAbortWaitingRefresh(st *state.State, workshop, projectId string) error {
+	chg, err := FindRunningChange(st, projectId, workshop, []string{"remove"})
+	if err != nil {
+		return err
+	}
+	if chg == nil || chg.Status() != state.WaitStatus {
```
jonathan-conder, line 167: We should also check here if the change is a launch or refresh, similar to `ResumeAfterWait`.

Actually maybe you could merge this function with `ResumeAfterWait` if you don't mind adding `ChangeDiscard` alongside `ChangeAbort` and `ChangeContinue`. The caller can set `action = ""` and ignore errors. If `action == ""` the function shouldn't check `chg.Kind() != action`.

internal/overlord/handlersetup/handler_setup.go
```diff
@@ -11,6 +12,26 @@ import (
 	"github.com/canonical/workshop/internal/workshop"
 )
 
+// OnUnDo helps to skip the undo handler if the change is an abort-backgroud refresh
+func OnUnDo(handler state.HandlerFunc) state.HandlerFunc {
```
jonathan-conder:
```suggestion
// OnUndo helps to skip the undo handler if the change is an abort-backgroud refresh
func OnUndo(handler state.HandlerFunc) state.HandlerFunc {
```

internal/overlord/handlersetup/handler_setup.go
```diff
@@ -11,6 +12,26 @@ import (
...
+		var abortBackground bool
+		if err := change.Get("abort-background", &abortBackground); err != nil && !errors.Is(err, state.ErrNoState) {
+			st.Unlock()
+			return err
+		}
+		if abortBackground {
+			st.Unlock()
+			return nil
+		}
+		st.Unlock()
```
jonathan-conder:
```suggestion
		err := change.Get("abort-background", &abortBackground)
		st.Unlock()
		if err != nil && !errors.Is(err, state.ErrNoState) {
			return err
		}
		if abortBackground {
			return nil
		}
```
I think in cases like this it's worth breaking the `if err := ...` convention. Some handlers do it this way already. Could even merge the two `if`s if you think it's readable enough.

internal/overlord/handlersetup/handler_setup.go
```diff
@@ -53,6 +74,11 @@ func OnDo(handler state.HandlerFunc) state.HandlerFunc {
 					}
 				}
 
+			} else if change.Kind() == "remove" {
+				if task.Kind() == "remove-state-storage" || task.Kind() == "remove-workshop-stash" {
+					// 'pool volume not found' / 'stash instance not found' are not considered as error when removing
+					return nil
+				}
```
jonathan-conder: I don't think this is the right approach. The errors should be surfaced in `workshop tasks` but not block other cleanup from running. Maybe these tasks should be in separate lanes.

internal/overlord/workshopstate/request.go
```diff
@@ -952,7 +952,8 @@ func (w *WorkshopManager) RemoveMany(ctx context.Context, names []string, projec
 		}
 		workshops = append(workshops, wp)
 
-		allowed := []healthstate.Status{healthstate.ReadyStatus, healthstate.ErrorStatus, healthstate.StoppedStatus}
+		allowed := []healthstate.Status{healthstate.ReadyStatus, healthstate.ErrorStatus,
+			healthstate.StoppedStatus, healthstate.WaitingStatus}
 		if err = healthstate.CheckWorkshopHealth(w.state, wp, allowed); err != nil {
```
jonathan-conder: I think this should remain as is. The waiting change can be aborted right above this check, no need for an async task.

internal/overlord/workshopstate/manager.go
```diff
@@ -26,17 +26,18 @@ func New(st *state.State, runner *state.TaskRunner) *WorkshopManager {
...
 	runner.AddHandler("stop-workshop", OnDo(manager.doStop), manager.doStart)
 	runner.AddHandler("remove-workshop", OnDo(manager.doRemoveWorkshop), nil)
 	runner.AddHandler("mount-project", OnDo(manager.doMountProject), manager.undoMountProject)
 	runner.AddHandler("create-workshop-storage", OnDo(manager.doCreateWorkshopStorage), manager.doRemoveWorkshopStorage)
 	runner.AddHandler("remove-workshop-storage", OnDo(manager.doRemoveWorkshopStorage), nil)
 	runner.AddHandler("remove-workshop-stash", OnDo(manager.doRemoveWorkshopStash), nil)
-	runner.AddHandler("stash-workshop", OnDo(manager.doStashWorkshop), manager.undoStashWorkshop)
-	runner.AddHandler("create-state-storage", OnDo(manager.doCreateStateStorage), manager.doRemoveStateStorage)
+	runner.AddHandler("stash-workshop", OnDo(manager.doStashWorkshop), OnUnDo(manager.undoStashWorkshop))
+	runner.AddHandler("create-state-storage", OnDo(manager.doCreateStateStorage), OnUnDo(manager.doRemoveStateStorage))
```
jonathan-conder: is there a reason to only wrap some of the undo handlers with `OnUndo`?

tests/main/remove/task.yaml
```diff
@@ -31,7 +31,16 @@ execute: |
   workshop_exec remove ws-remove-mount
   workshop_exec list | MATCH "^.+[[:space:]]+ws-remove-mount[[:space:]]+Off[[:space:]]+-$"
 
+  echo "Check the workshop can be removed if in Waiting"
+  workshop_exec launch ws-remove
+  echo -e "sdks:\n  - name: test-sdk-setup-fail\n    channel: latest/stable" >> .workshop/ws-remove.yaml
```
jonathan-conder: might be more readable to do `cat <<EOF` here

tests/main/remove/task.yaml
```diff
@@ -31,7 +31,16 @@ execute: |
...
+  workshop_exec refresh --wait-on-error ws-remove || true
```
jonathan-conder:
```suggestion
  ! workshop_exec refresh --wait-on-error ws-remove
```

jonathan-conder (changes requested):

internal/overlord/conflict/conflict.go
jonathan-conder: After reading the whole PR I understand why `remove` is here now. But I think we won't need it after moving the abort into `RemoveMany`.

internal/overlord/workshopstate/manager.go
jy5275: I only wrap the handlers that are added to remove (revise: refresh) change

jonathan-conder: I think it should be the ones which are added to `launch` or `refresh` changes. But to keep things simple we could just do it for all `workshopstate` and `sdkstate` changes, say.

jonathan-conder: Probably also `ifacestate`

jy5275: Oh typo mistake... I wanted to say the handlers that are added to **refresh** change. But yes launch related tasks should also be wrapped.

jonathan-conder: ah ok, yeah I think `stop-workshop` doesn't technically need it. But the other two, `create-workshop-storage` and `mount-project` are part of launch. I think `mount-project` is in refresh too but that was only merged recently.

jonathan-conder: Since there's only one exception I would just stick `OnUndo` on all of them

tests/main/remove/task.yaml
jy5275: Updated

internal/overlord/workshopstate/request.go
jy5275: I think do the abort (or say discard) in `RemoveMany` rather than in a async task is better. 

But the abort (discard) only update the status of task. The status of the change and workshop are set in the another goroutine - it's async. So when code reach this line, the workshop status may still be `WaitingStatus`.

internal/overlord/workshopstate/manager.go
jy5275: Has wrapped all undo handlers with `OnUnDo`

internal/overlord/workshopstate/request.go
jonathan-conder: Maybe the change is still waiting, but the workshop status should not be `WaitingStatus`. Because `CheckChangeConflict` should be returning `nil` now - or is that not how it works in practice?

jy5275: You're right. The `healthstate.CheckWorkshopHealth` returns nil rather than the change's status. Local spread test has confirmed this.

internal/overlord/conflict/conflict.go
jy5275: > We should also check here if the change is a launch or refresh, similar to ResumeAfterWait.

Currently only remove and refresh change support --wait-on-error. I think all waiting changes should be treated the same way like this. Are there other situations that sets the change status to `WaitStatus` ?

jonathan-conder: I don't think so. You could say the same about `ResumeAfterWait`. I think we should be consistent with it, and remove the check for `refresh` and `launch` from `ResumeAfterWait` instead. The expanded error message (`%q is in progress`) could apply to any change so we should keep it

internal/overlord/handlersetup/handler_setup.go
jy5275: I removed this else-if in `OnDo` and put the two tasks in a separate lane, but the `workshop changes` output the remove change as `Error` and `workshop remove` commands returned non-zero exit code. Is that acceptable?

jonathan-conder: That's not ideal. Maybe the handlers for these tasks just shouldn't report "not found" errors. At least, when removing something, if the thing you want to remove doesn't exist then your job is already done.

jy5275: > when removing something, if the thing you want to remove doesn't exist then your job is already done.

I'd print log by `tsk.Logf` rather than throw an error when the stashed workshop or state volume doesn't exist. In that way the user can still check whether workshop actually did the remove by `workshop tasks xx`.

jonathan-conder: I don't think there's a need to log that. A situation like that might be useful for us when debugging, but I don't see why the user would care whether the volume was deleted by a particular task or something else.

jy5275: LXD APIs `DeleteInstance` and `DeleteStoragePoolVolume` return error when unable to find the resource. However, LXD Go package doesn't export the error type like "Instance not found". So we have to check the error string like `strings.Contains(err.Error(), "Instance not found")` to check if the returned error can be safely ignored. 

I think a more graceful way may be to call `GetInstance` to check if it exists before `DeleteInstance`. Which way you think is better?

jonathan-conder: can you try `api.StatusErrorCheck(err, http.StatusNotFound)`?

jonathan-conder: If that doesn't work then calling `GetInstance` is an OK workaround. I usually don't like that sort of thing because things can change between the two calls. But if it's just used to decide if there's an error or not, I can't think of a reason not to.

jy5275:
@jonathan-conder I've updated as your review and squashed all changes today into one commit for your review convenience.

I still keep `BackgroundDiscardWaitingRefresh` and `ResumeAfterWait` as two seperate functions. Because reusing `ResumeAfterWait` requires to add some unecessary params to `RemoveMany` signature.

jy5275:
Hi @akcano, there is a minor documentation update (`workshop remove` command) in this PR so I invited you to review

akcano (approved):

cmd/workshop/remove.go
```diff
@@ -19,13 +19,13 @@ func (c *CmdRemove) Command() *cobra.Command {
 		Long: `
 This command removes the workshops listed as arguments. For each workshop, it:
 
-- Checks that the workshop isn't 'Off', 'Pending' or 'Waiting'
+- Checks that the workshop isn't 'Off' or 'Pending'
```
jonathan-conder, line 22: nice catch

internal/overlord/conflict/conflict.go
```diff
@@ -110,13 +110,23 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
 // Iterates over the list of running tasks and returns either nil or
 // a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
-func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
+// Ignore discard-background refresh changes.
+func FindRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
```
jonathan-conder: I think there's no need to rename this anymore, it doesn't seem to be used outside of `conflict.go`.

internal/overlord/conflict/conflict.go
```diff
@@ -110,13 +110,23 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
 // Iterates over the list of running tasks and returns either nil or
 // a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
-func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds []string) (*state.Change, error) {
+// Ignore discard-background refresh changes.
```
jonathan-conder:
```suggestion
// Ignores discarded changes.
```

internal/overlord/conflict/conflict.go
```diff
@@ -110,13 +110,23 @@ func checkWorkshop(task *state.Task, projectId, workshop string) (bool, error) {
...
 
+		var discardBackground bool
+		err := chg.Get("discard-background", &discardBackground)
+		if err != nil && !errors.Is(err, state.ErrNoState) {
+			return nil, err
+		}
+		if discardBackground {
+			continue
+		}
+
```
jonathan-conder, line 129: I think we can go even simpler to just `discard` (or `discarded` maybe). Will leave the decision to you though.

internal/overlord/conflict/conflict.go
```diff
@@ -131,8 +141,9 @@ func findRunningChange(st *state.State, projectId, workshop string, ignoreKinds
 // Iterates over the list of running tasks and returns a ChangeConflictError if
 // there is a change running for the provided projectID / workshop pair.
 // Ignores certain kinds of changes based on the ignoreKinds argument.
+// Ignore discard-background refresh changes.
```
jonathan-conder:
```suggestion
// Ignore discarded changes.
```

internal/overlord/conflict/conflict.go
jonathan-conder: Just to clarify, I was agreeing with you that we don't need to check for `refresh` or `launch`. But as long as things are consistent I'm happy with this approach too.

internal/overlord/conflict/conflict.go
```diff
@@ -147,6 +158,36 @@ func CheckChangeConflict(st *state.State, projectId, workshop string, ignoreKind
...
+	}
+
+	if chg.Kind() != "refresh" && chg.Kind() != "launch" {
+		return fmt.Errorf("cannot discard %s: %s is in progress", workshop, chg.Kind())
+	}
+
+	for _, tsk := range chg.Tasks() {
+		if tsk.Status() == state.WaitStatus {
+			tsk.SetStatus(state.DoStatus)
+			tsk.Logf("Aborting %q for workshop %q...", chg.Kind(), workshop)
```
jonathan-conder:
```suggestion
			tsk.Logf("Discarding %q for workshop %q...", chg.Kind(), workshop)
```

internal/overlord/handlersetup/handler_setup.go
jonathan-conder: Still needs renaming. Note that "undo" is a word but "un" is not.

internal/overlord/workshopstate/handlers.go
```diff
@@ -309,7 +311,15 @@ func (m *WorkshopManager) doRemoveWorkshopStash(task *state.Task, tomb *tomb.Tom
 	ctx, cancel := BackendContext(tomb, user, prj.ProjectId)
 	defer cancel()
 
-	return m.backend.RemoveWorkshopStash(ctx, w)
+	err = m.backend.RemoveWorkshopStash(ctx, w)
+	if err != nil {
```
jonathan-conder:
```suggestion
	if err := m.backend.RemoveWorkshopStash(ctx, w); err != nil {
```

internal/overlord/workshopstate/handlers.go
```diff
@@ -309,7 +311,15 @@ func (m *WorkshopManager) doRemoveWorkshopStash(task *state.Task, tomb *tomb.Tom
...
+		if api.StatusErrorCheck(err, http.StatusNotFound) {
```
jonathan-conder: Does this work? I haven't tested myself

internal/overlord/workshopstate/request.go
```diff
@@ -951,6 +951,9 @@ func (w *WorkshopManager) RemoveMany(ctx context.Context, names []string, projec
 			return nil, fmt.Errorf("cannot remove %q: %w", name, err)
 		}
 		workshops = append(workshops, wp)
+		if err = conflict.BackgroundDiscardWaitingRefresh(w.state, name, projectId); err != nil {
+			return nil, fmt.Errorf("failed to find waiting refresh change of %q before remove: %w", name, err)
```
```suggestion
			return nil, fmt.Errorf("cannot remove %q: %w", name, err)
```
I think this should mirror the error message for `CheckWorkshopHealth`

internal/overlord/workshopstate/handlers.go
```diff
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"syscall"
 	"time"
 
+	"github.com/canonical/lxd/shared/api"
```
jonathan-conder: This file should be backend-agnostic. Instead the LXD backend should wrap 404 errors in a common way. For example: https://github.com/canonical/workshop/blob/33fa7949a7a4391ab0d6c38d4bc420a8abc653c9/internal/workshop/lxd/lxd_backend.go#L682-L688

internal/workshop/fakebackend/backend.go
```diff
@@ -458,7 +458,7 @@ func (s *FakeWorkshopBackend) RemoveWorkshopStash(ctx context.Context, name stri
 	defer s.workshopLock.Unlock()
 
 	if s.StashedWorkshops[projectId][workshop.StashNamePrefix+name] == nil {
-		return fmt.Errorf("stashed workshop %q not found", name)
+		return api.StatusErrorf(http.StatusNotFound, fmt.Sprintf("stashed workshop %q not found", name))
```
jonathan-conder: I think this is OK given that it appears elsewhere in the same file. But hopefully if you move the 404 detection to the LXD backend this will also become backend-agnostic.

jonathan-conder (changes requested):

internal/overlord/workshopstate/handlers.go
jy5275: Yes this file should be beckend-agnostic. But I don't think it's necessary for `RemoveWorkshopStash` to convert one type error to another (`http.StatusNotFound` -> `workshop.ErrWorkshopNotLaunched`). Since the resource we want to remove not exist shouldn't be considered as an error, we can just return err=nil here. What do you think

internal/overlord/workshopstate/handlers.go
jy5275: Yes it works. It passed my local test and all CI checks

internal/overlord/conflict/conflict.go
jy5275: I think "background" keyword could remind the future developer that the discard is async. Getting a true value here dones't means that the change is fully "discarded" because some status is not set to final value.

internal/overlord/conflict/conflict.go
jonathan-conder: i would at least remove the refresh

internal/overlord/workshopstate/handlers.go
jonathan-conder: Yeah that sounds good too

jonathan-conder (approved):
Looking good now, thanks! There's a merge conflict and you might want to do a bit more squashing, even just squash everything into 1 commit if you want.

jy5275:
> Looking good now, thanks! There's a merge conflict and you might want to do a bit more squashing, even just squash everything into 1 commit if you want.

I just rebased origin/main branch and squashed some commits. I think some intermediate commits like (cli-side abort before remove) may serve as reference for future work, so I didn't squash all commits into one

PR #374, opened 2025-04-14
Support project-defined SDKs

jonathan-conder:
# Description

Adds in-project SDKs, which are lightweight SDKs with the following layout:
```
.workshop/
  nav2/
    sdk.yaml
    hooks/
      setup-base
```

Also updates the sketch SDK's internal layout to match (i.e. `meta/sdk.yaml -> sdk.yaml` and `sdk/hooks -> hooks`).
The old layout is still valid, so you can just extract a store SDK into the project, but any ambiguity results in an error.

Workshops within the project can use these SDKs with the `project-` prefix:
```yaml
name: dev
base: ubuntu@22.04
sdks:
  - name: project-nav2
    plugs:
      ...
    slots:
      ...
```

On launch or refresh, copies are stored in `$XDG_DATA_HOME/workshop/id/****/dev/sdk/project-nav2/{x1,x2,...}` and bind-mounted in the workshop, similar to the sketch SDK. Unlike the sketch SDK, hooks must be separate files.

Documentation will follow in the next PR (`sketch-sdk --eject`).

---

Fixes a bug where unmodified local SDKs are updated on every refresh if the user and root have different `umask`s.

Fixes an issue which broke the workshop's integrity if the `download-base` task failed during refresh.

Allows local SDKs (i.e. sketch and in-project SDKs) to optionally specify a base. For example, `setup-base` might install a package only available in `24.04`, or in-project SDKs might include binary artifacts (which is not recommended, but still possible).

---

Will probably break existing sketch SDKs

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/workshop_file.go
```diff
@@ -36,6 +36,17 @@ func Filepath(project, name string) string {
 	return filepath.Join(project, Directory, filename(name))
 }
 
+func ProjectSdkPath(project, name string) string {
+	return filepath.Join(project, Directory, name)
+}
+
+func ExpandSdkSource(source, project string) string {
+	if strings.HasPrefix(source, "$PROJECT/") {
```
dmitry-lyfar, line 44: Perhaps, checking if it has the prefix is not necessary.

internal/overlord/workshopstate/request.go
```diff
@@ -379,7 +393,9 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
 		all.AddAll(ts)
 	}
 
+	base := retrieveBase(st, file)
```
dmitry-lyfar, line 396: I wonder how the client is going to render progress bars for an SDK and a base that are relatively same size. Not a significant concern for now.

internal/workshop/workshop_file.go
jonathan-conder: This is consistent with similar code elsewhere. Definitely a huge edge case but technically someone could set their `XDG_DATA_HOME='/home/user/$PROJECT'` or something.

dmitry-lyfar: I see, yes, we use it in the mount interface for this reason, I think.

internal/overlord/workshopstate/request.go
jonathan-conder: I didn't think about that. The SDKs are already downloaded concurrently from what I can tell. I feel like we should improve the client handling rather than serialising things on the backend.

internal/workshop/workshop_file.go
jonathan-conder: We could do a regex replace instead, e.g. `^\$PROJECT`

dmitry-lyfar (approved):

PR #375, opened 2025-04-15
Improve sketch template

dmitry-lyfar:
# Description

* Remove redundant URLs from the sketch template.
* Simplify sketch implementation and improve `sketch-sdk` info messages shown to the end-user (e.g. instead of showing a refresh operation outcome which is the underlying sketch operation, show `"<workshop>" sketch refreshed`).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -80,19 +80,17 @@ var sketchTemplate = `# Sketch SDK for %s
 # Sketch SDK provides local customisation of this specific workshop.
 
 # To read more about SDKs, their components and syntax, see:
-# https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/
+# https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/sdks/#sketch-sdk
 name: sketch
 
 hooks:
   # EXAMPLE: setup-base runs once at workshop launch, use it to install some packages.
-  # See https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/hooks/
   # setup-base: |
     # apt-get update
     # apt-get install PACKAGE...
     # snap install SNAP...
 
   # EXAMPLE: check-health runs after all SDK setup completes, call 'workshopctl set-health okay' for OK.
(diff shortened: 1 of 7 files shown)


cmd/workshop/sketch.go
```diff
@@ -80,19 +80,17 @@ var sketchTemplate = `# Sketch SDK for %s
 # Sketch SDK provides local customisation of this specific workshop.
 
 # To read more about SDKs, their components and syntax, see:
```
jonathan-conder:
```suggestion
# To read more about the sketch SDK, its components and syntax, see:
```

jonathan-conder (approved):
Looks good, I like the simplifications. At some point I think we should return to only refreshing the sketch SDK. But the implementation will probably look different so it's good to remove the current approach.

There's still the issue of `workshop refresh --abort` being too quiet. Might make sense to fix here as well?

cmd/workshop/sketch.go
```diff
@@ -244,11 +239,11 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
 		}
 	}
 
-	// Ensure that the workshop is Ready,
-	// aborting previous sketch if necessary.
+	// Ensure that the workshop is Ready, aborting previous sketch if necessary.
 	if wp.Status == "Waiting" {
+		fmt.Fprintf(Stdout, "Reverting incomplete change for %q...\n", wp.Name)
```
dmitry-lyfar, line 244: @akcano  It would be good to know your thoughts when you're back. If `sketch-sdk` detects an outstanding change, it will abort the change first. Could you try to play with it and see if the messaging gets it right.

jonathan-conder: I added a similar message for `sketch-sdk --remove`, see https://github.com/canonical/workshop/pull/380:
```console
$ workshop sketch-sdk --remove
- Run hook "setup-base" for "project-tools" SDK (command exit code 1)
"dev" refresh aborted
Restoring "dev" sketch
```

Wasn't able to add `...` because the staticcheck linter doesn't like it.

PR #380, opened 2025-04-22
Eject option for sketch SDK command

jonathan-conder:
# Description

Adds `sketch-sdk --eject [--name <NAME>]`, which is like `sketch-sdk --remove` but copies the sketch SDK files into `.workshop/<NAME>/` first. It renames the SDK to `<NAME>` and extracts hooks from `sdk.yaml`. The default SDK name is the project directory name. The output looks like:
```console
$ workshop sketch-sdk --eject dev
"dev" sketch ejected to ".workshop/myproject"
To use it, add "project-myproject" to the list of SDKs and run 'workshop refresh dev'
```

The main part I'm not too happy with is that transforming `sdk.yaml` changes the formatting. This seems to be due to how `yaml` parses certain comments. In particular, something like:
```yaml
foo:
  # comment
```
is treated like:
```yaml
foo: null
  # comment
```
and so the comment is attached to the next node, or the document itself, rather than `foo`. This affects the sketch template pretty badly but should be less of an issue for cleaner SDK definitions.

If the refresh fails for `sketch-sdk --remove`, the sketch SDK files are not removed. Subsequent attempts to remove the sketch SDK will work as expected (once the issue causing refresh failure is fixed). The output in this case looks like:
```console
$ workshop sketch-sdk --remove
- Run hook "setup-base" for "project-tools" SDK (command exit code 1)
"dev" refresh aborted
Restoring "dev" sketch
```

I made a first pass on the docs, will probably revisit again when I've updated SDKcraft to make our SDK definitions more consistent.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -54,9 +54,6 @@ Notes:
 
 - In addition to hooks, the 'sketch' SDK can use interfaces,
   define plugs, slots, connections and bindings.
-
-- You can partially refresh the workshop, targeting the 'sketch' SDK
-  with the 'workshop refresh <WORKSHOP>/sketch' command.
 `,
 		Example: `
 Edit the sketch SDK definition for the 'nimble' workshop
@@ -410,8 +407,7 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
 		cmdrefresh := &CmdRefresh{root: c.root}
 		if err = cmdrefresh.RunRefresh(cli, p, []string{wp.Name}); err != nil {
 			// Refresh failed, revert the stash operation so a possible subsequent
-			// "workshop refresh <WORKSHOP>/sketch" won't fail due to the lack of
-			// sketch SDK definition.
+			// refresh won't fail due to the lack of a sketch SDK definition.
 			return err
 		}
 		fmt.Fprintf(Stdout, "%q sketch stashed\n", wp.Name)
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -82,14 +82,12 @@ var sketchTemplate = `# Sketch SDK for %s
 # To read more about the sketch SDK, its and syntax, see:
 # https://canonical-workshop.readthedocs-hosted.com/en/latest/explanation/sdks/sdks/#sketch-sdk
 name: sketch
-
 hooks:
   # EXAMPLE: setup-base runs once at workshop launch, use it to install some packages.
   # setup-base: |
     # apt-get update
     # apt-get install PACKAGE...
     # snap install SNAP...
-
   # EXAMPLE: check-health runs after all SDK setup completes, call 'workshopctl set-health okay' for OK.
   # check-health: |
     # if CHECK_HEALTH_COMMAND ; then
@@ -97,17 +95,14 @@ hooks:
     # else
     #   workshopctl set-health --code=installation-failed error "Installation failed"
     # fi
-
 plugs:
   # EXAMPLE: forward your SSH agent into the workshop enabling 'git push' inside the workshop.
   # ssh-agent:
   #   interface: ssh-agent
-
   # EXAMPLE: expose well-known config file locations to the workshop
   # vs-code-settings:
   #   interface: mount
   #   workshop-target: /home/workshop/.config/Code/User
-
 slots:
   # EXAMPLE: expose SDK services to the host
   # dashboard:
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -523,6 +523,9 @@ func editSketchSdk(sketchdir, workshopFile string) error {
 		return err
 	}
 	if err := writeSketchHooks(temp, content); err != nil {
+		// If writeSketchHooks failed, we don't want to refresh
+		// but we do want to remember the user's edits for next time.
+		_ = osutil.Exchange(temp, sketchdir)
 		return err
 	}
 
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -190,30 +190,36 @@ func restoreSketch(sketchdir, stashdir string) error {
 	return nil
 }
 
-func removeSketch(sketchdir string) error {
+func hideSketch(sketchdir string) (string, *revert.Reverter, error) {
 	_, err := os.Stat(sketchdir)
 	if err != nil && !osutil.IsDirNotExist(err) {
-		return err
+		return "", nil, err
 	}
 	if osutil.IsDirNotExist(err) {
 		// Nothing to do.
-		return fmt.Errorf(`cannot remove: the 'sketch' SDK doesn't exist`)
+		return "", nil, fmt.Errorf(`cannot remove: the 'sketch' SDK doesn't exist`)
 	}
+
+	reverter := revert.New()
+	defer reverter.Fail()
 
 	temp, err := os.MkdirTemp(filepath.Dir(sketchdir), "sketch-*")
 	if err != nil {
-		return err
+		return "", nil, err
 	}
-	defer os.RemoveAll(temp)
+	reverter.Add(func() { _ = os.RemoveAll(temp) })
 	if err := os.Chmod(temp, 0755); err != nil {
-		return err
+		return "", nil, err
 	}
 
 	if err := osutil.Exchange(sketchdir, temp); err != nil {
-		return err
+		return "", nil, err
 	}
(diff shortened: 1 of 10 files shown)


docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
 ------------------------
 
+.. @artefact project SDK
+
 If you're satisfied with the sketch to a degree
-where think others may benefit from it as well,
-the next possible step is to refine it into a permanent SDK for publishing.
+where others may benefit from it,
+you can either
+add it to your project
```
akcano:
```suggestion
you can either add it to your project
```

Overdoing it a bit IMO.

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
 allowing you to restore and continue working later.
 
 
-Craft the SDK (optional)
+Eject the SDK (optional)
 ------------------------
 
+.. @artefact project SDK
```
akcano:
```suggestion
.. @artefact project SDK
.. @artefact SDK Store
```

This adds some new bits of knowledge.

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
+
 If you're satisfied with the sketch to a degree
-where think others may benefit from it as well,
-the next possible step is to refine it into a permanent SDK for publishing.
+where others may benefit from it,
+you can either
+add it to your project
+or publish it in the SDK store.
```
akcano:
```suggestion
or publish it in the SDK Store.
```

Capitalization, globally.

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
+
+   $ workshop sketch-sdk --eject
+
+     "dev" sketch ejected to ".workshop/hello-workshop"
+     To use it, add "project-hello-workshop" to the list of SDKs and run 'workshop refresh dev'
+
+
+This moves the SDK definition files
+into the :file:`.workshop/` subdirectory of the project,
+and removes the sketch from the running workshop.
```
akcano:
```suggestion
into the :file:`.workshop/` subdirectory of the project
and removes the sketch from the running workshop.
```

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
+Workshops can use SDKs in this directory directly,
```
akcano:
```suggestion
Workshops can pull SDKs from this directory directly,
```

I think this better emphasizes that the resulting SDK happens elsewhere; the directory stores the blueprint, but nothing occurs *in* it per se.

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
+bypassing the SDK store.
+
+If you'd prefer to publish the SDK for other projects to use,
+feel free to move the files elsewhere.
```
akcano:
```suggestion
If you'd rather publish the SDK for other projects to use,
use the common crafting procedure.
```

This was a bit vague to my taste.

docs/how-to/use-workshops/sketch-sdk.rst
```diff
@@ -222,14 +222,49 @@ Stashing does not delete the SDK,
...
 For details, see the |sdk_markup| :ref:`how-to guide <how_sdkcraft>`.
 
+The new SDK is named after the project directory by default,
+but this can be customised:
```
akcano:
```suggestion
The new SDK is named after the project directory by default;
use the :option:`!--name` option to change this:
```

docs/.coverage.yaml
```diff
@@ -205,6 +205,11 @@
     - SK000
     - SK013
 
+"project SDK":
```
akcano: I suggest stopping on `in-project SDK` for consistency, otherwise there'll be no end to it.

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -179,6 +182,27 @@ To the contrary, plugs added under the system SDK can be auto-connected
...
+In-project SDKs
+~~~~~~~~~~~~~~~
+
+.. @artefact project SDK
+
+Projects can define their own SDKs
+to configure the workshop in a project-specific way.
+These SDKs are defined by files named like :file:`.workshop/<NAME>/sdk.yaml`,
+relative to the project directory.
+They can also define hooks in :file:`.workshop/<NAME>/hooks/`.
```
```suggestion
Accordingly, SDK hooks are stored under :file:`.workshop/<NAME>/hooks/`.
```

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -179,6 +182,27 @@ To the contrary, plugs added under the system SDK can be auto-connected
...
+
+Workshops within the project consume these SDKs
+using names like :samp:`project-<NAME>`.
+A :samp:`channel` is not required.
```
```suggestion
using names like :samp:`project-<NAME>`;
a :samp:`channel` is not required in this case.
```

Correlated bits of info.

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -179,6 +182,27 @@ To the contrary, plugs added under the system SDK can be auto-connected
...
+When a workshop is refreshed,
+|ws_markup| detects if the SDK definition files have changed,
+and updates the workshop if needed.
```
akcano: I'd drop this altogether because this is our normal behaviour, so dwelling on it only creates chances of confusion (e.g. why reiterate this?).

akcano (changes requested):

docs/how-to/use-workshops/sketch-sdk.rst
jonathan-conder: I'll also remove "directly" as it reads a bit odd. And change "Workshops" to "Workshop."

docs/how-to/use-workshops/sketch-sdk.rst
jonathan-conder: Agree but I don't like having "use, use." Trying another way

docs/.coverage.yaml
jonathan-conder: Not sure what you mean, for now I'll make everything say "in-project SDK" but we can discuss the naming further.

docs/reference/definition-files/workshop-definition.rst
jonathan-conder: Fine with me. We might want to mention something about version controlling the files. But I think it's kinda obvious so maybe not

cmd/workshop/sketch.go
```diff
@@ -301,21 +437,50 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
...
+		if err != nil {
+			return fmt.Errorf("cannot remove: %w", err)
+		}
+		defer reverter.Fail()
+
 		cmdrefresh := &CmdRefresh{root: c.root}
 		if err = cmdrefresh.RunRefresh(cli, p, []string{wp.Name}); err != nil {
-			return err
+			// Refresh failed, revert the hide operation so the user can try again.
+			return fmt.Errorf("%v\nRestoring %q sketch", err, wp.Name)
```
dmitry-lyfar:
```suggestion
			return fmt.Errorf("%v\n%q sketch restored", err, wp.Name)
```
? Not exactly correct, but in line with other messages, e.g. `"dev" refresh aborted`.

jonathan-conder: only reason I didn't do it is because the restore happens in a `defer`. But maybe it's worth reworking the code to get a better message?

dmitry-lyfar (approved):

cmd/workshop/sketch.go
jonathan-conder: We ended up deciding to remove it. Not attaching to the error message means it has to appear first, which looks weird. In most cases these messages only appear on success anyway. Keeping the sketch is likely what users expect.

PR #385, opened 2025-05-01
Add verbose mode for hooks

dmitry-lyfar:
# Description

* Implement verbose mode for `launch` and `refresh`.
* Show tasks duration in `workshop tasks` instead of spawn and ready times that often do not bear any useful information as `SpawnTime` is the time when a task was created for the change (not when it was scheduled for running). Thus, it's difficult to conclude when each task has actually started.

## Considerations and assumptions made for the implementation

- Minimise changes to the client and API.

- The client needs to be significantly extended to support chunked streaming or websockets for the functionality like 
verbose logs streaming and it is unclear whether such an  investment will pay off in our use case which I define as "occasional usage of verbose output for progress tracking and debugging".

- The final task log must be limited so that logs do not accumulate in the state file which leads to performance issues quite quickly as it has been seen in pebble.

- The combined stdout and stderr streams from hooks are stored in the state's cache. These are available in the `/v1/change/<id>?verbose=true` response for the duration of the change execution and up until the next ensure pass. After that, the cache will be invalidated and only a truncated version of the output streams remain available in the task's log (stored in state).

- The `state.Task` stores logs as `[]string`; the client also expects logs as `[]string`. Thus, the API will send the combined output as lines which means buffering for the output data that wasn't finished with a new line.

- There are possible strategies to optimise the verbose mode performance, and extend the lifetime of verbose logs (e.g. store in a temporary file up until the time the associated change is pruned, extend the cache's expiration if it was requested after the change has finished and so on). These were considered as premature to be implemented now.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/launch.go
+++ b/cmd/workshop/launch.go
@@ -72,30 +72,17 @@ $ workshop launch`,
 	cmd.PersistentFlags().BoolVar(&c.NoWait, "no-wait",
 		false,
 		"Return the change ID, don't wait for the operation to finish.")
+	cmd.PersistentFlags().BoolVar(&c.verbose, "verbose",
+		false,
+		"Combine stdout and stderr output from hooks.")
+
+	cmd.MarkFlagsMutuallyExclusive("abort", "continue", "wait-on-error")
 
 	return cmd
 }
 
 func (c *CmdLaunch) Run(cmd *cobra.Command, av []string) error {
 	av = strutil.Deduplicate(av)
-
-	if c.Abort && c.Continue {
-		return fmt.Errorf("cannot launch: '--abort' incompatible with '--continue'")
-	}
-
-	if c.WaitOnError && c.Abort {
-		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with '--abort'")
-	}
-
-	if c.WaitOnError && c.Continue {
-		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with '--continue'")
-	}
-
-	// We should have no more than one argument (a single workshop) for a
-	// wait-on-error operation
-	if (c.Abort || c.Continue || c.WaitOnError) && len(av) > 1 {
-		return fmt.Errorf("cannot launch: '--wait-on-error' incompatible with multiple workshops")
-	}
 
 	cli, err := c.root.client()
 	if err != nil {
--- a/cmd/workshop/launch_test.go
+++ b/cmd/workshop/launch_test.go
@@ -161,36 +161,6 @@ func (m *workshopLaunch) TestLaunchWaitOnErrorContinuedSuccessfully(c *check.C) 
 	c.Assert(m.stdout.String(), check.Matches, `"ws" launched\n`)
 }
 
-func (m *workshopLaunch) TestLaunchIncompatibleOptions(c *check.C) {
-	cmd := &CmdLaunch{root: &CmdRoot{}}
-	cmd.Abort = true
-	cmd.Continue = true
-
-	err := cmd.Run(nil, []string{"ws"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--abort' incompatible with '--continue'")
-
-	cmd.WaitOnError = true
-	cmd.Abort = false
-	cmd.Continue = true
-
-	err = cmd.Run(nil, []string{"ws"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--wait-on-error' incompatible with '--continue'")
-
-	cmd.WaitOnError = true
-	cmd.Abort = true
-	cmd.Continue = false
-
-	err = cmd.Run(nil, []string{"ws"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--wait-on-error' incompatible with '--abort'")
-
-	cmd.WaitOnError = true
-	cmd.Abort = false
-	cmd.Continue = false
-
-	err = cmd.Run(nil, []string{"ws", "ws-1"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--wait-on-error' incompatible with multiple workshops")
-}
-
 func (m *workshopLaunch) TestLaunchCompletions(c *check.C) {
 	wsInfo := []*client.WorkshopInfo{
 		{
--- a/internal/daemon/api_changes.go
+++ b/internal/daemon/api_changes.go
@@ -17,11 +17,13 @@ package daemon
 import (
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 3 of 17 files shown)


cmd/workshop/launch_test.go
```diff
@@ -161,36 +161,6 @@ func (m *workshopLaunch) TestLaunchWaitOnErrorContinuedSuccessfully(c *check.C)
...
-
-	err = cmd.Run(nil, []string{"ws"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--wait-on-error' incompatible with '--abort'")
-
-	cmd.WaitOnError = true
-	cmd.Abort = false
-	cmd.Continue = false
-
-	err = cmd.Run(nil, []string{"ws", "ws-1"})
-	c.Assert(err, check.ErrorMatches, "cannot launch: '--wait-on-error' incompatible with multiple workshops")
```
jonathan-conder, line 191: This one is still valid, might be worth keeping? Maybe not since the backend handles this case anyway.

internal/daemon/api_changes.go
```diff
@@ -86,12 +88,32 @@ func change2changeInfo(chg *state.Change) *changeInfo {
 	for j, t := range tasks {
 		label, done, total := t.Progress()
 
+		var taskLog []string
+		// Only hooks support verbose output.
+		if verbose && t.Kind() == "run-hook" {
+			cached := chg.State().Cached(t.ID())
```
jonathan-conder: The cache key should be wrapped in a new type, similar to `ExecArgsKey`

internal/overlord/hookstate/handlers.go
```diff
@@ -77,21 +84,65 @@ func (h *HookManager) doRunHook(task *state.Task, tomb *tomb.Tomb) error {
...
+	return &HookLog{
+		buf: strutil.NewLimitedBuffer(-1, hookRawOutputBufferSize),
+	}
+}
+
+func (h *HookLog) Output() *bytes.Buffer {
+	h.l.Lock()
+	defer h.l.Unlock()
+
+	return bytes.NewBuffer(h.buf.Bytes())
```
jonathan-conder: I think this needs a copy, looks like `NewBuffer` and `Bytes` both use the underlying array.

internal/overlord/hookstate/handlers.go
```diff
@@ -77,21 +84,65 @@ func (h *HookManager) doRunHook(task *state.Task, tomb *tomb.Tomb) error {
...
+}
+
+func (h *HookLog) taskLog() []byte {
+	h.l.Lock()
+	defer h.l.Unlock()
+
+	// 10 lines per task log to be stored in the state as a sensible default
+	// (see task.Logf).
+	return strutil.TruncateOutput(h.buf.Bytes(), 10, hookTaskLogSize)
```
jonathan-conder: should probably also make a copy

internal/overlord/hookstate/handlers_test.go
```diff
@@ -518,3 +518,55 @@ func (s *hookSuite) launchWorkshop(c *check.C, newsdk string) {
...
 	c.Check(err, check.IsNil)
 }
+
+func (s *hookSuite) TestExecCombinedHookOutputStoredInLogs(c *check.C) {
+	s.state.Lock()
+	defer s.state.Unlock()
+	t1 := hookstate.Hook(s.state, s.project.ProjectId, "ws", "one", 0, hookstate.FakeHook)
+	// The context handler will return an error that must be the final error of
+	// the task.
+	//s.mockHandler.ErrorError = true
```
jonathan-conder: is this needed?

internal/overlord/hookstate/manager.go
```diff
@@ -78,6 +78,7 @@ func New(s *state.State, runner *state.TaskRunner) *HookManager {
 	s.Unlock()
 
 	runner.AddHandler("run-hook", handlersetup.OnDo(manager.doRunHook), nil)
+	runner.AddCleanup("run-hook", manager.doHookCleanup)
```
jonathan-conder, line 81: As discussed, the logs can get truncated on the client side. Apart from adjusting the client, this might be too early to remove the cache entry

cmd/workshop/launch.go
```diff
@@ -72,31 +72,18 @@ $ workshop launch`,
 	cmd.PersistentFlags().BoolVar(&c.NoWait, "no-wait",
 		false,
 		"Return the change ID, don't wait for the operation to finish.")
+	cmd.PersistentFlags().BoolVar(&c.verbose, "verbose",
+		false,
+		"Display combined stdout and stderr for hooks.")
```
akcano:
```suggestion
		"Combine stdout and stderr output from hooks.")
```

Globally.

akcano (changes requested):
LGTM, one minor update.

internal/overlord/hookstate/manager.go
dmitry-lyfar: The cleanup tasks are run on every ensure which means every next change or in 5 minutes. The tasks will have a truncated output attached to them and stored in the state until their changes are pruned as, arguably, the last few lines are often most important. 

Potentially, we can store these logs for longer (in memory or as in a file), but it requires another command or an option to view those. Something like, `workshop tasks --verbose` or `workshop logs` which would also accommodate `workshop launch --no-wait` as if it fails, we only see truncated logs in the tasks output. For now, it is possible to:

```
workshop launch --verbose > full.log
```

cmd/workshop/launch_test.go
dmitry-lyfar: Yes, I would leave this kind of knowledge for the API, so if it becomes available we don't have to change the client. 
On the other hand, the action mode is something that the client has to know about, though, as it needs to choose what to pass to the API.

internal/overlord/hookstate/manager.go
jonathan-conder: This might be a useful way to store the logs: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/. Not sure if it's possible currently because the socket uses `SOCK_DGRAM` and I assume proxy devices use `SOCK_STREAM`. Well, I suppose we can just wrap hooks with a mini program that handles the logging (or make them systemd units).

internal/progress/progress.go
```diff
@@ -22,6 +22,14 @@ import (
 	"golang.org/x/term"
 )
 
+type NotifyerType uint8
```
jonathan-conder:
```suggestion
type NotifierType uint8
```

jonathan-conder (approved):

jonathan-conder (approved):

PR #390, opened 2025-05-13
Add setup-project hooks

jonathan-conder:
# Description

Adds the `setup-project` hook, which runs after interfaces are connected. This allows it to take advantage of cache mounts (e.g. for `pip`) and access the project directory. The project directory is now mounted after all SDKs are installed, so the post-`setup-base` snapshots can't be influenced by the contents of the project.

Some tangentially related changes:
- Workshop runs save and restore state hooks for all SDKs that remain in the refreshed workshop, not just those which were refreshed.
- Hooks are run without the `nounset` bash option (I noticed in `setup-project` there are scripts in `/etc/profile.d` which don't respect this option).
- Attaching volumes won't modify the permissions of the root directory. This workaround is no longer needed for the apt cache, and SDKs are mounted read-only so there's no point changing it for them.
- Non-test code uses `dirs.XdgRuntimeDirBase` instead of hard coding `"/run/user"`.
- The cloud-init config uses the `apt` module instead of creating a config file manually.

This will break SDKs which currently use `/project` in `setup-base`. Only one I'm aware of is ROS

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/reference/sdks.rst
+++ b/docs/reference/sdks.rst
@@ -326,6 +326,7 @@ which can be defined when the SDK is built using |sdk_markup|:
 .. @artefact SDK base image
 .. @artefact setup-base
 .. @artefact workshop base image
+.. @artefact setup-project
 
 .. list-table::
    :header-rows: 1
@@ -336,26 +337,24 @@ which can be defined when the SDK is built using |sdk_markup|:
      - When |ws_markup| runs it
      - What it does
 
-   * - :samp:`check-health`
-     - At :ref:`ref_workshop_launch`:
-       after running :samp:`setup-base` hooks for *all* SDKs.
+   * - :samp:`setup-base`
 
-       At :ref:`ref_workshop_refresh`:
-       after running :samp:`restore-state` hooks for *all* SDKs.
+     - At :ref:`ref_workshop_launch`, :ref:`ref_workshop_refresh`:
+       after unpacking the base image
+       and starting the workshop,
+       but before mounting the project directory
+       and connecting plugs and slots.
 
-     - Sets the state of the SDK
-       (:samp:`okay`, :samp:`waiting` or :samp:`error`)
-       using :ref:`workshopctl <ref_workshopctl>`,
-       which affects the :ref:`status <ref_workshop_status>` of the workshop.
+     - Configures system packages and services required by the SDK.
 
-   * - :samp:`restore-state`
+   * - :samp:`setup-project`
 
-     - At :ref:`ref_workshop_refresh`:
-       after launching the new workshop
-       and running the :samp:`setup-base` hook for the SDK.
+     - At :ref:`ref_workshop_launch`, :ref:`ref_workshop_refresh`:
+       after mounting the project directory
+       and connecting plugs and slots
+       but before the workshop is set to *Ready*.
 
-     - Restores SDK-specific data from the :ref:`state directory <ref_sdk_state>`.
-       The hook itself comes from the *new* SDK revision.
+     - Configures the user environment for the SDK to become operational.
 
    * - :samp:`save-state`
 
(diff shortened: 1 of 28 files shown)


docs/reference/sdks.rst
```diff
@@ -370,15 +371,28 @@ which can be defined when the SDK's is built using |sdk_markup|:
...
+     - At :ref:`ref_workshop_launch`, :ref:`ref_workshop_refresh`:
+       after connecting plugs and slots
+       but before setting its status to *Ready*.
+
+     - Configures the user environment for the SDK to become operational.
+
 
 Each hook is defined in a text file of the same name
 under :samp:`hooks/` in the :ref:`source directory <ref_sdk_directory>`.
 At run-time, they are executed by :command:`bash`
```
akcano: :program:`bash` btw

docs/reference/sdks.rst
```diff
@@ -370,15 +371,28 @@ which can be defined when the SDK's is built using |sdk_markup|:
...
-as :samp:`root` inside the workshop.
+inside the workshop.
+Most hooks run as :samp:`root`,
+with the hook's parent directory as the working directory.
+The exception is :samp:`setup-project`,
+which runs as the :samp:`workshop` user
+in the :file:`/project` directory.
```
```suggestion
in the :file:`/project/` directory.
```

akcano (changes requested):
LGTM, a couple minor fixes.

jonathan-conder:
Thanks, I fixed those and also added an `$SDK` environment variable for all hooks. For example in `setup-project` I want to do something like `PATH="$SDK/bin:$PATH"`.

docs/reference/sdks.rst
```diff
@@ -370,15 +371,28 @@ which can be defined when the SDK's is built using |sdk_markup|:
      - At :ref:`ref_workshop_launch`, :ref:`ref_workshop_refresh`:
        after unpacking the base image
        and starting the workshop,
-       but before setting its status to *Ready*.
+       but before connecting plugs and slots.
```
dmitry-lyfar: It seems worth mentioning that the project directory is not mounted by the time `setup-base` is executed.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -1056,24 +1062,14 @@ write_files:
...
-    # Bypass confirmation prompts
-    APT::Get::Assume-Yes "1";
-  path: /etc/apt/apt.conf.d/01norecommend
 - content: |
     # Workaround for https://bugs.launchpad.net/snapd/+bug/2104066
     [Service]
     Environment=SNAPD_STANDBY_WAIT=1m
   path: /etc/systemd/system/snapd.service.d/override.conf
 runcmd:
+  # Project directory is required for 'workshop exec'.
```
dmitry-lyfar, line 1070: Not sure I understand when `/project` does not exist but `workshop exec` can be executed?

dmitry-lyfar (approved):

internal/workshop/lxd/lxd_backend.go
jonathan-conder: it's possible during `--wait-on-error`, if I recall correctly. I definitely triggered it by accident at the sprint

jonathan-conder:
@akcano you might want to see if you're happy with these changes to the hooks reference: https://github.com/canonical/workshop/compare/412b4b0a4442cb731fa848633b9110c736eac255..a78211c9747050e717e78a774b3dc27796765ebe

Also, I had to update some links due to the recent Snapcraft documentation updates. The missing piece is that https://snapcraft.io/docs/using-snapctl no longer exists outside the forum. I started a thread in the snap channel if you're interested. Otherwise I just set myself a reminder one month from now to check if it exists again.

docs/reference/sdks.rst
```diff
@@ -336,26 +337,24 @@ which can be defined when the SDK's is built using |sdk_markup|:
...
-   * - :samp:`restore-state`
+   * - :samp:`setup-project`
 
-     - At :ref:`ref_workshop_refresh`:
-       after launching the new workshop
-       and running the :samp:`setup-base` hook for the SDK.
+     - At :ref:`ref_workshop_launch`, :ref:`ref_workshop_refresh`:
+       after mounting the project directory
+       and connecting plugs and slots
+       but before setting its status to *Ready*.
```
```suggestion
       but before the workshop is set to *Ready*.
```

docs/reference/sdks.rst
```diff
@@ -365,20 +364,36 @@ which can be defined when the SDK's is built using |sdk_markup|:
...
 
 
 Each hook is defined in a text file of the same name
 under :samp:`hooks/` in the :ref:`source directory <ref_sdk_directory>`.
-At run-time, they are executed by :command:`bash`
-as :samp:`root` inside the workshop.
+At run-time, they are executed by :program:`bash`
+inside the workshop.
+Most hooks run as :samp:`root`,
+with the hook's parent directory as the working directory.
```
akcano: Needs clarification - is it `.workshop/<NAME>/hooks/` or `.workshop/<NAME>/`?

akcano (changes requested):

docs/reference/sdks.rst
jonathan-conder: Neither, since we're inside the workshop now. I tried to clarify a bit: https://github.com/canonical/workshop/compare/cb94189622ec21c53c012a1c327dc40c07177de0..62b4c785ab4db2502a9e9856f964291e07bcbc7a

akcano (approved):

PR #394, opened 2025-05-28
Fix issues related to partial SDK profile updates

jonathan-conder:
# Description

Works around the LXD issue where updating a profile takes effect even if an error occurs. In this case, we roll back the profile update. Other changes to the workshop (e.g. creating scripts in `/etc/profile.d`) are also rolled back if an error occurs later on.

Since it's now harder to get into an invalid state, we can be a bit more relaxed about creating directories before setting up a mount interface connection. I don't think the effects of this change are visible in practice though.

Disconnecting desktop and ssh-agent interfaces is now more relaxed, and no error occurs when removing a file which has already been deleted. On the other hand, if such files already exist when connecting the interface, an error occurs. This prevents connecting more than one plug of these types, which I think is what we want anyway.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/lxd_device/backend.go
+++ b/internal/interfaces/lxd_device/backend.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -14,6 +15,7 @@ import (
 
 	lxd "github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/x-go/strutil/shlex"
 
 	"github.com/canonical/workshop/internal/dirs"
 	"github.com/canonical/workshop/internal/interfaces"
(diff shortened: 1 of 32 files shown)


internal/revert/revert.go
```diff
@@ -19,6 +19,15 @@ func (r *Reverter) Add(f Hook) {
 	r.revertFuncs = append(r.revertFuncs, f)
 }
 
+// AddAll steals every revert function from another Reverter.
+func (r *Reverter) AddAll(other *Reverter) {
```
dmitry-lyfar: Probably: `CopyFrom`

internal/interfaces/lxd_device/backend.go
```diff
@@ -309,173 +433,184 @@ func sftpFs(conn lxd.InstanceServer, pid, w string) (workshop.WorkshopFs, error)
...
+
+	name := lxdbackend.ProfileName(sdkRef.ProjectId, sdkRef.Workshop, sdkRef.Sdk)
+	if err := conn.UpdateProfile(name, newp, etag); err != nil {
+		// By design, LXD does not roll back profile changes if the update failed,
+		// so we have to do it ourselves.
+		_, etag2, err2 := conn.GetProfile(name)
+		if err2 != nil {
+			logger.Noticef("cannot get updated profile: %v", err2)
+		} else if etag2 != etag {
+			if err2 := conn.UpdateProfile(name, oldp, etag2); err2 != nil {
```
dmitry-lyfar: It does not seem to differentiate whether the profile was created just yet (see the assignment above) or existed before. If the former, it should not `UpdateProfile` as it would leave an empty profile for the failed SDK. It has to unassign and remove instead, supposedly (so no empty profiles on a failed launch).

dmitry-lyfar: The above (an empty LXD profile left after a failed launch) leads to this behaviour:

1. Create two worktrees for the workshop repository.
2. Launch the dev workshop in the first one and stop (it has a tunnel from 8080).
3. Launch the second workshop.
4. Start the workshop from p.2. No conflicts reported, they both will start successfully. Both will show tunnel working from 8080 on the host.

Check `lxc profile show <profile-of-a-failed-sdk>` which would contain the redirect but note its "used" count in the `lxc profile list` (it is 0) which explains why this is possible.

dmitry-lyfar (commented):
Nice work. Left a note that I'd like to discuss further.

internal/revert/revert.go
jonathan-conder: how about `MoveFrom` or `StealFrom`?

internal/interfaces/lxd_device/backend.go
jonathan-conder: I think this isn't related to stop/start. First you have to launch 2 workshops to get a conflict. After the second launch fails, the profile still exists, and a retry succeeds (without having to stop the other workshop) because the existing profile won't be reattached to the instance.

When reading the old profile, we can detect whether it's unattached and just recreate it in that case. That way it doesn't really matter if we fail to clean up (unless we fail to detach the profile from the instance, but I don't see a way around that). However, we should attempt to clean up anyway.

internal/revert/revert.go
```diff
@@ -54,3 +45,12 @@ func (r *Reverter) Clone() *Reverter {
 
 	return rNew
 }
+
+// Copy adds the revert functions of the source reverter to the target reverter.
+func Copy(target, source *Reverter) {
+	if source == nil {
```
dmitry-lyfar, line 51: Should `target` be checked for nil too?

internal/interfaces/lxd_device/backend.go
```diff
@@ -43,101 +46,227 @@ func (b *Backend) Name() interfaces.SecuritySystem {
...
+		}
+	}
 
-		mounts, err := readMountProfile(fs)
+	var added []string
+	for _, mnt := range next {
+		if mnt == prev[mnt.Name] {
+			continue
+		}
+		found, err := addMountEntry(fs, user, mounts, mnt)
```
dmitry-lyfar: `found` does not seem to be a correct word here, as `addMountEntry` does look for an entry in fstab for WorkshopWorkshop mounts and returns false if it's found (e.g. not added to the mount entry).

`fstab`? `created`?

dmitry-lyfar (approved):

internal/revert/revert.go
jonathan-conder: If `target` is nil I think it's a bug, not much we can do except panic

internal/interfaces/lxd_device/backend.go
jonathan-conder: I'll go for `added`

jonathan-conder: Oh that's already in use. Modified then

jonathan-conder: But really this function is doing too many things, I'll break it up

internal/revert/revert.go
jonathan-conder: Currently it works the same as `maps.Copy` (panic if target is `nil` but do nothing if source is `nil`)

PR #396, opened 2025-06-04
Doc: Add purge how-to guide, LXD install updates

akcano:


docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+Always attempt this command first.
+If it completes successfully, your workshop should be purged.
+You can verify the outcome by running :command:`workshop list`.
+
+
+If standard procedure fails
+---------------------------
+
+You may need manual intervention if:
+
```
dmitry-lyfar, line 42: I think this whole guide should become a workshop's `remove` command option. E.g. `--purge` which would ignore errors such as the lack of the project directory. The described approach to purging is what `snap remove` going to do to any remaining workshop on the system.

All other cases except the known "project directory has been deleted" should be resolved as bugs, it is very possible that if a workshop is still running after remove, LXD commands will also fail. Ditto for other reasons in this list, they look like bugs or transient errors (i.e. LXD may fail to stop a container in which case remove should be repeated). 

@jonathan-conder thoughts?

jonathan-conder: I think `remove` should just remove everything by default. I think Yan's changes will take a step towards that. Maybe we can introduce `--purge` after we have a lifecycle system, i.e. `workshop remove` no longer removes everything associated to the workshop.

But also I think it's useful to record this information somewhere in the docs. I hope no one has to actually use it, but knowing the info is there might help create some goodwill.

akcano:
I suggest merging it as is, then dropping as soon as the `--purge` functionality comes along.

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+
+#. When you're sure, delete the orphaned custom volume:
+
+   .. code-block:: console
+
+      $ sudo lxc storage volume delete workshop <VOLUME>
+
+
+Aggressive cleanup
```
dmitry-lyfar, line 152: This section should be just `snap remove workshop --purge` which will execute the snap's remove hook to clean up everything. A user would then need to reinstall it. The instructions listed below will definitely break things for other workshops running (e.g. `rm -rf /var/snap/workshop/current/*` would kill the state file that maintains workshop connections).

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+
+  .. code-block:: console
+
+     $ sudo lxc profile delete --project workshop.<USERNAME> <PROFILE>
+
+
+LXD storage volumes
+```````````````````
+
+Workshops may use custom storage volumes
```
dmitry-lyfar: Some 'custom' LXD volumes are expected to have a longer-than-a-workshop lifecycle, e.g. SDKs. 

The workshop itself has a 'container' type volume associated with it. If one removes a workshop following the instructions above, they don't have to remove anything from here as LXD takes care of it (it is not possible to have a container removed AND its filesystem remaining in the list of volumes, LXD ties these two together behind the scenes and guarantees that if the container is gone, the volumes will be gone).

Therefore, in the context of purging workshops, this paragraph is not needed.

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+Clean up LXD resources
+~~~~~~~~~~~~~~~~~~~~~~
+
+Refer to the :ref:`how_troubleshoot_lxc` section in the troubleshooting guide
+for initial steps on listing and deleting orphaned LXD containers, e.g.:
+
+.. code-block:: console
+
+   $ sudo lxc list --all-projects | grep workshop.<USERNAME>
+   $ sudo lxc delete --project workshop.<USERNAME> <CONTAINER>
```
dmitry-lyfar: Also, 

```
   $ sudo lxc list --all-projects | grep workshop-stash.<USERNAME>
   $ sudo lxc delete --project workshop-stash.<USERNAME> <CONTAINER>
```
To ensure there are no backup copies of the workshop remaining.

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+its profile might remain.
+
+- List profiles for your workshop user project:
+
+  .. code-block:: console
+
+     $ sudo lxc profile list --project workshop.<USERNAME>
+
+
+- Inspect a specific profile (they are often named after the workshop or SDKs):
```
dmitry-lyfar: They are always named after SDKs, it's 1-1 relationship between an SDK and a LXD profile, so the name would be `<CONTAINER>-<SDK>`.

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+
+     $ sudo lxc list --project workshop.<USERNAME>
+
+
+  Then, for each container that should remain,
+  check its configuration to see which profiles it uses:
+
+  .. code-block:: console
+
+     $ sudo lxc config show --project workshop.<USERNAME> <CONTAINER>
```
dmitry-lyfar, line 133: This instruction won't work because the workshop is already removed by the time user gets here.

A profile can only be used by the workshop it is associated with. If a user removed the container following the instructions above, they can list orphaned profiles that require manual removal with this:
```
lxc profile list --project workshop.<USERNAME>
```

and check `USED BY` column in the output. Everything that has 0 can be safely removed.

docs/how-to/use-workshops/purge.rst
```diff
@@ -0,0 +1,289 @@
...
+
+Clean up LXD resources
+~~~~~~~~~~~~~~~~~~~~~~
+
+Refer to the :ref:`how_troubleshoot_lxc` section in the troubleshooting guide
+for initial steps on listing and deleting orphaned LXD containers, e.g.:
+
+.. code-block:: console
+
+   $ sudo lxc list --all-projects | grep workshop.<USERNAME>
```
dmitry-lyfar, line 80: Use `--force` flag for delete to account for cases when a container is still running.

docs/how-to/use-workshops/purge.rst
dmitry-lyfar: @akcano we discussed it further. It makes sense to disclose these details and we should use this how-to to explain how exactly things work to a developer. We believe that devs would want to know that and it is good to establish some trust with them if we are transparent about the implementation details. So, perhaps, this how-to should become part of the reference documentation in the future.

My point still stays, e.g. a product problem should not be fixed by the docs. I left my notes about this guide and happy to merge when these are fixed.

dmitry-lyfar (approved):
This looks good, although the generated doc does seem to include bits from the previous version (?)

Let's squash this fix up and see that is it correct.

PR #397, opened 2025-06-05
Fix plug binding issues

jonathan-conder:
# Description

Limits plug binding so you can only bind identical plugs in regular SDKs. The intended use case is to resolve conflicts between mount plugs with the same `workshop-target`, but it still works for other interfaces. Tunnel interface plugs can also conflict, but since they aren't auto-connected for regular SDKs, you can just leave one disconnected instead of binding them.

There's an outstanding issue where conflicts between unix domain sockets aren't detected. If you disconnect one plug, the socket is deleted even though the other plug remains connected. Since Workshop expands environment variables in socket paths it's not easy to detect these conflicts early on. I have a fix in mind and will do it as part of the partial profile update fix.

Fixes a few other bugs:
- Conflicts between `workshop-target`s are now detected before connecting the plugs (by moving all validation to a new kind of task).
- Plugs in different SDKs with the same name can be connected simultaneously (previously the LXD device names matched, so LXD would ignore all but one of them).
- Workshop definitions now support the shortcut `name: interface` for plugs, in addition to slots. SDK definitions already supported both.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/ifacestate/handlers.go
+++ b/internal/overlord/ifacestate/handlers.go
@@ -198,6 +198,7 @@ func (m *InterfaceManager) connectAuto(task *state.Task, wp *workshop.Workshop, 
 			// its master.
 			continue
 		}
+
 		candidates := m.repo.AutoConnectCandidateSlots(info.ProjectId, info.Workshop,
 			info.Name, plug.Name, autoConnectChecker(wconns))
 
@@ -207,6 +208,11 @@ func (m *InterfaceManager) connectAuto(task *state.Task, wp *workshop.Workshop, 
 				slotDynamic[connRef.ID()] = make(map[string]interface{})
 			}
 
+			// remounts may be not nil when a previously existing mount
+			// interface connection (e.g. pre-refresh) needs to be recreated
+			// without changes to its 'source' attribute in the new workshop
+			// (given the new workshop also has an SDK with exactly the same
+			// plug; the target directory may change in the new workshop).
 			if src, ok := remounts[connRef.ID()]; ok {
 				slotDynamic[connRef.ID()]["host-source"] = src
 			}
@@ -235,6 +241,7 @@ func (m *InterfaceManager) connectAuto(task *state.Task, wp *workshop.Workshop, 
 	for _, slot := range info.Slots {
 		candidates := m.repo.AutoConnectCandidatePlugs(info.ProjectId, info.Workshop,
 			info.Name, slot.Name, autoConnectChecker(wconns))
+
 		for _, plug := range candidates {
 			ref := plug.Ref()
 			master, slaves := MaybeBound(wp, ref)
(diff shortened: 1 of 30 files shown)


akcano (approved):

internal/overlord/ifacestate/handlers.go
```diff
@@ -195,6 +208,11 @@ func (m *InterfaceManager) connectAuto(task *state.Task, wp *workshop.Workshop,
 				slotDynamic[connRef.ID()] = make(map[string]interface{})
 			}
 
+			// remounts may be not nil when a previously existing mount
```
dmitry-lyfar, line 211: Not sure I understand/remember the purpose of my old comment here. Even if `remounts` is nil it would work fine.

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -420,31 +438,43 @@ func (m *InterfaceManager) resolveWorkshopConnections(w *workshop.Workshop) erro
 	return nil
 }
 
-func (m *InterfaceManager) checkConflictingTargets(sdkInfo *sdk.Info) error {
-	allPlugs := m.repo.AllPlugs("mount")
+func (m *InterfaceManager) checkConflictingTargets(w *workshop.Workshop) error {
```
dmitry-lyfar:
```suggestion
func (m *InterfaceManager) checkForConflictingMounts(w *workshop.Workshop) error {
```
?

internal/overlord/ifacestate/handlers.go
```diff
@@ -21,6 +21,37 @@ import (
 	"github.com/canonical/workshop/internal/workshop"
 )
 
+func (m *InterfaceManager) doValidateInterfaces(task *state.Task, tomb *tomb.Tomb) (err error) {
```
dmitry-lyfar: Not insisting on changing the name here, but here is what I think: the function may still return an error even if all the interfaces are _valid_ as it also resolves conflicts and binding names. So, should it be smth. like `doResolveInterfaces`?

dmitry-lyfar (approved):

internal/overlord/ifacestate/handlers.go
jonathan-conder: I think it's just explaining "these are the conditions under which remounts might be relevant"

internal/overlord/ifacestate/ifacemgr.go
jonathan-conder: I think `Mounts` is clearer but `For` isn't consistent with other `check*` functions

dmitry-lyfar: That sounds good, yes.

internal/overlord/ifacestate/handlers.go
jonathan-conder: I think that's better. If we could have something like `doResolveInterfaceRelationships` without being too much of a mouthful I think that covers all 3 steps. But I can't think of anything.

PR #398, opened 2025-06-06
Rework LxdProjectName to accommodate non-POSIX usernames

dmitry-lyfar:
# Description

If a username is not POSIX-compliant, it can cause issues. For example, including an '@' character in the username breaks snapshot logic, as the LXD snapshot mechanism interprets '@' as a separator when using the project name. To prevent such conflicts, only alphanumeric usernames are allowed. If the username contains unsupported characters, the user's UID will be used as the LXD project name suffix instead.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/workshopstate/manager.go
+++ b/internal/overlord/workshopstate/manager.go
@@ -63,8 +63,8 @@ func (w *WorkshopManager) Workshop(ctx context.Context, name, pId string) (*work
 	return workshop, nil
 }
 
-// Returns latest file for a workshop. The state must be locked,
-// as listing projects can update project metadata.
+// Returns latest file for a workshop. The state must be locked, as listing
+// projects can update project metadata.
 func (w *WorkshopManager) WorkshopFile(ctx context.Context, name, pId string) (*workshop.File, error) {
 	user, ok := ctx.Value(workshop.ContextUser).(string)
 	if !ok {
@@ -85,8 +85,8 @@ func (w *WorkshopManager) WorkshopFile(ctx context.Context, name, pId string) (*
 	return p.Workshop(name)
 }
 
-// Returns all workshop files for a project. The state must be locked,
-// as listing projects can update project metadata.
+// Returns all workshop files for a project. The state must be locked, as
+// listing projects can update project metadata.
 func (w *WorkshopManager) WorkshopFiles(ctx context.Context, pId string) (map[string]string, error) {
 	user, ok := ctx.Value(workshop.ContextUser).(string)
 	if !ok {
--- a/internal/overlord/workshopstate/request.go
+++ b/internal/overlord/workshopstate/request.go
@@ -317,7 +317,7 @@ func installSdks(st *state.State, pid, w string, sdks []sdk.Setup, retrieveTasks
 	return all
 }
 
-func launchWorkshop(st *state.State, file *workshop.File, project workshop.Project) *state.TaskSet {
+func launchWorkshop(st *state.State, file *workshop.File) *state.TaskSet {
 	construct := state.NewTaskSet()
 
 	var prev *state.Task
@@ -397,7 +397,7 @@ func launch(st *state.State, file *workshop.File, sdks []sdk.Setup, project work
 	createDirs := st.NewTask("create-workshop-storage", fmt.Sprintf("Create %q storage directories", file.Name))
 	addTaskSet(state.NewTaskSet(createDirs))
 
-	create := launchWorkshop(st, file, project)
+	create := launchWorkshop(st, file)
 	addTaskSet(create)
 
 	install := installSdks(st, project.ProjectId, file.Name, sdks, rmap)
@@ -472,7 +472,7 @@ func (w *WorkshopManager) RefreshMany(ctx context.Context, projectId string, nam
 			return nil, err
 		}
 
-		tasks, err := refresh(ctx, w.state, plan, wp, req.file)
+		tasks, err := refresh(w.state, plan, wp, req.file)
 		if err != nil {
 			return nil, fmt.Errorf("cannot refresh %q: %w", name, err)
 		}
@@ -637,7 +637,7 @@ func resolveRefresh(w *workshop.Workshop, newfile *workshop.File, candidates []s
 	return plan, nil
 }
 
-func refresh(ctx context.Context, st *state.State, plan *refreshPlan, w *workshop.Workshop, file *workshop.File) (*state.TaskSet, error) {
+func refresh(st *state.State, plan *refreshPlan, w *workshop.Workshop, file *workshop.File) (*state.TaskSet, error) {
 	refresh := state.NewTaskSet()
 	prev := (*state.TaskSet)(nil)
 	addTaskSet := func(ts *state.TaskSet) {
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -113,22 +113,26 @@ func New() (*Backend, error) {
 		imageServer = srv
 	}
 
-	// Create LXD storage pool if it doesn't exist
+	// TODO: run this logic for a specific user. The code below implies the
+	// default project activated for the connection. As we have seen, every user
+	// has to create its own storage pool to avoid issues with id mapping of a
+	// volume with the same name (e.g. both users have system-1 volume for the
+	// system SDK that cannot be successfully mounted for another user).
 	conn, err := lxd.ConnectLXDUnix("", nil)
 	if err != nil {
 		return nil, ErrorLxdBackend(err)
 	}
 	defer conn.Disconnect()
 
+	// Create LXD storage pool if it doesn't exist.
 	pools, err := conn.GetStoragePools()
 	if err != nil {
 		return nil, err
 	}
-	// Workshop does not require an existing storage pool.
-	// However, once the workshop storage pool exists,
-	// `lxd init --auto` won't add another one,
-	// and non-Workshop LXD containers can't be launched
-	// without further manual configuration.
+	// Workshop does not require an existing storage pool. However, once the
+	// workshop storage pool exists, `lxd init --auto` won't add another one,
+	// and non-Workshop LXD containers can't be launched without further manual
+	// configuration.
 	if len(pools) == 0 {
 		return nil, errors.New(`LXD not initialized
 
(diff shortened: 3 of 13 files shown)


internal/workshop/lxd/lxd_backend_project.go
```diff
@@ -25,41 +28,82 @@ func lxdProjectConfig(username string) map[string]string {
 	}
 }
 
-func LxdProjectName(user string) string {
-	return "workshop." + user
+// Checks if a user name can be used in a LXD project name.
+var isValidUsername = regexp.MustCompile(`^[a-zA-Z0-9-._]*$`).MatchString
```
jonathan-conder: I think we should require it to start with `[a-zA-Z]` to avoid purely numeric usernames. The `-` should probably be right before the `]` so it doesn't look like `9-.`.

internal/workshop/lxd/lxd_backend_project.go
```diff
@@ -25,41 +28,82 @@ func lxdProjectConfig(username string) map[string]string {
...
+				Config:      lxdProjectConfig(username),
+				Description: fmt.Sprintf(`Workshop stash project for "%s" user`, username),
+			},
+			Name: stash,
+		})
+
+		if err = cmp.Or(err1, err2); err != nil {
+			return err
+		}
+		rev.Success()
```
jonathan-conder: There's no `defer revert.Fail()`. Also I don't think we need to proceed with creating the stash if creating the project failed. But we should probably ensure the stash exists even if the project already exists.

internal/overlord/workshopstate/manager.go
```diff
@@ -63,8 +63,8 @@ func (w *WorkshopManager) Workshop(ctx context.Context, name, pId string) (*work
 	return workshop, nil
 }
 
-// Returns latest file for a workshop. The state must be locked,
-// as listing projects can update project metadata.
+// Returns latest file for a workshop. The state must be locked, as listing
+// projects can update project metadata.
```
jonathan-conder, line 67: not a review comment, just curious what format comments should be in?

internal/overlord/workshopstate/request.go
```diff
@@ -317,7 +317,7 @@ func installSdks(st *state.State, pid, w string, sdks []sdk.Setup, retrieveTasks
 	return all
 }
 
-func launchWorkshop(st *state.State, file *workshop.File, project workshop.Project) *state.TaskSet {
+func launchWorkshop(st *state.State, file *workshop.File) *state.TaskSet {
```
jonathan-conder, line 320: Sorry that one was my fault, need to get round to adding a lint for this.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -926,20 +935,22 @@ func (s *Backend) WorkshopFs(ctx context.Context, name string) (workshop.Worksho
...
+	if err != nil {
+		return nil, ErrorLxdBackend(err)
 	}
+
+	return switchLxdProject(conn, user)
+}
+
+func (s *Backend) LxdClient(ctx context.Context) (lxd.InstanceServer, error) {
+	return ConnectLxd(ctx)
 }
```
jonathan-conder, line 966: I'm not sure this needs to exist. But I guess we might add fields to the `Backend` struct at some point. In that case I think `lxd_device` should always use `ConnectLxd`. There's one exception: https://github.com/canonical/workshop/blob/33fa7949a7a4391ab0d6c38d4bc420a8abc653c9/internal/interfaces/lxd_device/tests/integration/backend_test.go#L108-L111
Also `lxd` should use `LxdClient` (see below).

internal/workshop/lxd/lxd_backend_stash.go
```diff
@@ -45,7 +44,16 @@ func (s *Backend) StashWorkshop(ctx context.Context, name string) error {
...
 
-	if err = s.copyInstance(conn, instance, stashedInsance, LxdProjectName(user), LxdStashProjectName(user)); err != nil {
+	instance := InstanceName(name, projectId)
+	stashed := instanceStashName(name, projectId)
+
+	sourceProject, err1 := lxdProjectName(user)
+	targetProject, err2 := lxdStashProjectName(user)
+	if err = cmp.Or(err1, err2); err != nil {
+		return err
+	}
```
jonathan-conder, line 54: I didn't know about `cmp.Or` and I think it's useful e.g. when cleaning up, but in this case I don't think it's necessary. Even though two `if err != nil` blocks is longer I think it's easier to read somehow.

internal/workshop/lxd/tests/integration/project_test.go
```diff
@@ -343,3 +346,45 @@ func (f *wsProject) TestLxdBackendLoadProjectAsDifferentUser(c *check.C) {
...
+		}
+		return &user.User{
+			Username: "ubuntu",
+			Uid:      "1001",
+		}, nil
+	})
+	defer restore()
+	ctx := context.WithValue(context.Background(), workshop.ContextUser, "ubuntu")
+
+	p, err := lxdbackend.ConnectLxd(ctx)
```
jonathan-conder:
```suggestion
	be := lxdbackend.Backend{}
	p, err := be.LxdClient(ctx)
```
See above. I guess these ones are fine, I just think that if `lxd_device` was removed there would be no need for a `ConnectLxd` function, so we should only use it in `lxd_device`.

internal/workshop/lxd/tests/integration/project_test.go
```diff
@@ -343,3 +346,45 @@ func (f *wsProject) TestLxdBackendLoadProjectAsDifferentUser(c *check.C) {
...
+		}
+		return &user.User{
+			Username: "ubuntu@canonical.com",
+			Uid:      "1001",
+		}, nil
+	})
+	defer restore()
+	ctx := context.WithValue(context.Background(), workshop.ContextUser, "ubuntu@canonical.com")
+
+	p, err := lxdbackend.ConnectLxd(ctx)
```
jonathan-conder:
```suggestion
	be := lxdbackend.Backend{}
	p, err := be.LxdClient(ctx)
```

jonathan-conder (changes requested):

internal/overlord/workshopstate/manager.go
dmitry-lyfar: I use Rewrap extension for VS Code that ensures hard wrapping at 80 symbols. We should probably standardise that in the contribution guide.

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: I only left it as the original implementation had an idea of a connection pool as a property of the backend struct. It is not required / meaningful for the unix socket connections in our context, so no pool and, thus, it does not look required.

internal/workshop/lxd/lxd_backend_project.go
dmitry-lyfar: It should probably be named differently as its purpose is not to validate a user name, but rather ensure it contains alphanum + "-._" symbols only that we are willing to accept for the suffix. All else must fall back to using a user's ID. What's the issue with fully numeric usernames?

internal/workshop/lxd/lxd_backend.go
jonathan-conder: We could possibly attach the user data to the backend struct. Not sure how that would look but it might be a reason to keep it around.

internal/workshop/lxd/lxd_backend_project.go
jonathan-conder: A user could spy on another user's workshops by pretending their username is, say 1000. I'm not sure it's possible to trick the daemon into reading a fake username without `root`, but I don't think it's worth risking it

jonathan-conder: In any case fully numeric usernames are banned in Ubuntu according to `useradd(8)`

internal/workshop/lxd/lxd_backend_stash.go
dmitry-lyfar: Yes, my purpose here was to try it, really struggled with verbosity lately. I think we should at least use it in clean ups and used it in this PR to agree on using it consistently. I learnt it here:

```
func printSum(a, b string) error {
    x, err1 := strconv.Atoi(a)
    y, err2 := strconv.Atoi(b)
    if err := cmp.Or(err1, err2); err != nil {
        return err
    }
    fmt.Println("result:", x+y)
    return nil
}
```

https://go.dev/blog/error-syntax

dmitry-lyfar: It's not about verbosity of error handling per se for me but rather verbosity in independent auxiliary operations that may produce errors (cleanup is a good example, but it's not the only case).

jonathan-conder: Ah interesting. I read that post but, in contrast to our tutorial, I completely ignored the code and only read the text.

My only real objection is that `cmp.Or` was surprising at first. If it becomes more common then I don't see an issue using it here too. Just have to avoid using it if the second operation depends on the first one succeeding.

dmitry-lyfar: Yes, I think it is unconventional which makes it a bit harder to read at first. I suggest we try it in independent operations and in clean ups for now to see how that unfolds. The project's example was to show how (in my opinion) revert logic reads clearer when there is no immediate error handling for the op, not the best use case ofc as these are dependent, agree.

jonathan-conder: They look independent to me, I don't think there are any side effects

dmitry-lyfar: Sorry, I mean the use case with creating LXD projects in `initLxdProject()`.

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: User data = user name <-> project relationship?

jonathan-conder: Could just be the name, or name + UID. Not really something I would do now, but we could leave open the possibility of the Backend having some data attached.

jonathan-conder (approved):

PR #406, opened 2025-06-17
Doc: Add architecture section

akcano:


docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+with :program:`workshopd` serving as the central component
+that interfaces with LXD for container management.
+This design ensures proper security boundaries
+while keeping a responsive user experience.
+
+
+Prerequisites
+-------------
+
+|ws_markup| is designed to run on Linux systems
```
dmitry-lyfar:
```suggestion
|ws_markup| is designed to run on Linux systems
```
This is already said in the first sentence of this document.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+- :program:`workshopd`, the main daemon that manages workshop life cycles
+- :program:`workshopctl`, an internal control tool for workshop management
+
+
+Each one has a specific role in |ws_markup|'s operation and security model.
+
+The system uses Unix domain sockets for inter-process communication,
+with :program:`workshopd` serving as the central component
+that interfaces with LXD for container management.
+This design ensures proper security boundaries
```
dmitry-lyfar:
```suggestion
```
I'm not sure I can pinpoint the exact reason for this design, I would suggest not to explain it here. We wanted a client-server for multiple reasons at the time, e.g. project directory tracking, lean clients, clearer separation of the logic, etc.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+|ws_markup| is designed to run on Linux systems.
+It is primarily distributed as a snap
+and relies on LXD as its containerization backend
+and ZFS for efficient storage management.
+
+It follows a client-server architecture with three main components:
+
+- :program:`workshop`, the CLI tool for user interaction
+- :program:`workshopd`, the main daemon that manages workshop life cycles
+- :program:`workshopctl`, an internal control tool for workshop management
```
dmitry-lyfar: This tool's primary purpose is to communicate with the daemon from within a workshop container, `workshopd` manages workshops, `workshopctl` does not.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+that interfaces with LXD for container management.
+This design ensures proper security boundaries
+while keeping a responsive user experience.
+
+
+Prerequisites
+-------------
+
+|ws_markup| is designed to run on Linux systems
+and relies on LXD as its containerization backend
```
dmitry-lyfar:
```suggestion
and relies on LXD as its containerization backend
```
This is also said in the previous paragraph.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+  are essential for storage operations.
+
+- You must have a compatible Linux kernel with necessary modules loaded
+  (this is handled by the :samp:`zfsutils-linux` package and LXD),
+  sufficient disk space for the ZFS storage pool,
+  and appropriate system permissions
+  for container operations.
+
+
+Main components
```
dmitry-lyfar, line 12: I think this paragraph repeats and then enhances the intro paragraph. One of them must go.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+Each one has a specific role in |ws_markup|'s operation and security model.
+
+The system uses Unix domain sockets for inter-process communication,
+with :program:`workshopd` serving as the central component
+that interfaces with LXD for container management.
+This design ensures proper security boundaries
+while keeping a responsive user experience.
+
+
+Prerequisites
```
dmitry-lyfar: This section should not be a part of the reference architecture. It is in the tutorial already. It should be enough to state external dependencies (which you do above and then later on in this document).

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+  and appropriate system permissions
+  for container operations.
+
+
+Main components
+---------------
+
+|ws_markup| installation on your host system includes three primary components:
+
+- The :program:`workshop` CLI serves as the main user interface,
```
dmitry-lyfar, line 17: The client is so thin so it does not handle definition parsing. It works with `workshopd` API through a Unix domain socket connection.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+
+- The :program:`workshopd` daemon runs as a background process
+  with elevated privileges,
+  managing the complete workshop life cycle
+  including container creation,
+  SDK installation,
+  interface coordination
+  and state persistence through a REST API.
+
+- The :program:`workshopctl` tool operates with reduced privileges
```
dmitry-lyfar: Suggestion:

`workshopctl` runs in a workshop and has access to a very limited subset of  `workshopd` REST API for service and reporting tasks.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+
+The :program:`workshopd` daemon integrates with :program:`systemd`
+for process management and socket activation,
+implementing the :program:`systemd` watchdog protocol
+to ensure process health monitoring.
+When running under :program:`systemd`,
+:program:`workshopd` enters a degraded mode if system health checks fail,
+retrying the checks every 5 minutes while continuing to serve requests.
+
+The :program:`workshopd` daemon also supports graceful shutdown handling
```
dmitry-lyfar: I would not mention this in the architecture description as it is an expected baseline behaviour for any Linux program.

dmitry-lyfar: I would also not mention socket activation when talking about architecture.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+Daemon
+~~~~~~
+
+The :program:`workshopd` is the core component of |ws_markup|
+responsible for managing the complete workshop life cycle:
+it talks to LXD through its native API
+to handle user-specific LXD projects
+and ensure proper isolation between different users' workshops.
+
+The :program:`workshopd` daemon integrates with :program:`systemd`
```
dmitry-lyfar: This paragraph should emphasise the following:

> workshopd is a systemd service responsible for the complete 
> workshop life cycle. At its core, it relies on the proven state package 
> from snapd to ensure that any changes to a workshop container are handled transactionally—safe, consistent, and reversible.

dmitry-lyfar: Also, no need to talk about LXD communication as you describe this in the following paragraph.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,401 @@
...
+
+    flowchart LR
+        user([User])
+
+        subgraph HOST [Host system]
+            direction TB
+            cli([CLI])
+            daemon(["workshopd"])
+            lxd([LXD daemon])
+            zfs_utils(["zfsutils-linux<br>(indirect dependency)"])
```
dmitry-lyfar: hmm.. I'm not sure where you are coming from with `zfsutils`. We don't interact with this package directly nor we depend on it. It is solely a LXD dependency that we never touch as we work with LXD API to manage ZFS volumes and snapshots. At this scale, it is probably easier to talk about LXD zfs storage driver that handles all this and has these utils as a dependency.

dmitry-lyfar (commented):
Thanks! I commented only on the introduction part so far, let's resolve these and have another pass on the remaining document (I read it all, but there is quite a bit I would like to change; mostly likely because we understand the target reader & whats important in the architecture differently -- I'll help with this and with the details that I think worth exposing as we go).

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+or if the pool is missing,
+it instructs LXD to create a dedicated ZFS storage pool named :samp:`workshop`
+if one doesn't already exist with that name.
+LXD then handles the creation and management of this pool,
+which will be used for all workshop containers and their data.
+
+The creation process includes intelligent sizing logic
+to prevent common space exhaustion issues.
+When LXD's default allocation would result in a pool smaller than 5 GB,
+|ws_markup| automatically overrides the size to ensure at least 5 GB is available.
```
dmitry-lyfar:
```suggestion
Workshop requires a minimum pool size of 5 GiB.
```

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+Storage pool
+~~~~~~~~~~~~
+
+|ws_markup| leverages ZFS for its storage needs, managed via LXD.
+When :program:`workshopd` starts for the first time,
+or if the pool is missing,
+it instructs LXD to create a dedicated ZFS storage pool named :samp:`workshop`
+if one doesn't already exist with that name.
+LXD then handles the creation and management of this pool,
+which will be used for all workshop containers and their data.
```
dmitry-lyfar:
```suggestion
```

docs/explanation/architecture/installation.rst
dmitry-lyfar: The ZFS pool must be named `workshop`. To switch Workshop to use a different (non-default) ZFS pool, follow these steps:

Note: All running Workshop containers must be removed and relaunched after switching the storage pool.
```
lxc storage delete workshop
lxc storage create workshop zfs source=my-pool/workshop
snap restart workshop
```

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+|ws_markup| establishes a dedicated network infrastructure
+through the :samp:`workshopbr0` bridge network,
+providing isolated networking for workshop containers.
+
+This bridge network includes DNS resolution
+configured with the :samp:`workshop` domain.
+The network configuration supports port forwarding capabilities
+for exposing workshop services to the host system
+and implements proper network isolation
+to prevent unauthorized cross-workshop communication.
```
dmitry-lyfar:
```suggestion
```
Probably, not here. Port forwarding happens at a different level and maintained by LXD via something called proxy-device.

dmitry-lyfar: Instead:

One can use workshop's container DNS name from host:
```
# Obtain bridge IP address that works as a DNS resolver
$ ip -brief addr show workshopbr0
workshopbr0      UP             10.157.98.1/24 fd42:ace5:e936:8549::1/64 fe80::216:3eff:fe54:5e23/64
$ dig @10.157.98.1 <lxd-container-name>.workshop
```

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+(:samp:`workshop-stash.<USERNAME>`)
+used for temporary storage during workshop rebuild operations.
+
+Note that this LXD project name is different from the workshop project ID:
+the LXD project name (:samp:`workshop.<USERNAME>`) provides user-level isolation,
+while each workshop project gets its own unique ID for tracking and management.
+
+This project-based isolation ensures that users cannot access
+each other's workshops, volumes, or configuration data,
+even when multiple users share the same host system.
```
dmitry-lyfar:
```suggestion
```

The reason for multiuser support is not protecting other persons stuff at this level, as if I'm part of `lxd` group, I would see everything. It is mostly done for id mapping. So, if you have multiple users running workshops, each user would have it's own ipmap configuration for the `workshop` user as opposed if we did not separate based on users, there could be a situation when/if a workshop used by multiple users, it would have permissions / ownership issues.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+
+The networking layer integrates with |ws_markup|'s interface system
+to provide controlled resource sharing between containers and the host.
+This includes X11 forwarding for graphical applications,
+SSH agent forwarding for authentication,
+and custom proxy configurations for specific workshop requirements.
+
+The network isolation ensures that workshops operate in separate namespaces
+while still allowing necessary communication channels
+as defined by the workshop's interface specifications.
```
dmitry-lyfar:
```suggestion

```

No, not necessarily.

The proxy device in LXD does not inherently rely on the bridge interface. Instead, it creates its own network forwarding rules (using iptables, nftables, or socat) to forward traffic from a host address/port to a container address/port. Often, a tunnel interface connection would not use networking directly (unless you bind to something like `0.0.0.0` on the host side).

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,339 @@
...
+
+            subgraph ZFS_Store [ZFS storage]
+                zfs_pool[("ZFS pool<br>(workshop)")]
+            end
+        end
+
+
+
+        subgraph EXT_IMG [Image server]
+            img_server[("cloud-images.ubuntu.com<br>or a custom server")]
```
dmitry-lyfar:
```suggestion
            img_server[("cloud-images.ubuntu.com")]
```

dmitry-lyfar (commented):
Looks much better, I left some comments.

docs/explanation/architecture/installation.rst
jonathan-conder: In our case the proxy is always "like socat." Workshops do live in separate network namespaces, but LXD creates paired virtual ethernet (`veth`) devices which allows anything on the LAN to pierce the sandbox if it knows the IP address and has appropriate routing set up.

docs/explanation/architecture/installation.rst
akcano: I'll add these steps to a ZFS subsection in 'Troubleshooting' instead.

docs/explanation/architecture/installation.rst
akcano: These commands would suit the troubleshooting section better as well, I think. Probably a dedicated 'Networking' subsection?

docs/explanation/architecture/installation.rst
akcano: Again, these details would fit the Troubleshooting section better. I'll make a note.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+
+The :program:`workshopctl` tool serves as a secure bridge
+between workshop containers and the host system.
+It operates with the workshop user's permissions (UID 1000)
+and communicates with :program:`workshopd`
+through a dedicated Unix domain socket
+at :file:`/var/lib/workshop/run/workshop.socket.untrusted`.
+
+When a workshop needs to execute commands or access host resources,
+:program:`workshopctl` handles the communication.
```
dmitry-lyfar:
```suggestion
```
Not correct (wokshopctl does not handle host resources or command execution)

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+Each workshop session is assigned a unique context ID,
+which :program:`workshopctl` uses to ensure
+that commands are executed in the correct environment.
+The tool manages command execution,
+including input or output streams and exit codes,
+while maintaining the workshop's security.
```
dmitry-lyfar:
```suggestion
```
I would drop it as context ID is a very implementation specific detail unrelated to choosing a correct environment (it's about dispatching and identifying which SDK called a given hook) as well as 
printing input and output streams, this is just an expected program behaviour. Also, workshopctl should not rely on the context ID (it's a hook specific thing) and should execute from within workshopctl normally (which it doesn't right now because there are no relevant commands to have outside of hooks contexts, but will be for `workshop scripts`).

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+
+The communication architecture between these components uses a layered approach
+where the CLI communicates with :program:`workshopd` via Unix domain sockets,
+while :program:`workshopd` interfaces with LXD through the latter's native API.
+
+
+CLI
+~~~
+
+The :command:`workshop` CLI provides a comprehensive command-line interface
```
dmitry-lyfar:
```suggestion
The :program:`workshop` CLI provides a comprehensive command-line interface
```
?

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+with proper error reporting and progress tracking.
+
+
+Daemon
+~~~~~~
+
+The :program:`workshopd` is the core component of |ws_markup|
+responsible for managing the complete workshop life cycle:
+it talks to LXD through its native API
+to ensure proper isolation between different users' workshops.
```
dmitry-lyfar:
```suggestion
responsible for managing the complete workshop life cycle. It uses LXD as its container backend.
```
We discuss isolation below (in the LXD section) and in this section I suppose it is important to state that workshopd uses LXD as a container backend.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+The :program:`workshopd` daemon is a :program:`systemd` service
+responsible for the complete workshop life cycle.
+At its core, it relies on the proven state package from :program:`snapd`
+to ensure that any changes to a workshop are handled in a transactional manner:
+safely, consistently, and reversibly.
+
+When integrated with :program:`systemd`,
+:program:`workshopd` implements the watchdog protocol for health monitoring,
+entering a degraded mode if system health checks fail
+while continuing to serve requests and retrying the checks every 5 minutes.
```
dmitry-lyfar:
```suggestion
When integrated with :program:`systemd`,
:program:`workshopd` implements the watchdog notification mechanism for health monitoring and supports socket activation to reduce memory footprint.
```
We can do something like this, but still it feels like a small detail for the architecture topic (we talk about relationships between components in the previous paragraph and suddenly switch to discussing the details of integration with systemd which may change or go away and that won't affect how these components work together).

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+~~~~~~~~~~~~
+
+|ws_markup| leverages ZFS for its storage needs, managed via LXD.
+
+Workshop requires a minimum pool size of 5 GiB.
+
+The ZFS pool serves multiple purposes beyond simple container storage,
+managing container root file systems,
+custom volumes for SDK data persistence,
+snapshot storage for workshop state management,
```
dmitry-lyfar:
```suggestion
snapshots for quick workshop updates and rollbacks,
```

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,314 @@
...
+Storage pool
+~~~~~~~~~~~~
+
+|ws_markup| leverages ZFS for its storage needs, managed via LXD.
+
+Workshop requires a minimum pool size of 5 GiB.
+
+The ZFS pool serves multiple purposes beyond simple container storage,
+managing container root file systems,
+custom volumes for SDK data persistence,
```
dmitry-lyfar:
```suggestion
volumes for SDKs and SDK data persistence,
```

docs/explanation/architecture/installation.rst
akcano: This is a borderline case, but you're right.

docs/explanation/architecture/installation.rst
```diff
@@ -0,0 +1,302 @@
...
+Each user project includes a corresponding stash project
+(:samp:`workshop-stash.<USERNAME>`)
+used for temporary storage during workshop rebuild operations.
+
+Note that this LXD project name is different from the workshop project ID:
+the LXD project name (:samp:`workshop.<USERNAME>`) provides user-level isolation,
+while each workshop project gets its own unique ID for tracking and management.
+
+
+The project initialization process includes setting up appropriate ID mapping,
```
dmitry-lyfar:
```suggestion
The workshop launch includes setting up appropriate ID mapping,
```
The project initialization does not include ID mapping setting. It only happens when a workshop is launched at the container-level.

dmitry-lyfar (approved):
One more thing, but otherwise looks good.

PR #407, opened 2025-06-19
Temporarily restrict unix socket paths

jonathan-conder:
# Description

This works around https://github.com/canonical/lxd/issues/15782, which causes LXD to delete the host's wayland socket when disconnecting the desktop interface. The fix is to move the socket elsewhere and create a symlink in the usual location.

Also fixes a couple of other issues I noticed in the desktop interface:
- it now supports absolute paths in `$WAYLAND_DISPLAY`
- it's more explicit about not supporting certain forms of `$DISPLAY`
- socket paths inside the workshop are fixed and can't be influenced by the user environment
- the `XAuthority` mount is now named using the same scheme as other devices (`<SDK>_<PLUG>`)

When LXD 6.5 is available we should revert 73fdf90 but keep the other commits.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/builtin/desktop.go
+++ b/internal/interfaces/builtin/desktop.go
@@ -75,29 +75,31 @@ func (iface *desktopInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInf
 }
 
 func (iface *desktopInterface) MountConnectedPlug(spec *lxd_device.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	xdg := spec.Environment["XDG_RUNTIME_DIR"]
-	if xdg == "" {
-		return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", spec.User.Username)
-	}
-
 	desktop := workshop.Desktop{}
 
 	wayland := spec.Environment["WAYLAND_DISPLAY"]
 	display := spec.Environment["DISPLAY"]
-
 	if wayland == "" && display == "" {
 		return fmt.Errorf("neither DISPLAY nor WAYLAND_DISPLAY are set for user %q", spec.User.Username)
 	}
 
 	if wayland != "" {
+		if !filepath.IsAbs(wayland) {
+			xdg := spec.Environment["XDG_RUNTIME_DIR"]
+			if xdg == "" {
+				return fmt.Errorf("XDG_RUNTIME_DIR is either empty or unset for user %q", spec.User.Username)
+			}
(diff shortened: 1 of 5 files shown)


internal/interfaces/builtin/desktop.go
```diff
@@ -24,6 +24,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/canonical/lxd/shared/logger"
```
dmitry-lyfar:
```suggestion
	"github.com/canonical/workshop/internal/logger"
```
?

internal/interfaces/builtin/desktop.go
```diff
@@ -75,29 +77,31 @@ func (iface *desktopInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInf
...
 		desktop.Wayland = &workshop.ProxyEntry{
 			Name: fmt.Sprintf("%s_wayland", plug.Name()),
 			Connect: workshop.ProxyTarget{
-				Address:  filepath.Join(xdg, wayland),
+				Address:  wayland,
 				Protocol: "unix",
 			},
 			Listen: workshop.ProxyTarget{
-				Address:  filepath.Join(dirs.XdgRuntimeDirBase, workshop.User.Uid, wayland),
+				Address:  filepath.Join("/tmp", "wayland-0"),
```
dmitry-lyfar, line 103: I remember there was some problem with, at least, `/run` in the container due to the mounting race so that LXD would mount `tmpfs` on top which would conceal the socket (e.g. on a workshop's restart).  I'm going to smoke test that.

internal/interfaces/builtin/desktop.go
jonathan-conder: nice catch, I blame `gopls` but will need to keep a closer eye on that sort of thing

internal/interfaces/builtin/desktop.go
jonathan-conder: I think `/tmp` should be OK since that's where the X11 socket is. Maybe we can put it in a subdirectory just in case.

jonathan-conder: Although I don't see a reason why I didn't just put it in `/var/lib/workshop/run`

dmitry-lyfar: Yes, `/var/lib/workshop/run` works I suppose.

dmitry-lyfar (approved):

PR #415, opened 2025-06-25
Doc: Fix spellchecker

dmitry-lyfar:


jonathan-conder (approved):
I feel like the description could just be blank so it doesn't look like a work in progress

PR #416, opened 2025-06-27
Support SDKs containing symlinks, compressed using Zstandard

jonathan-conder:
# Description

Prepare for upcoming SDK format changes in SDKcraft. The main difference is using `zstd` for compression. Store SDKs will now allow symlinks, so we should preserve them for local SDKs too.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/lxd/lxd_backend_volume.go
```diff
@@ -149,16 +150,30 @@ func (s *Backend) ImportVolume(ctx context.Context, name string, tarball string)
...
+	// just not setuid, setgid, etc. The parent directory should be empty
+	// with 700 permissions. For more details see
+	// https://www.gnu.org/software/tar/manual/html_section/Security.html
+	unpack := exec.CommandContext(ctx, "bash",
+		"-c",
+		`umask 0 && exec -- "$0" "$@"`,
+		"tar",
 		"--extract",
+		"--no-same-owner",
+		"--no-same-permissions",
```
dmitry-lyfar, line 163: How `--no-same-permissions` would affect SDKs like `flutter` that ship a executable binary? I suppose the permissions would be umasked?

jonathan-conder: It's a poorly named option. It means the permissions are affected by the umask, which we set to 0 so it does nothing. I tested with all permissions and the effect is to just clear the setuid, setgid and sticky bits, i.e. `**** -> 0***`.

dmitry-lyfar (approved):

PR #417, opened 2025-06-27
Improve LXD device error messages

jonathan-conder:
# Description

Some stuff I noticed while testing https://github.com/canonical/workshop/pull/394 but didn't make it in time.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/lxd_device/backend.go
+++ b/internal/interfaces/lxd_device/backend.go
@@ -53,7 +53,7 @@ func setupMounts(conn lxd.InstanceServer, fs workshop.WorkshopFs, user *user.Use
 		var err error
 		mounts, content, err = readMountProfile(fs)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("read filesystem table: %w", err)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/interfaces/lxd_device/backend.go
```diff
@@ -278,21 +278,32 @@ func writeMountProfile(fs workshop.WorkshopFs, mounts io.WriterTo) error {
...
+		case "Command not executable", "Command not found":
+			// Usually a nonzero exit status is not an error,
+			// but LXD translates 126 and 127 into the above messages.
+		default:
+			return fmt.Errorf("%s: %w", strings.Join(cmd, " "), err)
+		}
+	}
 
-	return op.Wait()
+	if status := int(op.Get().Metadata["return"].(float64)); status != 0 {
```
dmitry-lyfar, line 302: I assume we always rely on the fact that the key exists. There is another place like that where we don't check it.

dmitry-lyfar (approved):

internal/interfaces/lxd_device/backend.go
jonathan-conder: Yes I just copied this code from somewhere else

akcano (approved):

PR #432, opened 2025-07-10
Doc: Revamp tutorial

akcano:


docs/tutorial/get-started.rst
```diff
@@ -282,6 +282,23 @@ The workshop stays operational with no extra steps on your part
...
+
+
+   To the contrary, the definition and the :file:`.workshop/` directory
+   are *meant* to be stored in a repository;
+   if your :file:`.gitignore` file uses rules
+   such as "ignore everything except these files and directories,"
+   add them to the list of explicitly tracked items.
+
+
 To see how |ws_markup| keeps track of the directory,
```
dmitry-lyfar: I think this is not relevant anymore as we do not move project directory in the tutorial.

dmitry-lyfar: Also, if we we did, tasks do not demonstrate how the project directory is tracked, one can see this in the `list` output once they try to move their project around.

docs/tutorial/get-started.rst
```diff
@@ -7,7 +7,7 @@
 Get started with workshops
```
dmitry-lyfar, line 7: Nice restructure. I think this part of the tutorial should end up not only running a workshop but also building a trivial Go example since the workshop has go SDK. It will indicate why workshops at all for that matter and will rid you of the need to create `hello-world` example as part of the tutorial.

This is the official `go` samples repo which can be just `git clone` in the tutorial.

docs/tutorial/index.rst
```diff
@@ -8,51 +8,23 @@
...
    :maxdepth: 1
 
-   craft-sdks
+   Part 1: Get Started <get-started>
+   Part 2: Work with interfaces <work-with-interfaces>
+   Part 3: Sketch SDKs <sketch-sdks>
+   Part 4: Craft SDKs <craft-sdks>
 
 
 Next steps
```
dmitry-lyfar: Maybe drop the next steps from this page?

docs/tutorial/get-started.rst
dmitry-lyfar: Maybe we drop changes and tasks from here and move them to a separate pre-remove paragraph. Something like:

> Type `workshop tasks` and you'll see an output similar to the below. Every task expresses a simple token of logic, e.g. run a hook, connect an interface an so on. A set of tasks constitute a change. By default, every workshop change must have all its tasks succeeding or it will be undone. Etc, etc.

docs/tutorial/work-with-interfaces.rst
```diff
@@ -0,0 +1,231 @@
...
+
+This is the second section of the :ref:`four-part series <tut_index>`;
+it explains how to work with interfaces.
+It relies on the knowledge gained in the :ref:`tut_get_started` section,
+where you learned how to create and run workshops.
+
+.. @artefact interface
+.. @artefact system SDK
+
+SDKs use interfaces to interact in an organized manner,
```
dmitry-lyfar, line 22: We have a section in the first part of the tutorial that introduces SDKs implicitly, that is `Launch a workshop -> Define`. Maybe, we should be more explicit there about introducing an SDK and tell me if it's a valid concern. When I put myself in the first user's shoes that continues with this part, I might want to come back to the first part to remind myself of SDKs (as we mostly speak in workshop command terms there). And I'm not finding `SDK` in the navigation easily (although, it's part of the Define paragraph). I have no particular solution in mind, but this I think is a problem.

dmitry-lyfar (commented):
OK, separating interfaces is definitely a good idea. See my comments, we should continue on a call.

docs/tutorial/index.rst
```diff
@@ -8,51 +8,23 @@
...
-To create SDKs for |ws_markup|, SDK publishers use |sdk_markup|.
-This tool is installed separately and accepts an SDK definition
-to build and publish the SDK in the SDK Store (credentials provided on request):
+Follow the sections of the tutorial in order.
 
 .. toctree::
    :maxdepth: 1
 
-   craft-sdks
+   Part 1: Get Started <get-started>
```
dmitry-lyfar: It looks like the main page needs to change to have four sections too.

docs/tutorial/get-started.rst
```diff
@@ -500,195 +483,55 @@ This isn't the only way the host interacts with the workshop;
...
-|ws_markup| matches the plug to the slot using their names,
-then validates and enables the connection.
-Check the result using :command:`workshop info`:
+.. @artefact workshop tasks
 
 .. code-block:: console
 
-   $ workshop refresh
-   $ workshop info
+   $ workshop tasks
```
dmitry-lyfar, line 518: I kept not noticing it but actually our `workshop tasks` output has changed!

```
Status    Duration  Summary
Done          35ms  Retrieve "system" SDK
Done          35ms  Retrieve "go" SDK from channel "noble/stable"
...
Done          95ms  Install "go" SDK
Done          31ms  Register "go" SDK plugs and slots
Done     1m44.117s  Run hook "setup-base" for "go" SDK
...
......................................................................
Run hook "setup-base" for "go" SDK
```

docs/readme.rst
```diff
@@ -74,15 +74,17 @@ create a workshop definition file named ``workshop.yaml``
 to list your project's prerequisites,
 then run ``workshop launch``:
 
-.. code-block:: console
+.. code-block:: yaml
+   :caption: workshop.yaml
 
-   cat > workshop.yaml <<EOF -
    name: golang
```
dmitry-lyfar: A couple of small things are broken here, maybe we can add it to the PR: https://warthogs.atlassian.net/browse/WSP-747

dmitry-lyfar (approved):
Nice. Some updates left but I think it should be good to go after that.

PR #436, opened 2025-07-17
Move tools setup to setup-project hook

dmitry-lyfar:


jonathan-conder (approved):
This is fine given our current setup. But I'm leaning more towards running `setup-base` as the workshop user, and reserving `setup-project` for cases which need `/project`, interfaces or environment variables

PR #437, opened 2025-07-17
Doc: Update structure, add local release notes

akcano:


.workshop/starter-pack/hooks/setup-base
```diff
@@ -1,6 +1,10 @@
+set -eux
```
dmitry-lyfar: This is set by the caller, so no need.

.workshop/starter-pack/hooks/setup-project
```diff
@@ -1,8 +1,24 @@
-rm -rf /tmp/sp /project/docs/{Makefile,.sphinx}
+set -eux
```
dmitry-lyfar: This is set by the caller, so no need.

docs/how-to/use-workshops/connect-vscode-to-workshop.rst
```diff
@@ -1,22 +1,16 @@
...
+                 instance to a remote workshop.
 
-How to use workshops with Visual Studio Code
-============================================
+How to connect your local VS Code to a workshop
+===============================================
 
 .. @artefact workshop (container)
 
 One of the goals for |ws_markup| is robust integration with developer tooling,
```
dmitry-lyfar: Maybe we should drop the intro here and for the VS code in the web browser and have same short intro for both, e.g. `This article shows how to use vscode with Workshop`. Not that blunt but along the lines. I sense we should not oversell it here.

docs/tutorial/get-started.rst
```diff
@@ -14,7 +14,9 @@ of the essential |ws_markup| activities.
 
 .. @artefact workshop (container)
 
-A workshop is an environment
+A workshop
```
dmitry-lyfar: If you check the explanation section (Workshop concepts), it says:

```
A workshop (lowercase; not to be confused with Workshop itself) is a container that enables consistent environment builds.
```

OR (in the Explanation index):

```
Workshops are containers for consistent environment builds, and projects are the working directories where workshop definitions are placed.
```

We should choose one. I like this one ^, but I also think it can be simpler written for developers without too much generalisation, e.g. A workshop is a development environment running in a LXD container. No insisting on this one, but to give an idea.

.github/workflows/markdown-style-checks.yml
```diff
@@ -19,6 +19,9 @@ jobs:
...
         fetch-depth: 0
-    - uses: DavidAnson/markdownlint-cli2-action@v16
-      with:
-        config: "docs/.sphinx/.markdownlint.json"
+    - name: Create venv
+      working-directory: "docs"
+      run: make install
+    - name: Lint markdown
+      working-directory: "docs"
+      run: make lint-md
```
dmitry-lyfar: No new line.

.workshop/starter-pack/hooks/setup-base
jonathan-conder: I think we only `set -eo pipefail` now. But I think we were planning to add `-x` in verbose mode IIRC

docs/tutorial/get-started.rst
akcano: I've dropped this change for now.

dmitry-lyfar (approved):

PR #441, opened 2025-07-22
Doc: Add Jupyter how-to, rearrange how-tos

akcano:


docs/how-to/develop-with-workshops/run-jupyterlab-in-browser.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. code-block:: yaml
+   :caption: workshop.yaml
+   :emphasize-lines: 5-10
+
+    name: dev
+    base: ubuntu@24.04
+    sdks:
+    - name: system
+        plugs:
```
dmitry-lyfar: plugs must be at the same level with `name`

docs/how-to/develop-with-workshops/run-jupyterlab-in-browser.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. code-block:: yaml
+   :caption: workshop.yaml
+   :emphasize-lines: 11-12
+
+    name: dev
+    base: ubuntu@24.04
+    sdks:
+    - name: system
+        plugs:
```
dmitry-lyfar: plugs must be at the same level with `name`

docs/how-to/develop-with-workshops/run-jupyterlab-in-browser.rst
```diff
@@ -0,0 +1,91 @@
...
+        jupyter:
+            interface: tunnel
+            endpoint: 127.0.0.1:8989
+    - name: jupyter
+        channel: noble/edge
+    - name: openvino-toolkit
```
dmitry-lyfar: I would drop the `openvino-toolkit` example as it should be about the IDE only. We should probably specify that `jupyter` starts as a user service with `/project` as a default work directory to serve from.

It's fine if the `how-to` is short. Maybe we should join them all in on a single page?

docs/how-to/develop-with-workshops/run-jupyterlab-in-browser.rst
```diff
@@ -0,0 +1,91 @@
...
+    name: dev
+    base: ubuntu@24.04
+    sdks:
+    - name: system
+        plugs:
+        jupyter:
+            interface: tunnel
+            endpoint: 127.0.0.1:8989
+    - name: jupyter
+        channel: noble/edge
```
dmitry-lyfar: `channel` should be a the same level with `name`.

dmitry-lyfar (changes requested):

docs/how-to/develop-with-workshops/run-jupyterlab-in-browser.rst
akcano: I'd rather have them separate at this point; smaller is actually better.

dmitry-lyfar (approved):

PR #442, opened 2025-07-22
Doc: Improve SECURITY, CONTRIBUTING, add meta tags to release notes, fix CLI ref wording

akcano:


dmitry-lyfar (approved):
It looks OK to me, the policy requirement is to have the markdown security doc in the root of the repo. We can come back to it once GA.

PR #447, opened 2025-07-29
Doc: Extend workshop description, add review config

akcano:


docs/explanation/workshops/concepts.rst
```diff
@@ -15,8 +15,10 @@ Workshop concepts
 A *workshop*
 (lowercase; not to be confused with |ws_markup| itself)
 is a container that enables consistent environment builds.
-It is tied to a definition that lists SDKs
-and is stored as a :file:`.yaml` file under a project directory.
+A workshop is defined by a single YAML file
+that acts as the blueprint for |ws_markup| to implement at launch time.
+It describes how individual components fit together
+to create a cohesive development environment.
```
dmitry-lyfar, line 21: We need an update to the workshop statuses table below which hasn't changed in this PR so am commenting here. 

1. The definition column is not relevant anymore, as a workshop container can exist without having it's definition and it won't be an error. You can remove it but cannot refresh for obvious reasons. 

2. It also looks like that Description and Container columns can be merged together if p1 is applied (e.g. no need to have Container it seems). Lmk what you think.

akcano:
@dmitry-lyfar can you please review and approve?

docs/explanation/workshops/concepts.rst
```diff
@@ -114,33 +116,100 @@ A more complete definition would usually list several SDKs
...
+any plugs and slots available within the workshop,
+on top of what the :ref:`auto-connection mechanism <exp_interface_connections>`
+in |ws_markup| provides:
+eventually, all interface connections are
+:ref:`resolved, validated, and established <exp_interfaces_validation>`
+in a single task *after* all the SDK layers have been created,
+because all components must be in place before the wiring can be done.
 
-This example adds a slot, a plug and two connections to its SDKs:
+This example adds a slot, a plug, and two connections to its SDKs:
```
dmitry-lyfar, line 207: I reviewed this example and it feels a bit too complex (I know it's not a part of the PR, but it's hard to follow). Let me suggest a refined version:

```
base: ubuntu@22.04
name: dev
sdks:
  - name: tensorflow
    channel: latest/stable
    plugs:
      cuda:
        interface: mount
        workshop-target: /usr/local/cuda/lib64 
  - name: cuda
    channel: latest/stable
connections:
  - plug: tensorflow:cuda
    slot: cuda:libs
```

This example:
1. Extends `tensorflow SDK` with a standard path for CUDA runtime libraries.
2. We explicitly connect the newly defined plug to the slot from `cuda SDK`. Thus, upon the workshop creation, the plug will not be automatically connected to the default location on the host coming from the `system SDK` (`.../<id>/<workshop>/...`) but to the `libs` at a path in the workshop.

We actually use this mechanism for RTOS environments, but the actual example would more complex than the above.

docs/reference/definition-files/sdk-definition.rst
```diff
@@ -16,10 +16,10 @@ Filename convention
 .. @artefact sdkcraft (CLI)
 
 The name of the SDK definition file must be :file:`sdk.yaml`;
-the file is usually created using the :command:`sdkcraft init` command
-in the source directory when :ref:`sketching a local SDK <tut_sketch_sdks>`
-or
-:ref:`crafting and publishing an SDK <tut_craft_sdks>`.
+the file is usually stored under :file:`$XDG_DATA_HOME/workshop/`
```
dmitry-lyfar, line 19: Please use an updated example for the `go` SDK which also includes a proper part instead of `nil`: https://github.com/canonical/sdks/blob/main/go/sdk.yaml

dmitry-lyfar (commented):
Nice. I'd like to suggest some changes (see the comments).

docs/explanation/workshops/concepts.rst
```diff
@@ -42,38 +42,32 @@ A workshop's life-cycle can see it switch between several statuses:
...
+     - Not operational;
+       the workshop container is running
+       but is being updated and is not ready for use.
 
    * - *Waiting*
-     - Paused for debugging a launch or refresh error
-     - In the project directory
-     - Running
+     - Operational;
+       the workshop container is running but paused,
```
dmitry-lyfar: I'd say it is running and available for command execution, however its current change is in progress. The container itself is not paused.

dmitry-lyfar (approved):

PR #448, opened 2025-07-31
Support trying out packed SDKs

jonathan-conder:
# Description

Support `sdkcraft try`. After running `sdkcraft try` in an SDKcraft project, the local user can use the SDK(s) it built by adding `- name: try-<SDK>` to a workshop definition.

Currently the SDK volume is imported synchronously (i.e. not part of the `launch` or `refresh` change). This is something we'll look to rework once we know more about the Store API (and maybe have nicer output for dynamic `workshop tasks`).

Another outstanding issue is cleaning up SDK volumes (and SDKs copied to the try area) when no longer in use.

Breaking changes:
- The list of installed SDKs in the workshop's metadata now tracks where the SDK came from. Existing workshops may need to be removed before switching to this branch.
- SDK volumes now include additional metadata. I don't think any of this is currently used outside of `try`, but it might be worth removing old volumes.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/explanation/sdks/concepts.rst
+++ b/docs/explanation/sdks/concepts.rst
@@ -192,6 +192,27 @@ Note that :samp:`sketch` is a reserved name,
 and the sketch SDK is always installed last.
 
 
+.. _exp_try_sdk:
+
+Trying out SDKs
+---------------
+
+.. @artefact sdkcraft (CLI)
+.. @artefact try SDK
+
+The :command:`sdkcraft try` command allows publishers to test SDKs
+before uploading them to the Store.
+Once installed in a workshop, these SDKs behave identically to SDKs from the Store.
+
+|sdk_markup| does not install SDKs in a workshop by itself;
+it simply copies packed SDKs to a directory called the *try area*.
+|ws_markup| looks in this directory when installing an SDK with the :samp:`try-` prefix.
+
+The try area has no channels;
+only one version of an SDK can be tested at a time.
+However, this version can be tested in multiple workshops with different :ref:`bases <exp_base>`.
+
+
 .. _exp_in_project_sdk:
 
 In-project SDKs
--- a/internal/overlord/sdkstate/handlers.go
+++ b/internal/overlord/sdkstate/handlers.go
@@ -90,6 +90,7 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
 	volume := workshop.VolumeInfo{
 		Name: sdk.VolumeName(rec.Name, rec.Revision),
 		Kind: "sdk",
+		Sdk:  rec.Name,
 	}
 	err = m.backend.ImportVolume(ctx, volume, rec.Filepath())
 	if errors.Is(err, workshop.ErrVolumeAlreadyExists) {
--- a/internal/overlord/sdkstate/handlers.go
+++ b/internal/overlord/sdkstate/handlers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"gopkg.in/tomb.v2"
 
@@ -93,7 +94,12 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
 		Sdk:      rec.Name,
 		Revision: rec.Revision,
 	}
-	err = m.backend.ImportVolume(ctx, volume, rec.Filepath())
+	file, err := os.Open(rec.Filepath())
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	err = m.backend.ImportVolume(ctx, volume, file)
 	if errors.Is(err, workshop.ErrVolumeAlreadyExists) {
 		logger.Debugf("SDK Manager on maybeCreateVolume: reuse existing SDK volume %q", sdk.VolumeName(rec.Name, rec.Revision))
 		return nil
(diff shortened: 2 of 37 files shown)


internal/overlord/workshopstate/request.go
```diff
@@ -93,7 +101,7 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
 			return nil, fmt.Errorf("cannot launch %q, other changes in progress: %w", name, err)
 		}
 
-		localSdks, err := local.resolveSdks(name, req.file, nil)
+		localSdks, err := local.resolveSdks(ctx, nil, req.file)
```
dmitry-lyfar: Currently:

```
error: cannot refresh "dev": open /home/dmitry/.local/share/workshop/sdk/zephyr: no such file or directory
```

is given for a `try-` SDK. Can we give more context in the error message, so that a user understands what this path is about? I don't think `sdkcraft try` exposes it either. So there is some time when this "try area" has to become explicit. We are aiming `sdk list` command, so, if not here, we can try to solve it there.

docs/explanation/sdks/concepts.rst
```diff
@@ -192,6 +192,29 @@ Note that :samp:`sketch` is a reserved name,
...
+
+There are no channels;
+only one version of an SDK can be tested at a time.
+However, this version can be tested in multiple workshops with different :ref:`bases <exp_base>`.
+
+SDK publishers can define testing workshops
+in their |sdk_markup| project.
+A typical workflow is to update the SDK definition
+and run :command:`sdkcraft try && workshop refresh <WORKSHOP>`.
+The :samp:`check-health` :ref:`hook <exp_hooks>` can be used to ensure the SDK is functioning correctly.
```
dmitry-lyfar:
```suggestion
```

I think it should go to a separate section about building & maintaining an SDK, this one should only talk about the try area. Perhaps, we should rework [this](https://canonical-workshop.readthedocs-hosted.com/latest/explanation/sdks/workshopctl/) with more examples and explaining the intent clearer. I added an item to discuss it with @akcano when he is back.

internal/overlord/sdkstate/handlers.go
```diff
@@ -153,7 +165,7 @@ func (m *SdkManager) mountSdk(ctx context.Context, user string, project *worksho
 	name := sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision)
 	sdkPath := sdk.SdkDir(sdkSetup.Name)
 
-	if sdkSetup.Revision.Store() {
+	if sdkSetup.Source.NeedsVolume() {
```
dmitry-lyfar: I would decouple the knowledge of `NeedsVolume` from the source itself. The source should not know whether it needs a volume. Perhaps, `sdkSetup.IsVolume`?

internal/overlord/workshopstate/request.go
```diff
@@ -71,7 +74,12 @@ func (w *WorkshopManager) LaunchMany(ctx context.Context, names []string, projec
 	if err != nil {
 		return nil, err
 	}
-	local := localStore{usr, userDataDir, project}
+	local := localResolver{
```
dmitry-lyfar: It feels like we add new meaning to the `resolve` verb. Refresh has a resolver that produces a plan for the refresh change. This resolver does something else, i.e. its result is not a blueprint for a change but a list of SDKs from local sources. Maybe `localSdkRetriever`?

dmitry-lyfar: Same logic applies to `resolveWorkshops` and `resolveStoreSdks`.

internal/workshop/workshop_dirs.go
```diff
@@ -59,6 +59,10 @@ func SketchSdkStash(userDataDir, pid, w string) string {
 	return filepath.Join(SketchSdkDir(userDataDir, pid, w), "stash")
 }
 
+func TrySdkDir(userDataDir, sdk string) string {
```
dmitry-lyfar, line 62: `.../workshop/try` ? When / if we add a local SDK, where does it go?

tests/main/try-sdks/task.yaml
```diff
@@ -0,0 +1,44 @@
...
+  echo fakehash | sudo -u ubuntu tee "$try_area/test-sdk-basic_all.sdk.sha3-384"
+  sudo -u ubuntu cp "$store_area/latest/stable/sdk.yaml" "$try_area/test-sdk-basic_all.sdk.yaml"
+
+  workshop_exec launch
+  workshop_exec exec -- findmnt /var/lib/workshop/sdk/test-sdk-basic | MATCH 'workshop/custom/default_test-sdk-basic-x1'
+  try_area_contracted="~${try_area#*ubuntu}"
+  workshop_exec info | grep -v notes | yq '.sdks["test-sdk-basic"].tracking' | MATCH "^${try_area_contracted}\$"
+  workshop_exec info | grep -v notes | yq '.sdks["test-sdk-basic"].installed' | MATCH '\(x1\)$'
+
+  echo "Check try and in-project SDKs can share revision numbers"
```
dmitry-lyfar, line 24: Why can they not in theory?

jonathan-conder: It relates to this change I made in `maybeRefresh`:
https://github.com/canonical/workshop/blob/413e877857b4ac1a7e9885ee18fc630809ba0a66/internal/overlord/workshopstate/request.go#L658-L660
We used to be able to assume that different SDKs have different revision numbers, but not so anymore. I wanted to catch other places which made this assumption (as far as I can tell there aren't any now).

docs/explanation/sdks/concepts.rst
jonathan-conder: I think it's a good idea to mention the try area, and yeah the last part bleeds a bit much into how-to.

dmitry-lyfar (approved):

PR #457, opened 2025-08-12
Remove unused SDK volumes

dmitry-lyfar:
# Description

Implement tracking and removing unused SDK volumes. The implementation uses a clean up handler for the tasks of `install-sdk` or `unregister-sdk` type. Once a change with such a task is complete, it may indicate that an SDK volume became unused due to:
- workshop was removed
- SDK volume was detached from a workshop as a result of a refresh change and is no longer used by any other workshop
- workshop launch failed

If an SDK volume is found to be unused for longer than the `cool down time` (1 hr), it will be removed from the system. This is done to prevent `workshopd` to remove volumes that have a high chance to be immediately reused on a following refresh or launch.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/sdkstate/handlers.go
+++ b/internal/overlord/sdkstate/handlers.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
 	"time"
 
 	"gopkg.in/tomb.v2"
@@ -89,7 +90,7 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
 		}
 	}
 
-	volume := workshop.VolumeInfo{
+	volume := workshop.VolumeSetup{
 		Name:     sdk.VolumeName(rec.Name, rec.Revision),
 		Kind:     "sdk",
 		Sdk:      rec.Name,
@@ -267,26 +268,30 @@ func (m *SdkManager) doUnregisterSdk(task *state.Task, tomb *tomb.Tomb) error {
 
 type SdkVolumeCooldownTimeKey string
 
-func sameVolume(task *state.Task, other *state.Task) bool {
-	var sdkSetup sdk.Setup
-	if err := task.Get("sdk-setup", &sdkSetup); err != nil {
-		return false
-	}
-
-	var otherSdkSetup sdk.Setup
-	if err := other.Get("sdk-setup", &otherSdkSetup); err != nil {
-		return false
-	}
-
-	if !sdkSetup.IsVolume() || !otherSdkSetup.IsVolume() {
-		return false
-	}
-
-	return sdkSetup.Name == otherSdkSetup.Name && sdkSetup.Revision == otherSdkSetup.Revision
-}
-
+// doDeleteUnusedSdkVolumes is a cleanup handler that deletes unused SDK volumes.
+// It is called periodically to ensure that SDK volumes that are no longer
+// needed are removed from the system. The cleanup is done only if there are no
+// tasks in progress that use the same SDK volume. The cleanup is also
+// throttled to avoid deleting volumes too frequently. The cooldown time is
+// defined by the sdkVolumeCooldownTime constant.
+//
+// There are multiple scenarios when this cleanup is triggered:
+//   - A workshop was removed and an SDK volume was detached from it (unregister-sdk task)
+//   - An SDK volume was detached during a refresh change,
+//     and is no longer used by any other workshop (unregister-sdk or, if refresh failed, install-sdk task)
+//   - A workshop launch failed (install-sdk task)
 func (m *SdkManager) doDeleteUnusedSdkVolumes(task *state.Task, tomb *tomb.Tomb) error {
 	st := task.State()
+
+	sdkSetup, err := SdkSetup(task)
+	if err != nil {
+		logger.Debugf("On SdkManager.Cleanup: the %q task is not associated with a SDK setup", task.ID())
+		return nil
+	}
+
+	if !sdkSetup.IsVolume() || sdkSetup.Source == sdk.SystemSource {
+		return nil
+	}
 
 	user, prj, _, err := UserProjectWorkshop(task)
 	if err != nil {
@@ -306,22 +311,14 @@ func (m *SdkManager) doDeleteUnusedSdkVolumes(task *state.Task, tomb *tomb.Tomb)
 		return nil
 	}
 
-	var sdkSetup sdk.Setup
-	if err := task.Get("sdk-setup", &sdkSetup); err != nil {
-		logger.Debugf("On SdkManager.Cleanup: the %q task is not associated with a SDK setup", task.ID())
-		return nil
-	}
-
-	if !sdkSetup.IsVolume() || sdk.IsSystem(sdkSetup.Name) {
-		return nil
-	}
-
 	vk := SdkVolumeCooldownTimeKey(sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision))
 	cooldownStart, ok := st.Cached(vk).(time.Time)
 	if !ok {
 		// No cooldown start time cached means that clean up has just started or
 		// workshopd was restarted.
 		st.Cache(vk, task.ReadyTime())
+		// We need to return here to give other competing cleanup tasks a chance
+		// to find out who was the last one to initiate the cleanup.
 		return &state.Retry{}
 	} else {
 		// Imagine the situation when a change with multiple tasks that use this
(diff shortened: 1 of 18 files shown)


internal/workshop/fakebackend/backend.go
```diff
@@ -211,13 +217,23 @@ func (f *FakeWorkshopBackend) RemoveWorkshop(ctx context.Context, name string, f
...
-	defer f.workshopLock.Unlock()
-
-	if _, ok := f.Workshops[prj.ProjectId][name]; !ok {
+	_, ok := f.Workshops[prj.ProjectId][name]
+	f.workshopLock.Unlock()
+	if !ok {
 		return workshop.ErrWorkshopNotLaunched
 	}
 
+	for _, sk := range f.Workshops[prj.ProjectId][name].Sdks {
```
```suggestion
	wp, ok := f.Workshops[prj.ProjectId][name]
	f.workshopLock.Unlock()
	if !ok {
		return workshop.ErrWorkshopNotLaunched
	}

	for _, sk := range wp.Sdks {
```
I don't think it's safe to access `f.Workshops` without holding the lock (or at the top level of a test). But an individual workshop should be OK.

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -473,39 +473,45 @@ func (f *wsOps) TestLxdBackendWorkshopRebuild(c *check.C) {
...
-	err = system.RetrieveSystemSdk(setup, nil)
-	c.Assert(err, check.IsNil)
-	meta, err := system.SystemSdkFs.ReadFile(filepath.Join("meta", "sdk.yaml"))
-	c.Assert(err, check.IsNil)
-
-	volume := workshop.VolumeInfo{
-		Name:     sdk.VolumeName(setup.Name, setup.Revision),
+	sdkfs := c.MkDir()
+	volume := workshop.VolumeSetup{
+		Name:     "test-1",
```
jonathan-conder:
```suggestion
		Name:     "test-sdk-1",
```

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -473,39 +473,45 @@ func (f *wsOps) TestLxdBackendWorkshopRebuild(c *check.C) {
...
+		Revision: sdk.R(1),
+		Metadata: testsdk,
 	}
-	file, err := os.Open(setup.Filepath())
+	setup := sdk.Setup{
+		Name:     "test-sdk",
+		Revision: sdk.R(5),
+		Source:   sdk.StoreSource,
+	}
+	tarball := helper.MockSdkTarball(c, "test", sdkfs, testsdk)
```
jonathan-conder:
```suggestion
	tarball := helper.MockSdkTarball(c, "test-sdk", sdkfs, testsdk)
```

internal/workshop/lxd/tests/integration/workshop_test.go
```diff
@@ -517,31 +523,88 @@ func (f *wsOps) TestLxdBackendWorkshopRestore(c *check.C) {
...
+
+	// Second workshop
+	other := workshop.Project{ProjectId: "24242424", Path: c.MkDir()}
+	otherCtx := helper.CreateTestContext(f.usr.Username, other.ProjectId)
+	helper.LaunchTestWorkshop(c, otherCtx, f.bd, other.Path)
+	defer helper.RemoveTestWorkshop(c, otherCtx, f.bd)
+
+	sdkfs := c.MkDir()
+	volume := workshop.VolumeSetup{
+		Name:     "test-1",
```
jonathan-conder:
```suggestion
		Name:     "test-sdk-1",
```

internal/workshop/fakebackend/backend.go
```diff
@@ -618,23 +658,30 @@ func (s *FakeWorkshopBackend) Volumes(ctx context.Context, kind string) ([]works
...
 	if !ok {
 		return workshop.VolumeInfo{}, workshop.ErrVolumeNotFound
 	}
 
+	info.Workshops = make(map[string][]string)
+	for mntp := range s.SdkVolumeMountPoints {
+		if mntp.VolumeName == name {
+			info.Workshops[mntp.ProjectId] = append(info.Workshops[mntp.ProjectId], mntp.Workshop)
+		}
+	}
```
jonathan-conder: I think `s.SdkVolumes[name].Workshops` is updated in sync with `s.SdkVolumeMountPoints` so this seems unnecessary

internal/overlord/sdkstate/handlers_test.go
```diff
@@ -484,3 +489,321 @@ func (s *sdkStateSuite) TestDoRegisterSdkBadInterfacesFound(c *check.C) {
...
+	t := s.state.NewTask("install-sdk", "...")
+	t.Set("sdk-retrieve-task", t.ID())
+	t.Set("sdk-setup", newSdk)
+
+	chg := s.state.NewChange("launch", "...")
+	chg.Set("user", "testuser")
+	setWorkshopProject("ws", s.project, t)
+	chg.AddTask(t)
+	s.state.Unlock()
+
```
jonathan-conder, line 567: I think the cooldown time should be 0 here

internal/overlord/sdkstate/handlers_test.go
```diff
@@ -484,3 +489,321 @@ func (s *sdkStateSuite) TestDoRegisterSdkBadInterfacesFound(c *check.C) {
...
+		s.se.Wait()
+	}
+	s.state.Lock()
+	defer s.state.Unlock()
+	c.Check(t.IsClean(), check.Equals, true)
+	_, err = s.backend.Volume(s.ctx, sdk.VolumeName("test", sdk.R(1)))
+	c.Assert(err, check.Equals, workshop.ErrVolumeNotFound)
+}
+
+func (s *sdkStateSuite) TestTaskExitCleanupOnSDKUsedAgain(c *check.C) {
```
jonathan-conder: this one seems very similar to `TestTaskQuitsCleanupIfOtherVolumeUsersExist` (but the state isn't unlocked at the end). I don't think it does what the name describes

internal/overlord/sdkstate/handlers.go
```diff
@@ -263,3 +265,122 @@ func (m *SdkManager) doUnregisterSdk(task *state.Task, tomb *tomb.Tomb) error {
...
+func (m *SdkManager) doDeleteUnusedSdkVolumes(task *state.Task, tomb *tomb.Tomb) error {
+	st := task.State()
+
+	sdkSetup, err := SdkSetup(task)
+	if err != nil {
+		logger.Debugf("On SdkManager.Cleanup: the %q task is not associated with a SDK setup", task.ID())
+		return nil
+	}
+
+	if !sdkSetup.IsVolume() || sdk.IsSystem(sdkSetup.Name) {
```
```suggestion
	if !sdkSetup.IsVolume() || sdkSetup.Source == sdk.SystemSource {
```

internal/overlord/sdkstate/handlers.go
```diff
@@ -263,3 +265,122 @@ func (m *SdkManager) doUnregisterSdk(task *state.Task, tomb *tomb.Tomb) error {
...
+
+	if len(vinfo.Workshops) > 0 {
+		logger.Debugf("On SdkManager.Cleanup: the %q volume is still in use, skip clean up", sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision))
+		st.Cache(vk, nil)
+		return nil
+	}
+
+	if time.Since(cooldownStart) < sdkVolumeCooldownTime {
+		return &state.Retry{}
+	}
```
jonathan-conder: I would move this up a bit as an optimisation

internal/overlord/sdkstate/handlers.go
```diff
@@ -263,3 +265,122 @@ func (m *SdkManager) doUnregisterSdk(task *state.Task, tomb *tomb.Tomb) error {
...
+
+	// Delete volume ignores ErrVolumeNotFound.
+	err = m.backend.DeleteVolume(ctx, sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision))
+	if err == nil || errors.Is(err, workshop.ErrVolumeInUse) {
+		if errors.Is(err, workshop.ErrVolumeInUse) {
+			logger.Debugf("On SdkManager.Cleanup: the %q volume is still in use, skip clean up", sdk.VolumeName(sdkSetup.Name, sdkSetup.Revision))
+		}
```
jonathan-conder: Since we can detect this error, is there a need to check `len(vinfo.Workshops) > 0`?

internal/overlord/sdkstate/handlers_test.go
dmitry-lyfar: Not sure, this is the test to validate the clean up will not continue attempts to clean up on successful "install-sdk" tasks after a launch as volumes are used.

internal/overlord/sdkstate/handlers.go
dmitry-lyfar: `IsVolume` already has this:

```
	return s.Source == StoreSource || s.Source == SystemSource || s.Source == TrySource

```

The choice to make the handler to return on system SDK was that it is tiny and an often reused core component that we don't need to recreate every time. So, to not keep the task unclean for the cool down time, it `return nil` here. Do you think we should not make exceptions for the system SDK?

internal/overlord/sdkstate/handlers_test.go
jonathan-conder: As discussed, I think my reasoning for this was to emphasize the main difference with the previous test (where install fails). But thinking about this test independently, keeping the default cooldown ensures the cleanup doesn't finish accidentally.

internal/overlord/sdkstate/handlers.go
jonathan-conder: Making exceptions makes sense, I just prefer not to match on the name if possible

internal/overlord/sdkstate/handlers.go
jonathan-conder: My main concern was that checking `UsedBy` might not be safe if volumes are attached concurrently. But I didn't realise the state is locked here. Both approaches have drawbacks: in future `UsedBy` might be populated with something other than a workshop, and the "in use" error message could change. To be safe we should probably start testing LXD release candidates periodically.

dmitry-lyfar: Modified `TestLxdBackendWorkshopUsedByInVolumeInfoOK` test to ensure that "volume is in use" error will be monitored in the integration runs, so that if the error message changes in the API we can detect it in periodic runs. Thus, it should be enough to rely on the `DeleteVolume` only and exit if a volume is in use.

jonathan-conder (approved):

PR #461, opened 2025-08-20
Add mode, uid, and gid options for mount plugs

jonathan-conder:
# Description

Ties in with https://github.com/canonical/sdkcraft/pull/170. The main issue addressed is for mount plugs like `~/go/pkg/mod` - previously the intermediate directories (`go` and `pkg`) were owned by `root:root`. Now they will be owned by `workshop:workshop`. This is configurable using the interface attributes `uid`, `gid` and `mode`.

The mount point itself should also be affected, but LXD has some issues (https://github.com/canonical/lxd/issues/12648 and https://github.com/canonical/lxd/issues/16264) which prevent this for now. Not a big deal since the mount point is only visible when the interface is disconnected.

Also fixes a bug where `workshop-source: $SDK` requires a trailing slash (which I thought I'd fixed already but I must have overlooked something).

Fixes a recurring issue with interface attributes being different types in workshop and SDK definitions. This is potentially a breaking change (if a launch or refresh is in progress). Removes the need to work around this in the tests.

Makes `.Xauthority` and `workshopctl` mounts read-only, because why not?

Adds some common directories to the cloud-config:
- `~/.cache`, `~/.config` and `~/.local` so SDKs don't need to worry about 0700 permissions
- `~/bin` and `~/.local/bin` so SDKs don't need to source `.profile` after installing something there

Fixes several correctness issues with `sftp`. I tried to reuse as much functionality as possible, but after fixing a few things the remaining `sftpfs` functions were just directly calling `sftp`, so I dropped `sftpfs`. Summary of issues:
- `Mkdir` and `OpenFile` aren't transactional, they ignore the umask,
  and don't return `ErrExist` (required for `afero.TempDir` and
  `afero.TempFile`). The rewrites are transactional, but not atomic.
  They respect a hardcoded umask of 0022, which is the Ubuntu default
  for `root`.
- Paths are stripped from certain error types. The wrappers add these back in.
- On error, functions can return non-nil file handles
- `Sync` and `WriteAt` aren't implemented at all. Fixed by removing from the
  interface.
- `Readdir` and `Readdirnames` don't use the open file handle. Fixed by adding
  `ReadDir` to the `Fs` interface.

After looking at the `sftp` package, I think SFTP is ultimately the right protocol
for us, but both the client and server side need some work to implement the `os`
filesystem functions correctly. Some custom extensions might be needed for
something like `os.Root`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/lxd/lxd_backend.go
```diff
@@ -1111,6 +1112,10 @@ write_files:
 runcmd:
   # Project directory is required for 'workshop exec'.
   - install --directory --mode=755 /project
+  # Create XDG base directories so SDKs don't need an extra mode=700 step.
+  - install --directory --mode=700 --owner=workshop --group=workshop /home/workshop/.cache /home/workshop/.config /home/workshop/.local
```
dmitry-lyfar, line 1116: I think it's OK, however, have you considered running `xdg-user-dirs-update`?

dmitry-lyfar (commented):
I reviewed everything except `fsutil` which I'm going to review separately now. There was just one comment so far.

Also, smoke-tested it on both types of mounts, works as expected.

internal/workshop/lxd/lxd_backend.go
jonathan-conder: It didn't look like it created the hidden dirs, from the docs. I just tried it now and it does create `.config` (for the purpose of `.config/user-dirs.{dirs,locale}`) but not `.local` or `.cache`.

I don't think workshops will have much use for some of the other dirs: `Desktop  Documents  Downloads  Music  Pictures  Public  Templates  Videos`

internal/fsutil/sftp.go
```diff
@@ -0,0 +1,226 @@
...
+
+func NewSftpFs(client *sftp.Client, umask os.FileMode) Fs {
+	return Fs{&sftpFs{client, umask}}
+}
+
+type sftpFs struct {
+	client *sftp.Client
+	umask  os.FileMode
+}
+
```
dmitry-lyfar, line 32:
```suggestion
var _ FsBackend = (*sftpFs)(nil)

```

internal/fsutil/basepath.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/spf13/afero"
+
+	"github.com/canonical/workshop/internal/revert"
+)
+
+type BasePathFile struct {
+	afero.BasePathFile
+}
+
```
dmitry-lyfar, line 16:
```suggestion
var _ FsBackend = (*BasePathFs)(nil)

```

internal/interfaces/builtin/mount.go
```diff
@@ -107,28 +97,99 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+	}
+
+	if _, err := parseBool(plug.Attrs, "read-only", false); err != nil {
+		return err
+	}
+
+	var fallbackUid, fallbackGid int64
+	for _, prefix := range []string{"/home/workshop", "/project", "/run/user/1000"} {
+		if target == prefix || strings.HasPrefix(target, prefix+string(filepath.Separator)) {
+			fallbackUid = 1000
```
dmitry-lyfar:
```suggestion
			fallbackUid = workshop.Uid
```

internal/interfaces/builtin/mount.go
```diff
@@ -107,28 +97,99 @@ func (iface *mountInterface) BeforePreparePlug(plug *sdk.PlugInfo) error {
...
+			fallbackGid = 1000
```
dmitry-lyfar:
```suggestion
			fallbackGid = workshop.Gid
```

docs/explanation/interfaces/mount-interface.rst
```diff
@@ -45,6 +45,8 @@ A basic structure would include the name of the plug itself,
 the interface (:samp:`mount`)
 the intended target path inside the workshop (:samp:`workshop-target`)
 and, optionally, whether the mount should be read-only (:samp:`read-only`).
+Some plugs customize the permissions (:samp:`mode`) and ownership (:samp:`uid`, :samp:`gid`) of the target path,
+and any parent directories that didn't already exist.
```
akcano: This tries to be many things at once.

```suggestion
Plugs may customize the permissions (:samp:`mode`) and ownership (:samp:`uid`, :samp:`gid`) of the target path.
Also, any directories in the target path that didn't already exist
will be created automatically.
```

Or something similar.

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -282,6 +282,28 @@ They are described by the following attributes:
...
+   * - :samp:`mode`
+     - integer
+     - File permissions to use when creating :samp:`workshop-target`
+       and any parent directories.
+       Default is 0o775 for normal users,
+       but changes to 0o755 if :samp:`uid` is zero.
+
+   * - :samp:`uid`
+     - integer
+     - User ID to use when creating :samp:`workshop-target`
```
```suggestion
User ID to apply when creating :samp:`workshop-target`
```

docs/reference/definition-files/workshop-definition.rst
```diff
@@ -282,6 +282,28 @@ They are described by the following attributes:
...
+       and any parent directories.
+       If :samp:`workshop-target` is in :file:`/project/`
+       or the :samp:`workshop` user's home or runtime directory,
```
akcano: What's runtime directory here? Doubt it's CWD.

akcano (changes requested):

docs/reference/definition-files/workshop-definition.rst
jonathan-conder: Looks like the indentation was changed unintentionally? But I will do `s/use/apply/`.

docs/reference/definition-files/workshop-definition.rst
jonathan-conder: It's `$XDG_RUNTIME_DIR`. I might as well just list the paths

docs/explanation/interfaces/mount-interface.rst
jonathan-conder: I went with
```
|ws_markup| will create the target path if it doesn't exist.
Plugs may customize the permissions (:samp:`mode`) and ownership (:samp:`uid`, :samp:`gid`)
of any directories created.
```

dmitry-lyfar (approved):

akcano (approved):

PR #462, opened 2025-08-20
Make verbose mode off for sketch-sdk by default

dmitry-lyfar:
# Description

1. Verbose mode does not seem to be a good choice to have on by default for `sketch-sdk`, especially if it's used for many consequential iterations. Add `--verbose` flag to control this behaviour similarly to `refresh`.
2. @akcano I turned the `project-starter-pack` temporarily from the project as it prevents the workshop from launching. It should come back with your fix. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -27,6 +27,7 @@ type CmdSketch struct {
 	eject   bool
 	name    string
 	remove  bool
+	verbose bool
 }
 
 func (c *CmdSketch) Command() *cobra.Command {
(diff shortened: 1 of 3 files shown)


akcano (approved):

cmd/workshop/sketch.go
```diff
@@ -491,7 +493,7 @@ func (c *CmdSketch) Run(cmd *cobra.Command, av []string) error {
 
 	cmdrefresh := &CmdRefresh{root: c.root}
 	cmdrefresh.WaitOnError = true
-	cmdrefresh.verbose = true
+	cmdrefresh.verbose = c.verbose
```
jonathan-conder, line 499: I don't think this is the only CmdRefresh that we could apply the flag to

dmitry-lyfar: You are right. For some reason, I considered `sketch-sdk` options as "maintenance commands"  that would not need an extended output, like `--stash`, but all of them would as other SDKs may get updates and can be refreshed too.

jonathan-conder (approved):

PR #465, opened 2025-08-29
Fix parsing mount plug attributes from the state

jonathan-conder:
# Description

Some tasks use plug and slot attributes from the state rather than the interface repository. These attributes can be mangled because JSON doesn't have a notion of integers (and they are deserialised into a `map[string]any`).

When undoing a failed refresh for a workshop with an auto-connected mount plug, the original connection is restored incorrectly. The mount interface ignored errors relating to float to int conversions, resulting in the `mode`, `uid` and `gid` attributes being zero, when they should have been more reasonable defaults. Ultimately this can cause permission issues when trying to read or write files in the mounted directory.

For now, this fix allows `float64 -> int64` conversions if they are safe. The mount interface will also report these errors, since we clearly can't rely on them being impossible (but I still think they are, currently).

Also, I don't think the mount interface should have been modifying the directory anyway, since it already existed. Although maybe you could argue that an updated mode should be reflected on refresh (this is automatic for the actual target because it goes away when restoring the snapshot).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/interfaces/lxd_device/backend.go
+++ b/internal/interfaces/lxd_device/backend.go
@@ -182,8 +182,10 @@ func prepareMount(fs fsutil.Fs, user *user.User, mnt workshop.Mount) error {
 }
 
 func prepareHostWorkshopMount(fs fsutil.Fs, user *user.User, mnt workshop.Mount) error {
-	whatIsDir := true
-	if mnt.MakeWhat {
+	info, err := os.Stat(mnt.What)
+	isDir := err == nil && info.IsDir()
+
+	if mnt.MakeWhat && !isDir {
 		uid, gid, err := osutil.UidGid(user)
 		if err != nil {
 			return err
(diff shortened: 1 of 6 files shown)


internal/interfaces/lxd_device/backend.go
```diff
@@ -182,8 +182,10 @@ func prepareMount(fs fsutil.Fs, user *user.User, mnt workshop.Mount) error {
 }
 
 func prepareHostWorkshopMount(fs fsutil.Fs, user *user.User, mnt workshop.Mount) error {
-	whatIsDir := true
-	if mnt.MakeWhat {
+	info, err := os.Stat(mnt.What)
+	whatIsDir := err == nil && info.IsDir()
```
dmitry-lyfar:
```suggestion
	whatDirExists := err == nil && info.IsDir()
```
?

or 

```
osutil.ExistsIsDir
```

jonathan-conder: I want to preserve the `os.Stat` error message for the case where `MakeWhat` is false. Maybe we call the variable `existsIsDir` though

dmitry-lyfar: Yes, the name was a bit confusing when read for the first time. The logic I realised: if what is not a dir, it means it is a file and we should make it if makeWhat is true.

jonathan-conder: actually we can just call it `isDir`, since it applies to both `what` and `where`

jonathan-conder: yeah in the case where it's a file, it should error out when calling mkdir

dmitry-lyfar (approved):

jonathan-conder:
Now I think not `chmod`ing the source if it exists is the right call. We preserve the contents of `host-source` between refreshes, so it makes sense to also preserve the directory itself.

PR #473, opened 2025-09-04
Check LXD version before using it as a backend

dmitry-lyfar:
# Description

Check LXD version before using it as a backend.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/overlord.go
+++ b/internal/overlord/overlord.go
@@ -104,8 +104,6 @@ func New(dir string, restartHandler restart.Handler) (*Overlord, error) {
 		inited:   true,
 	}
 
-	var err error
-
 	if !filepath.IsAbs(dir) {
 		return nil, fmt.Errorf("directory %q must be absolute", dir)
 	}
@@ -193,10 +191,7 @@ var timeNow = time.Now
 func (m *Overlord) StartOfOperationTime() (time.Time, error) {
 	var opTime time.Time
 	err := m.State().Get("start-of-operation-time", &opTime)
-	if err == nil {
-		return opTime, nil
-	}
-	if err != nil && !errors.Is(err, state.ErrNoState) {
+	if !errors.Is(err, state.ErrNoState) {
 		return opTime, err
 	}
 
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -1,6 +1,7 @@
 package lxdbackend
 
 import (
+	"cmp"
 	"context"
 	_ "embed"
 	"encoding/json"
@@ -26,6 +27,7 @@ import (
 	"github.com/canonical/workshop/internal/osutil"
 	"github.com/canonical/workshop/internal/revert"
 	"github.com/canonical/workshop/internal/sdk"
+	"github.com/canonical/workshop/internal/syscheck"
 	"github.com/canonical/workshop/internal/workshop"
 )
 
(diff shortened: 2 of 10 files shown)


internal/overlord/overlord.go
```diff
@@ -196,7 +194,7 @@ func (m *Overlord) StartOfOperationTime() (time.Time, error) {
 	if err == nil {
 		return opTime, nil
 	}
-	if err != nil && !errors.Is(err, state.ErrNoState) {
+	if !errors.Is(err, state.ErrNoState) {
 		return opTime, err
 	}
```
jonathan-conder: I think `if err == nil` is also unnecessary

internal/workshop/lxd/lxd_backend.go
```diff
@@ -125,6 +126,43 @@ To restart the workshop daemon: 'sudo snap restart workshop'`, err)
...
+	info, _, err := conn.GetServer()
+	if err != nil {
+		return err
+	}
+
+	if err := checkVersion(info.Environment.ServerVersion); err != nil {
+		return err
+	}
+
+	return nil
```
jonathan-conder:
```suggestion
	return checkVersion(info.Environment.ServerVersion)
```

tests/main/compatible-backend/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Test workshopd checks LXD compatibility
+restore: |    
+  snap refresh lxd --channel=6/stable
+execute: |
+  . "$TESTSLIB"/utils.sh
+
+  workshop_exec list | MATCH "^.+[[:space:]]+comp-check[[:space:]]+Off[[:space:]]+-$"
+
+  snap refresh lxd --channel=5.21/stable
+  ! workshop_exec list 2>&1 | MATCH "^error: incompatible backend: LXD server version .+ is not supported; required >= 6.3.\*$"
```
jonathan-conder: I don't think downgrading LXD works well, I did it myself recently and a lot of things broke. See https://documentation.ubuntu.com/lxd/latest/installing/#upgrade-lxd

We should deprioritise this test and maybe purge the snap and install it from scratch, rather than refreshing.

jonathan-conder (commented):
Couple of minor things, the main thing is I don't think it's safe to refresh to an older version of LXD

internal/workshop/lxd/lxd_backend.go
```diff
@@ -148,7 +194,8 @@ func New() (*Backend, error) {
 	if err != nil {
 		return nil, err
 	}
-	if idx := slices.IndexFunc(pools, func(p api.StoragePool) bool { return p.Name == storagePool }); idx < 0 {
+	idx := slices.IndexFunc(pools, func(p api.StoragePool) bool { return p.Name == storagePool })
+	if idx < 0 {
```
jonathan-conder: I think this one and a few below were already the right style

jonathan-conder (approved):
Yeah I think this makes sense. My main concern is that being in degraded mode doesn't seem to block GET requests, or prevent the backend from being initialised. I think it works for the purpose of improving error messages for new users, at least.

dmitry-lyfar:
@jonathan-conder the only command which does not start with POST and uses the backed is `workshop list --global`. It is going to return an empty list. Hence, I considered it safe to not touch the original condition on GET so a user can see tasks, changes and warnings.

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: It felt a little longer for comfortable reading, so I experimented. Reverted back.

jonathan-conder: I feel like that's a good enough reason to bend the rules a bit

PR #475, opened 2025-09-08
Add write permissions to the system SDK files and directories

jonathan-conder:
# Description

Without this, if workshopd is running as a non-superuser, importing the system SDK fails when tar tries to remove the temporarily extracted filed.

Since this only affects Workshop devs, it seems unnecessary to update the revision number for this change.

I took the simple path of archiving the system SDK files one by one, changing the mode as we go. I think the `FS` interface is still too unstable to attempt writing a comprehensive wrapper.

Note that `tar.Writer.AddFS` recently added symlink support, so I upgraded the Go version we use to follow suit. Even if this sees further changes, I think the logic I reimplemented will be robust enough for a while.

The Go upgrade simplifies a bit of the `os.Root` code, and I noticed I somehow forgot to close it, which is now fixed.

I had some trouble with `pre-commit`, since `golangci-lint` is still on `1.24`. In order to make it use `1.25` I had to refresh my `go` snap to `latest/edge`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/.github/workflows/cover.yaml
+++ b/.github/workflows/cover.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.24"
+          go-version: "1.25"
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,7 +17,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.24"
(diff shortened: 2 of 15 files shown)


jonathan-conder:
@dmitry-lyfar the Workshop snap isn't going to build until the Go snap moves to 1.25. So I say we just wait for that

dmitry-lyfar (approved):

PR #478, opened 2025-09-11
Reduce dpkg fsync calls

jonathan-conder:
# Description

Taken [from LXD](https://github.com/canonical/lxd/blob/5a2f6640c0c20e16792360cf2a223ce04d597abf/doc/lxd-test.yaml#L23). This option can improve unpacking performance, but if the system crashes there's a chance some files are truncated. This seems to be a reasonable trade-off for ephemeral environments like workshops.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/workshop/lxd/lxd_backend.go
+++ b/internal/workshop/lxd/lxd_backend.go
@@ -1123,6 +1123,12 @@ apt:
     # Bypass confirmation prompts
     APT::Get::Assume-Yes "1";
 write_files:
+# Sacrifice some durability for performance:
+# https://wiki.debian.org/Teams/Dpkg/FAQ#Q:_Why_is_dpkg_so_slow_when_using_new_filesystems_such_as_btrfs_or_ext4.3F
+- content: |
+    # Managed by workshop, do not remove
+    force-unsafe-io
+  path: /etc/dpkg/dpkg.cfg.d/unsafe-io
 - content: |
     # Managed by workshop, do not remove
     [Unit]


jonathan-conder:
I did a bit of testing on this one, but my results are inconclusive. I think I should have used `refresh --restore` rather than launching a new workshop every time, to remove the effect of variable network speeds.

dmitry-lyfar:
Smoke tests for `workshop refresh` for `ROS2` SDK on `ubuntu@22.04` (apt cache size is 212M). Enforced re-run of the `setup-base` by adding / removing the `go` SDK before `ros2` in the list.

## ZFS volume is a loop-back file

### With the patch

Times for `setup-base`:
- 1m13.97s
- 1m32.577s
- 1m35.104s

### Without the patch

Times for `setup-base`:
- 1m40.015s
- 1m39.857s
- 1m32.594s

## ZFS volume is on an SSD partition

### With the patch

Times for `setup-base`:
- 1m3.25s
- 1m2.094s
- 59.059s

### Without the patch

Times for `setup-base`:
- 1m13.668s
- 1m13.516s
- 1m12.487s

dmitry-lyfar (approved):

dmitry-lyfar:
Decided not to merge it as the benefit does not justify the default setting. Will need to see more evidence from the actual use cases where this behavior makes sense as a default one. ROS2 is one of the most "deb-intensive" SDKs, which means other environments would see even less of an advantage.

PR #479, opened 2025-09-11
Expand $SDK variable in mount plug targets

jonathan-conder:
# Description

I already started allowing this in SDKcraft, mostly to fix a mount slot validation bug quickly. But it looks like upcoming SDKs are going to want mount points in the SDK directory, so it makes sense to add `$SDK` for mount plugs in Workshop instead of removing it from SDKcraft.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/reference/definition-files/workshop-definition.rst
+++ b/docs/reference/definition-files/workshop-definition.rst
@@ -267,6 +267,7 @@ Mount interface plugs can't belong to the :ref:`system SDK <ref_system_sdk>`.
 They are described by the following attributes:
 
 .. @artefact mount interface attributes
+.. @artefact $SDK
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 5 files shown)


docs/reference/definition-files/workshop-definition.rst
```diff
@@ -280,7 +281,9 @@ They are described by the following attributes:
    * - :samp:`workshop-target` (required)
      - string
      - A path inside the workshop
-       to be used as the plug's target directory.
+       to be used as the plug's target directory;
+       :file:`/project` or :envvar:`$SDK`-based paths can be used;
```
```suggestion
       :file:`/project/` or :envvar:`$SDK`-based paths can be used;
```

For visibility, it's best to end directory names with a slash.

docs/reference/sdks.rst
```diff
@@ -239,7 +239,8 @@ Mount interface
 
 .. @artefact mount interface
 
-A mount plug in the definition must specify the plug name, the interface, the target directory, its permissions and ownership, and optionally whether to be read-only:
+A mount plug in the definition must specify the plug name, the interface, and the target directory.
+It can optionally specify permissions and ownership for the target, and whether it is read-only:
```
```suggestion
The plug can specify permissions and ownership for the target, and whether it is read-only:
```

The verb 'can' already implies the optionality semantics; also, the pronoun here is ambiguous, immediately following a list.

akcano (changes requested):
Minor improvements.

dmitry-lyfar (approved):

PR #483, opened 2025-09-15
Doc: Add 'SDK best practices'

akcano:


docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+.. code-block:: yaml
+   :caption: java/sdk.yaml
+
+   parts:
+     java-part:
+       plugin: nil
+
+
+This minimal part definition relies entirely on the :samp:`setup-base` hook
+to install packages via the system package manager.
```
dmitry-lyfar: This will be changed by the toolchain team, also, we do not require nil parts now. There are SDKs that can ship all their logic as hooks and I think it should be just a note here.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+   :caption: dotnet10/sdk.yaml
+
+   plugs:
+     nuget-cache:
+       interface: mount
+       workshop-target: /home/workshop/.nuget
+     dotnet-cache:
+       interface: mount
+       workshop-target: /home/workshop/.dotnet
+
```
dmitry-lyfar: Let's have a tunnel example that exposes a service here instead. The paragraph is about the interfaces layout, so I think we need to show some other common use cases, not just the mount. 

The `uv` SDKs looks good to highlight the first point from your intro above -- data persistence.
We should add an example that supports another one -- sharing needs. There is a separate section on it in the end of the document, but those example miss the point of content sharing, I will comment below.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+by setting :samp:`UV_LINK_MODE=copy` in its :samp:`setup-project` hook.
+This configuration prevents linking issues
+when using mounted cache directories,
+demonstrating how environment variables can address
+the interaction between SDK behavior and workshop architecture:
+
+.. code-block:: shell
+   :caption: uv/hooks/setup-project
+
+   cat <<EOF >> ~/.profile
```
dmitry-lyfar, line 242: An example from node js SDK (so we don't use `uv` only):

```
cat <<EOF >/etc/profile.d/nodejs.sh
export PATH="$SDK/bin:\$PATH"
EOF
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+
+The :samp:`go` SDK shows another pattern,
+setting up the module cache path to work with the mounted directory:
+
+.. code-block:: shell
+   :caption: go/hooks/setup-base
+
+   cache=/home/workshop/go/pkg/mod
+   sudo -iu workshop -- go env -w GOMODCACHE="$cache"
```
dmitry-lyfar: It's not a pattern, just using a variable for the default cache location here. We can drop this one.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+
+For a complete list, see :ref:`exp_interfaces`;
+for a discussion of interface capabilities, see :ref:`exp_interface_concepts`.
+
+
+Environment variables
+---------------------
+
+Environment variables provide a clean way
```
dmitry-lyfar, line 214: I think what's important to say in this intro instead is that Workshop containers are system containers and, therefore, SDKs should use the standard means for POSIX-compatible shells to add variables to the `workshop exec` and `workshop shell` environment (both are login and (non-) interactive shell sessions), i.e. `/etc/profile`, `/etc/profile.d/`, `~/.profile`. Users should avoid using shell-specific configuration files, e.g. `~/.bash_profile`.

@jonathan-conder please lmk if I missed anything here.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+particularly in :samp:`setup-base`,
+because they integrate with the system package manager
+and can leverage :program:`apt`'s local cache.
+Installing packages during SDK build
+would bypass distribution security updates
+and create larger SDK artifacts.
+
+The :samp:`java` SDK exemplifies this approach:
+
+.. code-block:: shell
```
dmitry-lyfar, line 286: Java SDK will likely change from there. I understand that you probably want to make the examples neutral, but here we can use something like:

```
apt-get update
apt-get install ros-dev-tools
apt-get install ros-jazzy-ros-environment ros-jazzy-ros-workspace ros-jazzy-ament-index-cpp ros-jazzy-ament-index-python
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+           uv pip install torch torchvision torchaudio
+           ;;
+   esac
+
+
+Shared objects that require :command:`ldconfig` updates
+are best shipped with the SDK as parts.
+These libraries become part of the system immediately upon SDK installation,
+avoiding the complexity of dynamically updating the linker cache
+and ensuring consistent library resolution.
```
dmitry-lyfar: This should probably be moved under `setup-project` vs `setup-base` for a more nuanced discussion as this section discusses the difference between using parts and hooks.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+fundamentally affects when and how SDK initialization occurs.
+This decision impacts performance, caching behavior,
+and the SDK's interaction with workshop snapshots.
+
+The :samp:`setup-base` hook runs once when the SDK is installed
+at launch or refresh time,
+making it ideal for system-wide configuration
+that doesn't change between projects.
+Operations in :samp:`setup-base` become part of workshop snapshots,
+improving startup performance for subsequent launches.
```
dmitry-lyfar:
```suggestion
improving startup performance for subsequent refreshes.
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+However, actually starting anything in :samp:`setup-base` is not recommended,
+as the workshop might not need the SDK immediately;
+moreover, connected interfaces might not be available yet,
+potentially causing failures.
+Finally, workshops can be started and stopped multiple times;
+the :samp:`setup-base` hook will not run again during these transitions.
```
dmitry-lyfar:
```suggestion
```
This is true for `setup-project` too. I think there are legitimate use cases for running a daemon after `setup-base`, so it should be OK.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+The :samp:`setup-project` hook runs in user mode after :samp:`setup-base`,
```
dmitry-lyfar:
```suggestion
The :samp:`setup-project` hook runs as a `workshop` user after :samp:`setup-base`,
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+.. code-block:: shell
+   :caption: uv/hooks/setup-project
+
+   cat <<EOF >> ~/.profile
+
+   # SDK uses 'mount' interface to preserve
+   # uv cache across refreshes, thus, hardlinking is
+   # not available on 'uv sync'.
+   export UV_LINK_MODE=copy
+   EOF
```
dmitry-lyfar:
```suggestion
```
I suggest we move the example with dynamic GPU detection from the above section here.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+
+In general, you use :samp:`setup-base` for:
```
dmitry-lyfar, line 427: This is a nice summary.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+
+The :samp:`dotnet10` SDK shows an even more compact approach:
+
+.. code-block:: shell
+   :caption: dotnet10/hooks/check-health
+
+   #!/usr/bin/bash
+   dotnet --info && workshopctl set-health okay || workshopctl set-health --code=installation-failed error "dotnet --info failed"
+
```
dmitry-lyfar:
```suggestion
```
It's OK to use the `java` example, the rest just repeat the same logic in subtle variations.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+       if lspci | grep -i 'NVIDIA' >/dev/null 2>&1; then
+           GPU_TYPE="nvidia"
+       elif lspci | grep -i 'AMD/ATI' >/dev/null 2>&1; then
+           GPU_TYPE="amd"
+       elif lspci | grep -i 'Intel.*Graphics' >/dev/null 2>&1; then
+           GPU_TYPE="intel"
+       fi
+   fi
+
+   echo "Detected GPU: $GPU_TYPE"
```
dmitry-lyfar: Why GPU detection from `setup-project` in `check-health` section?

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+
+Content sharing
+---------------
+
+The usefulness of mount interfaces extends beyond their host-based capabilities.
+They also enable sophisticated content sharing patterns between SDKs
+while maintaining security boundaries.
+This mechanism is particularly powerful
+for creating ecosystem-wide optimizations.
```
dmitry-lyfar:
```suggestion
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+
+For instance, Python-based SDKs can share package caches
```
dmitry-lyfar: The examples in this section do not demonstrate content sharing between SDKs within a workshop as per the intro. They are all going to be workshop->host bind mounts.

The pattern that we want to describe here is when an SDK exposes some libraries, executables or environments that other SDKs need to work properly. We use this pattern to avoid introducing explicit dependencies and dependency management between SDKs. So, maybe, the title should say "Dependencies between SDKs" instead.

You can find examples in `zephyr` and `comfy` workshops here: https://github.com/canonical/reference-workshops/tree/main

A simple scenario that demonstrates it is the Python reference SDKs set. Suppose you have multiple SDKs that install packages from PyPI. You need them to do it into a single venv. Any SDK can provide such a venv, e.g. `uv` which creates a default venv. That means the SDK exposes a slot (see https://github.com/canonical/sdks/blob/c32c82942a1b5817ab4d4c3473a3c6b321eb04cf/uv/sdk.yaml#L32).

Any Python SDK that uses venvs, should provide a plug that can be connected to the venv slot. E.g. https://github.com/canonical/sdks/blob/c32c82942a1b5817ab4d4c3473a3c6b321eb04cf/jupyter/sdk.yaml#L12

Thus, when both added to the workshop, `jupyter` need to connect to `uv` so that it uses the venv provided by the latter:

```
connections:
  - plug: jupyter:venv
    slot: uv:venv
```

When `setup-project` for jupyter runs it will install packages in the venv provided by `uv`.

dmitry-lyfar (approved):

dmitry-lyfar (changes requested):
Left comments after the first pass. I think this also affects the navigation and other design related sections in the docs, let's discuss on the call.

docs/explanation/sdks/best-practices.rst
jonathan-conder: That sounds like the current state of affairs to me.

I guess we can recommend `/etc/profile.d` for system-wide variables, but `~/.profile` is better if something only makes sense for the workshop user, like `PATH="$HOME/<something-installed-by-setup-project>/bin:$PATH"`.

I guess not having `~/.profile.d` is a reason we should add support for `~/.config/environment.d` at some point.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+The :samp:`setup-project` hook runs in user mode after :samp:`setup-base`,
+when interfaces are connected and the workshop is fully operational.
+This makes it suitable for project-specific initialization
+that might vary depending on the actual project files.
+
+The :samp:`uv` SDK uses both hooks strategically:
+:samp:`setup-base` configures system-wide shell completion
+and :envvar:`PATH` updates:
+
+.. code-block:: shell
```
dmitry-lyfar, line 353: There is a new version of this `setup-base`

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+- System package installation
+- Global environment configuration
+- One-time setup operations
+- Content that should be part of snapshots
+
+
+Choose :samp:`setup-project` for:
+
+- Project-specific configuration that depends on the project context
+- Operations requiring connected interfaces
```
dmitry-lyfar: auto-connected (?)

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,578 @@
...
+Checking version availability can be simpler but equally effective.
+The :samp:`go` SDK demonstrates a more concise approach to health checking:
+
+.. code-block:: shell
+   :caption: go/hooks/check-health
+
+   if sudo -iu workshop -- go version; then
+     workshopctl set-health okay
+   else
+     workshopctl set-health --code=go-version-failed error "Go version check failed"
```
dmitry-lyfar: Check `workshop info` for this error message when launched with `--wait-on-error`.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+     user-service:
+       plugin: dump
+       source: ollama.service
+       source-type: file
+
+
+The corresponding service file shows appropriate service configuration:
+
+.. code-block:: ini
+
```
dmitry-lyfar, line 57:
```suggestion

.. code-block:: ini
   :caption: ollama/jupyter.service
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+Interfaces define how SDKs interact with the host system and other SDKs.
+The layout of interfaces ultimately impacts an SDK's usability and security.
+Publishers should select interfaces
+based on the resources their SDK requires (via plugs) or exposes (via slots).
+
+In particular,
+the :ref:`mount interface <exp_mount_interface>` plugs are frequently used
+because they specifically address data persistence and sharing needs.
+The :samp:`uv` SDK demonstrates this by mounting :file:`/home/workshop/.cache/uv`
+to preserve package caches across workshop operations,
```
dmitry-lyfar:
```suggestion
to preserve package caches across workshop life-cycle,
improving performance for `workshop refresh`:
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+   # SDK uses 'mount' interface to preserve
+   # uv cache across refreshes, thus, hardlinking is
+   # not available on 'uv sync'.
+   export UV_LINK_MODE=copy
+   EOF
+
+
+Publishers should avoid shell-specific configuration files,
+such as :file:`~/.bash_profile` or :file:`~/.bashrc`,
+because |ws_markup| supports multiple shell environments
```
dmitry-lyfar:
```suggestion
because |ws_markup| supports multiple shell interpreters
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+   apt-get install ros-dev-tools
+   apt-get install python3-colcon-argcomplete python3-colcon-alias python3-colcon-clean python3-colcon-mixin
+   # ...
+
+
+Shared objects that require :command:`ldconfig` updates
+are best shipped with the SDK as parts.
+These libraries become part of the system immediately upon SDK installation,
+avoiding the complexity of dynamically updating the linker cache
+and ensuring consistent library resolution.
```
dmitry-lyfar:
```suggestion
Shared objects that require :command:`ldconfig` updates
are best shipped with the SDK as parts.
```
The part I suggest to remove from this paragraph is also true for deb packages. The `ldconfig` is not a rationale to justify shipping as a part, it's rather a means to do it in a conventional way (and without fiddling  with `LD_LIBRARY_PATH`).

I think we should change reasoning here a little bit to be in line with the `uv` example below. The reasons you want to ship binaries in an SDK are:
- Pin their versions (e.g. always ship an LTS of a given tool no matter what's in package repositories or your development environment requires the exact version of a package / tool / framework / lib).
- Build them yourself (e.g. custom build of opencv that's not available from apt repositories or have custom compilation flags). 
- Outdated versions or non-existing versions in package repositories.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+       nvidia)
+           uv pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu129
+           ;;
+       amd)
+           uv pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.4
+           ;;
+       *)
+           uv pip install torch torchvision torchaudio
+           ;;
+   esac
```
dmitry-lyfar, line 424:
```suggestion
   case "$GPU_TYPE" in
       nvidia)
           pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cu129
           ;;
       amd)
           pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.4
           ;;
       *)
           pip install torch torchvision torchaudio
           ;;
   esac
```

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+-------------
+
+Health check scripts provide essential feedback
+about SDK operational status and help users diagnose problems quickly.
+Well-designed health checks go beyond simple binary success or failure,
+reporting extra details to provide actionable diagnostic information.
+
+The :samp:`java` SDK demonstrates effective health check design:
+
+.. code-block:: shell
```
dmitry-lyfar: I think that `java` SDK health hook is not designed in a way it is claimed in the paragraph above :-) It does not provide any actionable diagnostic information.

The one that does so is `ollama`'s below. We should probably leave that one only.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+
+   [Service]
+   ExecStart=/bin/bash -lc "ollama serve"
+   Restart=always
+   RestartSec=3
+
+
+Finally, services that require persistent state
+should leverage the :ref:`mount interface <exp_best_interfaces>`
+to store data outside the workshop environment.
```
dmitry-lyfar:
```suggestion
```
It's true, but it's also true for the entire SDK: if you want persistent state -- use mount OR *-state hooks, so maybe we should state it elsewhere (we do it in the `Interface layout` already)

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+
+System services
+---------------
+
+System services within SDKs should be designed to integrate smoothly
+with the workshop's lifecycle and other SDK components.
+Most services benefit from being managed through :ref:`hooks <exp_sdk_hooks>`
+rather than being started automatically during SDK installation.
+The reason is that installation occurs only once,
+while workshops may start and stop multiple times.
```
dmitry-lyfar:
```suggestion
```
I think it's not entirely correct. The services can only be managed from hooks (including their start), they cannot start automatically during SDK installation. This is what you explain below anyway.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,584 @@
...
+
+Consider the approach used by the :samp:`ollama` SDK:
+it implements a :samp:`setup-project` hook
+that configures and starts the service only when the workshop requires it,
+rather than running continuously in the background.
```
dmitry-lyfar:
```suggestion
that configures and starts the `systemd` service.
```
The service will be running in the background continuously once installed and enabled in `setup-project`, just like a regular `systemctl --user` service should. 

To do that, Workshop, creates a lingered user-session for the workshop user, when a workshop is started (as otherwise user-level services won't start), but maybe that's not the detail to mention here. 

It's worth to quote the actual `setup-project` below:

```
install -D --mode=644 --target-directory ~/.config/systemd/user "$SDK/ollama.service"

systemctl --user daemon-reload
systemctl --user enable --now ollama
```

dmitry-lyfar (commented):
Nice, I think we're close.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,586 @@
...
+
+- Handle edge cases gracefully
+
+
+.. _exp_best_dependencies:
+
+SDK dependencies
+----------------
+
+SDK dependencies enable sophisticated collaboration patterns between SDKs
```
dmitry-lyfar:
```suggestion
The `mount` interface enables sophisticated collaboration patterns between SDKs
```
As we do not have "SDK dependencies", it is enabled by the interface.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,586 @@
...
+through corresponding plugs.
+The :samp:`jupyter` SDK shows this pattern:
+
+.. code-block:: yaml
+   :caption: jupyter/sdk.yaml
+
+   plugs:
+     venv:
+      interface: mount
+      workshop-target: /home/workshop/.venv
```
dmitry-lyfar:
```suggestion
      workshop-target: $SDK/venv
```
I think SDKs should bind mount their environment under `$SDK` to avoid name collisions in $HOME if multiple SDKs would want to given their environment a common name.

docs/explanation/sdks/best-practices.rst
```diff
@@ -0,0 +1,586 @@
...
+
+- Pin specific versions regardless of what's available in package repositories,
+- Distribute custom builds with specialized compilation flags,
+- Provide tools that aren't available through the system package manager.
+
+
+This approach ensures consistent environments
+and avoids eventual dependency conflicts.
+For instance, shared objects that require :command:`ldconfig` updates
+are best shipped with the SDK as parts.
```
dmitry-lyfar:
```suggestion
```
I still believe it is not true, as many dev deb packages would call `ldconfig`, e.g. `flex`, `libc` and so on.

dmitry-lyfar (approved):
Looks good, a few more notes to update, but ready to go afterwards.

dmitry-lyfar (approved):

PR #490, opened 2025-09-18
Update `project-tools` to install tools in `setup-base`

dmitry-lyfar:


.workshop/tools/hooks/setup-base
```diff
@@ -1,3 +1,6 @@
 apt-get update
 apt-get install shellcheck
 
+curl -L https://github.com/fsouza/fake-gcs-server/releases/download/v1.52.3/fake-gcs-server_1.52.3_Linux_amd64.tar.gz | tar -xz -C /usr/local/bin fake-gcs-server
+curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b /usr/local/bin v2.4.0
```
jonathan-conder, line 5: I'd feel a bit more comfortable if we did some checksumming here

jonathan-conder (approved):

.workshop/tools/hooks/setup-base
dmitry-lyfar: Two possible options to consider in the future:
- SBOM for the artefacts so we can automate checking in `check-health`.
- Moving out as a store SDK.

PR #491, opened 2025-09-18
Update base image before launching or rebuilding a workshop

jonathan-conder:
# Description

When launching a new workshop, or rebuilding one from an image (i.e. when refreshing after changing the `base`), Workshop will now attempt to download the latest image for the given `base`. This PR does not affect refreshes which restore from SDK snapshots (but this is planned as the next step).

To be safe I would probably clean up all local images before migrating to this version of Workshop. It will probably break wait-on-error launches and some refreshes.

There are some known issues with managing LXD images, most notably https://github.com/canonical/lxd/issues/16515. I don't know of a viable workaround, but we haven't observed it in the wild yet, and I don't think this PR does anything to make it worse. It only triggers when another LXD project uses the same images as us. On my machine the other projects are `snapcraft` and `sdkcraft`, and these both use a different image server (`buildd`).

The way image downloads work is:
1. `WorkshopManager` creates a `download-base` task if there's no SDK snapshot to restore from.
2. `WorkshopManager` passes the ID of that task to `create-workshop`.
3. The `download-base` task downloads the latest image and saves its fingerprint in the task data.
4. The `create-workshop` task obtains the fingerprint from the `download-base` task data and uses it to launch or rebuild a workshop.

In `download-base`, we query the image server before starting the download. This query is mostly redundant, we only use it to obtain the image size (and, for LXD servers, whether the image is public). If the image is updated before starting the download, the size might be inaccurate. We should fix this by adding numbers to LXD's progress reporting.

A workaround would be to use the fingerprint from the query for the download. This seems to be safe, since LXD's image servers seem to keep at least 2 images per alias at all times. However, the resulting local copy of the image won't have any provenance information. I don't think it's worth losing this in order to remove a tiny chance of inaccurate progress reporting.

Last thing I'll mention is that we now use aliases like `24.04` rather than `24.04/amd64`. This should be more reliable, see comments in code for more details.

See https://github.com/canonical/workshop/pull/492 for the included bug fixes.

Fixes some minor style issues, and adds support for more image download progress string formats.

From what I've seen in the documentation, I think it already describes the state of affairs we're aiming for with this series of PRs.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/backend.go
```diff
@@ -197,7 +197,7 @@ type Backend interface {
 	// rootfs and rebuild it from the workshop file's base image. The
 	// configuration and devices of the rebuilt workshop will be reset to the
 	// default one.
-	LaunchOrRebuildWorkshop(ctx context.Context, file *File) error
+	LaunchOrRebuildWorkshop(ctx context.Context, file *File, image string) error
```
dmitry-lyfar, line 200: `image`, is it an alias, fingerprint or? I think it's worth commenting.

internal/workshop/backend.go
```diff
@@ -133,7 +133,7 @@ type VolumeManager interface {
 }
 
 type BaseImageManager interface {
-	Download(ctx context.Context, base string, report *progress.Reporter) error
+	Download(ctx context.Context, base string, report *progress.Reporter) (string, error)
```
dmitry-lyfar, line 136: What is the returned string here (`image-id`?) ?

internal/overlord/workshopstate/handlers.go
```diff
@@ -53,7 +53,15 @@ func (m *WorkshopManager) doDownloadBase(task *state.Task, tomb *tomb.Tomb) erro
 		},
 	}
 
-	return m.backend.Download(ctx, base, reporter)
+	image, err := m.backend.Download(ctx, base, reporter)
```
dmitry-lyfar, line 56: The PR comment says: `In download-base, we query the image server before starting the download.`. It looks like we go straight into downloading here, is the comment dated?

dmitry-lyfar (approved):

internal/workshop/backend.go
jonathan-conder: It's a fingerprint. I didn't want to use `fingerprint` here since it might be specific to LXD. But can update the comment

internal/workshop/backend.go
jonathan-conder: same as above

internal/overlord/workshopstate/handlers.go
jonathan-conder: The query happens inside `Download`

PR #493, opened 2025-09-23
Incorporate base image in refresh plan

jonathan-conder:
# Description

Refresh now queries the image server at the same time as the Store. If the base image is outdated, the workshop is rebuilt from scratch based on the new image. This is the same behaviour as switching to a different base. If the current base is up to date, the workshop will likely be restored from a snapshot.

There's no cleanup logic for images at present. They will start to pile up over time, but the snap's remove hook should keep this under control for now.

## Implementation

I added a new method `GetBase` to `BaseImageManager`. It looks up the latest version of a given base. For the LXD backend, the version is a fingerprint, which is likely to be globally unique, but the API only assumes the version string is unique within a given base.

---

I also renamed `Download` to `DownloadBase`. It now accepts a version instead of returning one.

Originally I wanted `DownloadBase` to always download the latest image, for two reasons:
1. The fingerprint might expire between planning a refresh and downloading the image.
2. LXD includes an `UpdateSource` field in the local copy when downloading the latest image.

I don't think the first one is a major issue. Even the daily image servers keep at least 2 fingerprints alive at all times. So the fingerprint used in a refresh plan should be valid for at least 1 day.

The second one is not an issue for LXD, because we disable auto-updates for our images. But the `UpdateSource` field would have been convenient when cleaning up old images, and is currently useful for the image download integration tests.

I didn't want to continue using aliases, since the previous implementation had some concurrency issues and I don't know if they've been fixed in LXD. Also, we can't give different images the same alias.

Instead, we now add a `workshop-base` property to our images, similar to the existing user.workshop.* instance config options. I avoided the `user.workshop.*` format because LXD's filtering syntax treats dots like path separators. I don't think there's a high chance of name collisions, unless the image servers start supporting Workshop explicitly.

In order to apply the property, I had to use `CreateImage` instead of `CopyImage`. Note that `CopyImage` is already a client-side wrapper around `CreateImage`, so there's no difference on the server side. However, for LXD image servers with multiple addresses, `CopyImage` tries all the addresses, whereas `CreateImage` only uses the main address. As far as I can tell, servers only have multiple addresses when they listen on multiple network interfaces (on the same machine), so I don't think this makes much difference in practice.

The upside of getting `DownloadBase` to download a specific fingerprint is that it simplifies the API, and detaches it from the LXD implementation details. Given that both downsides are quite minor, I now think this is the right call.

---

I removed the `Base` field from the `Workshop` struct, and added `BaseVersion`. Note that `Name` and `Base` are already available in the `File` field. I opted not to remove `Name` since it is used much more heavily than `Base`. But we could consider embedding `File` into `Workshop` to preserve the current convenience without duplicating data.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/lxd/lxd_base_manager.go
```diff
@@ -134,35 +152,25 @@ func (b *Backend) download(ctx context.Context, op *downloadOp, base string) (fi
...
-	fingerprint, ok := target.Metadata["fingerprint"].(string)
-	if !ok {
-		return "", fmt.Errorf("fingerprint not found for %q", base)
-	}
-
-	return fingerprint, nil
+	return nil
 }
 
 func BaseImageSource(base string) (*api.ImageSource, error) {
```
dmitry-lyfar:
```suggestion
func baseImageSource(base string) (*api.ImageSource, error) {
```

internal/workshop/backend.go
```diff
@@ -133,9 +133,12 @@ type VolumeManager interface {
...
 type BaseImageManager interface {
-	// Download the latest image for the given base. On success, returns an
-	// image identifier that can be passed to LaunchOrRebuildWorkshop.
-	Download(ctx context.Context, base string, report *progress.Reporter) (string, error)
+	// Lookup the latest image for the given base. On success, returns a
+	// string that uniquely identifies the latest version. This can be
+	// passed to DownloadBase and LaunchOrRebuildWorkshop.
+	GetBase(ctx context.Context, base string) (string, error)
+	// Download the given version of the base image.
+	DownloadBase(ctx context.Context, base, version string, report *progress.Reporter) error
```
dmitry-lyfar:
```suggestion
	DownloadBase(ctx context.Context, base, fingerprint string, report *progress.Reporter) error
```
`version` is confusing (a base's version is `22.04`, `24.04`, etc.), we should probably just use `base-fingerprint` here to indicate uniqueness.

internal/workshop/lxd/lxd_base_manager.go
```diff
@@ -134,35 +152,25 @@ func (b *Backend) download(ctx context.Context, op *downloadOp, base string) (fi
...
+		return fmt.Errorf("%q download failed: %w", base, err)
 	}
 
 	copyop.AddHandler(func(o api.Operation) {
-		if o.Metadata == nil || aliasInfo.Size <= 0 {
+		if o.Metadata == nil || image.Size <= 0 {
 			return
 		}
-		if upd := handleLaunchUpdate(o.Metadata, int(aliasInfo.Size)); upd != nil {
+		if upd := handleLaunchUpdate(o.Metadata, int(image.Size)); upd != nil {
```
dmitry-lyfar:
```suggestion
		if upd := handleImageUpdate(o.Metadata, int(image.Size)); upd != nil {
```

internal/workshop/workshop.go
```diff
@@ -17,10 +17,11 @@ import (
 )
 
 var (
-	ConfigProjectId         = "user.workshop.project-id"
-	ConfigWorkshopFile      = "user.workshop.file"
-	ConfigWorkshopSdks      = "user.workshop.sdks"
-	ConfigProjectPathDevice = "workshop.project"
+	ConfigProjectId           = "user.workshop.project-id"
+	ConfigWorkshopFile        = "user.workshop.file"
+	ConfigWorkshopBaseVersion = "user.workshop.base-version"
```
dmitry-lyfar:
```suggestion
	ConfigWorkshopBaseVersion = "user.workshop.base-fingerprint"
```

dmitry-lyfar (commented):
I think the approach is fine, but am not sure about the interface and naming. Specifically, `version`.

internal/workshop/backend.go
jonathan-conder: I didn't want to use fingerprint since it feels specific to LXD. The only other context I've seen it in is SSH keys. But I took a look at https://en.wikipedia.org/wiki/Fingerprint_(computing) and I guess it makes sense. At least, it's better than `hash` (being less specific). Only other word I might go for is `ID` or `identifier`.

internal/overlord/workshopstate/handlers.go
```diff
@@ -33,11 +33,12 @@ func (m *WorkshopManager) doDownloadBase(task *state.Task, tomb *tomb.Tomb) erro
...
 	st := task.State()
-	var base string
+	var base, fingerprint string
 	st.Lock()
-	err = task.Get("workshop-base", &base)
+	err1 := task.Get("workshop-base", &base)
+	err2 := task.Get("workshop-base-fingerprint", &fingerprint)
 	st.Unlock()
-	if err != nil {
+	if err1 != nil || err2 != nil {
```
dmitry-lyfar: `cmp.Or`?

dmitry-lyfar (approved):

internal/overlord/workshopstate/handlers.go
jonathan-conder: I don't have a strong feeling either way. Might as well, given we're using it elsewhere

PR #495, opened 2025-09-24
Improve stash efficiency

jonathan-conder:
# Description

Uses cloned instances instead of snapshots alone, to enable instance-only copies for stashing and unstashing. This speeds up refresh for large workshops.

This is very much a breaking change, I suggest running the remove hook and maybe deleting the state before migrating to this branch. The "before" is important because this PR modifies the hook.

I also limited workshop and SDK names to 40 characters, following SDKcraft.

## Preserving workshop metadata when refresh fails

The first commit introduces a new task `rebuild-workshop`, which is like `create-workshop` but its undo handler is a no-op (and it doesn't remove the workshop if the do handler fails partway through). To make this work, `UnstashWorkshop` has to use `lxc copy --refresh` instead of `lxc copy`.

This removes the need to distinguish between "remove for good" and "remove but keep the DHCP lease." Instead, the workshop exists contiguously from `launch` to `remove`.

To avoid reintroducing DHCP instability, we simply give the stash a different MAC address from the main instance. For consistency, we avoid copying most `volatile.*` options to and from the stash.

## Design

SDK snapshots and the stash are both implemented as "layers," i.e. LXD instances which live in the `workshop-layers.<USER>` project. These instances are never started, they just store a clone of the workshop's `rootfs` and associated metadata like config options and devices.

I tried to confine the changes to the LXD backend and not change the API too much. One thing I did change is that `Snapshot` and `Restore` now take an SDK name instead of a `snapid`.

All SDK snapshots belong to a workshop or a stashed workshop. This means the behavioural changes are limited to:
- `Snapshot` is like `lxc copy` rather than `lxc snapshot`. LXD still creates a ZFS snapshot internally, but also creates a new instance based on the snapshot.
- `StashWorkshop` copies the workshop and all of its snapshots, as LXD instances. This is cheaper than before because there's no need to assemble the copied snapshots into a single filesystem. It's like `lxc copy --instance-only` rather than `lxc copy`.
- `Restore` removes the same snapshots as before (as does `LaunchOrRebuildWorkshop`). It's like `lxc copy --refresh` rather than `lxc restore`.
- `UnstashWorkshop` removes snapshots created during the refresh and restores missing snapshots from the stash. It's like `lxc copy --refresh` rather than `lxc copy`, thanks to the new `rebuild-workshop` task.
- `RemoveWorkshop` and `RemoveWorkshopStash` attempt to remove all snapshots owned by the workshop or stash.

I think the main downside of this approach is that most functions operate on several instances at a time. It's hard to recover if something goes wrong partway through. The global SDK cache should resolve most of these issues, but I think enabling it will require much bigger changes outside the LXD backend.

## Naming

For the global SDK cache, layers will likely be named using unique hashes. We can't do that yet because the backend doesn't have enough information to compute an appropriate hash. Instead SDK layers are named like `<SDK>-<RANDOM STRING>`. Stashed workshops and layers are prefixed by `stash-`.

The random string is 16 characters long, to avoid confusing it with a project ID. The longest possible name `stash-<40 chars>-<16 chars>` has length 63 exactly.

The previous scheme `<WORKSHOP>.<SDK>` can't be used as an instance name because it contains a `.`, and LXD limits instance names to 63 characters. To keep track of ownership, the workshop name and project ID are stored in layers as extra config options. SDK layers also store the SDK name. All layers have a type, either `sdk`, `stash` or `stash-sdk`. The latter will be removed when we switch to a global SDK cache.

I renamed the `workshop-stash.<USER>` project to `workshop-layers.<USER>`. It's probably possible to store all layers in the main `workshop.<USER>` project, if there's a reason to consolidate them.

## Performance

I tested performance using the following workshop:
```yaml
name: nav2
base: ubuntu@24.04
sdks:
  - name: project-ros2
  - name: project-nav2
```

To make `project-ros2` I extracted the ROS2 SDK, removed the `setup-project` hook and removed the snaps from `setup-base`. The other SDK is just:
```console
$ cat .workshop/nav2/sdk.yaml 
name: nav2
$ cat .workshop/nav2/hooks/setup-base 
echo 1
```
In between refreshes I incremented `echo 1` to `echo 2`, etc., so the Workshop is always restored from a `ros2` snapshot.

### Refresh times

Main branch:
```console
$ time workshop refresh
"nav2" refreshed

real	0m20.821s
user	0m0.198s
sys	0m0.068s
$ time workshop refresh
"nav2" refreshed

real	0m22.594s
user	0m0.296s
sys	0m0.101s
$ time workshop refresh
"nav2" refreshed

real	0m22.598s
user	0m0.275s
sys	0m0.093s
$ time workshop refresh
"nav2" refreshed

real	0m22.879s
user	0m0.254s
sys	0m0.083s
```
This branch:
```console
$ time workshop refresh 
"nav2" refreshed

real	0m18.656s
user	0m0.217s
sys	0m0.073s
$ time workshop refresh 
"nav2" refreshed

real	0m20.861s
user	0m0.248s
sys	0m0.100s
$ time workshop refresh 
"nav2" refreshed

real	0m20.086s
user	0m0.246s
sys	0m0.066s
$ time workshop refresh 
"nav2" refreshed

real	0m21.239s
user	0m0.278s
sys	0m0.080s
```

Looks like a small, but clear, improvement.

### Space consumed

Main branch, pre-refresh:
```
NAME                                              USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                               445M  27.7G    24K  legacy
workshop/containers/workshop.jono_nav2-bee72369   445M  27.7G   908M  legacy
workshop/deleted/containers                        24K  27.7G    24K  legacy
```
Main branch, mid-refresh:
```
NAME                                                          USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                                           890M  27.2G    24K  legacy
workshop/containers/workshop-stash.jono_stash-nav2-bee72369   446M  27.2G   908M  legacy
workshop/containers/workshop.jono_nav2-bee72369               445M  27.2G   908M  legacy
workshop/deleted/containers                                    24K  27.2G    24K  legacy
```
Main branch, post-refresh:
```
NAME                                              USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                               445M  27.7G    24K  legacy
workshop/containers/workshop.jono_nav2-bee72369   445M  27.7G   908M  legacy
workshop/deleted/containers                        24K  27.7G    24K  legacy
```
This branch, pre-refresh:
```
NAME                                                               USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                                               1.96M  27.7G    24K  legacy
workshop/containers/workshop-layers.jono_nav2-ffee8e20be0597bb     101K  27.7G   908M  legacy
workshop/containers/workshop-layers.jono_ros2-cf4dee669ba1a0c8     101K  27.7G   907M  legacy
workshop/containers/workshop-layers.jono_system-f982ce85d78e71d8    22K  27.7G   498M  legacy
workshop/containers/workshop.jono_nav2-bee72369                   1.72M  27.7G   908M  legacy
workshop/deleted/containers                                        445M  27.7G    24K  legacy
workshop/deleted/containers/4586daf1-c8c0-4d1c-b7ae-aa4c3086634f   445M  27.7G   907M  legacy
```
This branch, mid-refresh:
```
NAME                                                                     USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                                                     1.62M  27.7G    24K  legacy
workshop/containers/workshop-layers.jono_ros2-cf4dee669ba1a0c8           101K  27.7G   907M  legacy
workshop/containers/workshop-layers.jono_stash-nav2-bee72369             101K  27.7G   908M  legacy
workshop/containers/workshop-layers.jono_stash-nav2-ffee8e20be0597bb     101K  27.7G   908M  legacy
workshop/containers/workshop-layers.jono_stash-ros2-cf4dee669ba1a0c8     101K  27.7G   907M  legacy
workshop/containers/workshop-layers.jono_stash-system-f982ce85d78e71d8    22K  27.7G   498M  legacy
workshop/containers/workshop-layers.jono_system-f982ce85d78e71d8          22K  27.7G   498M  legacy
workshop/containers/workshop.jono_nav2-bee72369                         1.16M  27.7G   908M  legacy
workshop/deleted/containers                                              448M  27.7G    24K  legacy
workshop/deleted/containers/4586daf1-c8c0-4d1c-b7ae-aa4c3086634f         445M  27.7G   907M  legacy
workshop/deleted/containers/8786c42c-5f5c-4de9-a085-6abd6a15912d        2.34M  27.7G   908M  legacy
workshop/deleted/containers/da3e4e85-4a17-4c0b-ad4e-7a8229906bfb         101K  27.7G   908M  legacy
```
This branch, post-refresh:
```
NAME                                                               USED  AVAIL  REFER  MOUNTPOINT
workshop/containers                                               1.89M  27.7G    24K  legacy
workshop/containers/workshop-layers.jono_nav2-14e237260f0c5af6     101K  27.7G   908M  legacy
workshop/containers/workshop-layers.jono_ros2-cf4dee669ba1a0c8     101K  27.7G   907M  legacy
workshop/containers/workshop-layers.jono_system-f982ce85d78e71d8    22K  27.7G   498M  legacy
workshop/containers/workshop.jono_nav2-bee72369                   1.65M  27.7G   908M  legacy
workshop/deleted/containers                                        445M  27.7G    24K  legacy
workshop/deleted/containers/4586daf1-c8c0-4d1c-b7ae-aa4c3086634f   445M  27.7G   907M  legacy
```

In `main`, the stash consumes about the same space as the workshop. In this branch that overhead is almost completely gone. But there's a downside, which is that deleted containers can accumulate over time. This example isn't bad, but only because we didn't really change the workshop after installing `ros2`. It's quite easy to rack up a lot of wasted space.

We plan on working with the LXD team to address the issue. In the meantime, this script will remove the unnecessary filesystems:
```bash
set -eo pipefail

volumes() {
	zfs get -H -o value -d 1 -t filesystem,volume name workshop/deleted/containers
}

snapshots() {
	zfs get -H -o value -t snapshot name "$1"
}

clones() {
	zfs get -H -o value clones "$1"
}

volumes | while IFS= read -r volume; do
	[ "$volume" != workshop/deleted/containers ] || continue

	snapshots "$volume" | {
		latest=''
		while IFS= read -r snapshot; do
			clones="$(clones "$snapshot")"
			if [[ "$clones" =~ ^-?$ ]]; then
				zfs destroy "$snapshot"
			else
				latest="${clones%%,*}"
			fi
		done

		if [ -n "$latest" ]; then
			zfs promote "$latest"
		fi
		zfs destroy "$volume"
	}
done
```
NOTE: it assumes all snapshots are named `copy-<UUID>`. This is the case if they were created by Workshop. To make it more robust, it should probably rename the snapshots before calling `zfs promote`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


docs/explanation/workshops/concepts.rst
```diff
@@ -147,6 +147,8 @@ at launch time,
    and its :samp:`setup-base` hook is run.
 
 #. A ZFS snapshot is taken.
+   The snapshot is cloned
+   to decouple its lifetime from the workshop.
```
akcano: Needs a clarification of how this decoupling actually works, e.g. what are the implications of the clone operation in the circumstances.

akcano (changes requested):
LGTM, one minor change needed.

docs/explanation/workshops/concepts.rst
jonathan-conder: What do you make of this? https://github.com/canonical/workshop/compare/65c7f2f6e053f7eb9b46a72bfde05a109b3b3584..7db55bce7bbebaa6ab9c2bfd20426115495432e0

I think the current situation is inherently awkward to explain. It should make more sense once we have a global refresh cache!

internal/workshop/lxd/lxd_backend_layers.go
```diff
@@ -0,0 +1,395 @@
...
+	layers, err := layerConn.GetInstancesWithFilter(api.InstanceTypeContainer, filters)
+	if err != nil {
+		return "", nil, err
+	}
+
+	idx := slices.IndexFunc(layers, func(inst api.Instance) bool {
+		return inst.Config["user.workshop.sdk"] == sk
+	})
+	if idx < 0 {
+		return "", nil, fmt.Errorf("SDK snapshot %q not found in workshop %q", sk, w)
```
dmitry-lyfar:
```suggestion
		return "", nil, fmt.Errorf("%q SDK snapshot not found in workshop %q", sk, w)
```

jonathan-conder:
I think we'll need a transitional version of the `post-refresh` hook which also removes the stash project. Pre-refresh doesn't cut it because the daemon will still be running

internal/workshop/lxd/lxd_backend_layers.go
```diff
@@ -0,0 +1,395 @@
...
+		if _, ok := sdks[layer.Config["user.workshop.sdk"]]; !ok {
+			obstacles = append(obstacles, layer.Name)
+		}
+	}
+
+	return layers[idx].Name, obstacles, nil
+}
+
+// Copies the instance between LXD projects.
+func (s *Backend) copyInstance(src, dst lxd.InstanceServer, srcName, dstName string, refresh bool, config map[string]string, exclude []string) error {
```
dmitry-lyfar, line 288: This one would benefit from a more detailed commentary now given it is called by different clients. E.g. the `config` argument, what is it related to src or dst? Same for the `exclude`.

internal/workshop/lxd/lxd_backend_layers.go
```diff
@@ -0,0 +1,395 @@
...
+	for _, layer := range obstacles {
+		if err := s.deleteLayer(layerConn, layer); err != nil {
+			return err
+		}
+	}
+
+	instance := InstanceName(name, projectId)
+	config := map[string]string{workshop.ConfigWorkshopFile: string(f)}
+	exclude := []string{"user.workshop.sdk", "user.workshop.layer-type"}
+	return s.restoreLayer(conn, layerConn, instance, snapshot, config, exclude)
```
dmitry-lyfar: `restoreLayer`, `createLayer` and `copyLayer` should probably not accept config as an argument. These can encapsulate creating the config map to pass it on to the `copyInstance`. 

Also, I'm a bit lost in understanding which `config` is copied / excluded when and why. I think these functions would be a good place to have a commentary about the reason why a certain config is passed.

docs/explanation/workshops/concepts.rst
akcano: That's the level of detail I hoped for.
While it definitely may be puzzling for an outsider, we only have to have a place for the explanation of this logic at the moment.

internal/workshop/lxd/lxd_backend_layers.go
jonathan-conder: I went for a bit of a different approach. No 2 calls to copyInstance have the same `config` or `exclude`, so I don't think it's possible to encapsulate the logic without adding a big switch statement or something.

These 1-line wrapper functions seem to be an unnecessary abstraction though, so I removed them in favour of calling `copyInstance` directly. I was hesitant to allow direct calls previously, because `copyInstance` ignored `exclude` when `refresh=false`. So I just made it respect `exclude` regardless of `refresh`.

The config options are best understood in the context of paired operations. `Snapshot` adds some attributes so that `Restore` can use them to identify the snapshot. Also, `Restore` removes these artificial attributes when restoring.

I moved the `boot.autostart=false` override from `Snapshot` to `copyInstance`. It should be false for all copies, so we might as well override it even if we know it's already false in the source instance. That should make the symmetry between `Snapshot` and `Restore` more apparent (`user.workshop.file` notwithstanding).

dmitry-lyfar (approved):

PR #500, opened 2025-10-06
Add `sdk list` CLI command

dmitry-lyfar:
# Description

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/cmd/sdk/list.go
+++ b/cmd/sdk/list.go
@@ -1,0 +1,65 @@
+package main
+
+import (
+	"fmt"
+	"text/tabwriter"
+
+	"github.com/spf13/cobra"
+)
+
+type CmdList struct {
+	root *CmdRoot
+}
+
+func (c *CmdList) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List SDK volumes available on this machine",
+		Long: `
+This command lists all local SDK volumes.
+
+Use it to enumerate the SDK revisions currently stored on the system.
+Only volumes are reported, not the workshops that use them.
+
+Notes:
+
+- For per-workshop information, use 'workshop info'.
+- Multiple entries may appear for a single SDK
+  if several revisions are present simultaneously.
+`,
+		Args: cobra.NoArgs,
+		RunE: c.Run,
+	}
+
+	return cmd
+}
+
+func (c *CmdList) Run(cmd *cobra.Command, _ []string) error {
+	cli, err := c.root.client()
+	if err != nil {
+		return err
+	}
+
+	sdks, err := cli.Sdks()
+	if err != nil {
+		return err
+	}
+
+	w := tabwriter.NewWriter(Stdout, 4, 3, 2, ' ', 0)
+	fmt.Fprintf(w, "Name\tVersion\tRev\tSummary\n")
+	for _, sdk := range sdks {
+		version := sdk.Version
+		if version == "" {
+			version = "-"
+		}
+
+		summary := sdk.Summary
+		if summary == "" {
+			summary = "-"
+		}
+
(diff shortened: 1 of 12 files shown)


cmd/sdk/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+func main() {
+	l, err := logger.New(Stderr, 0)
+	if err != nil {
+		panic(err)
+	}
+	logger.SetLogger(l)
+
+	cmd := (&CmdRoot{}).Command()
+	cmd.SetArgs(slices.Clone(os.Args[1:]))
```
jonathan-conder, line 21: I forgot about this line. I think it's worth preserving the comment from `workshop`: https://github.com/canonical/workshop/blob/91ffede9d0e5fc5acfea99fea59e2c5d24e15679/cmd/workshop/main.go#L25-L26

Or maybe we don't need this workaround for the `sdk` command at all. I think it was mostly useful for the unusual arg parsing in `exec` and `run`.

cmd/sdk/root.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil, fmt.Errorf("cannot create client: %w", err)
+	}
+	c.cli = cli
+	return cli, nil
+}
+
+func (c *CmdRoot) postRun(cmd *cobra.Command, _ []string) {
+	if c.cli != nil && cmd.Name() != cobra.ShellCompRequestCmd {
+		c.cli.CloseIdleConnections()
+	}
```
jonathan-conder, line 56: Would this be useful for the other CLIs?

cmd/sdk/root.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+var (
+	Stdin  io.Reader = os.Stdin
+	Stdout io.Writer = os.Stdout
+	Stderr io.Writer = os.Stderr
+)
```
jonathan-conder, line 63: It's not something I would change only for `sdk`, but we can avoid these globals by using `cobra`, e.g.: https://pkg.go.dev/github.com/spf13/cobra#Command.OutOrStdout

internal/daemon/api_sdks.go
```diff
@@ -0,0 +1,53 @@
...
+func v1GetSdks(c *Command, r *http.Request, _ *userState) Response {
+	state := c.d.overlord.State()
+	state.Lock()
+	wb := c.d.overlord.WorkshopBackend()
+	state.Unlock()
+
+	volumes, err := wb.Volumes(r.Context(), "sdk")
+	if err != nil {
+		return statusInternalError("cannot list SDK volumes: %w", err)
+	}
```
jonathan-conder, line 27: Maybe there are some exceptions already, but I would expect the API to talk to the `SdkManager` rather than directly to the backend.

jonathan-conder (approved):
Looks good. Only real complaint is that `sdk list` doesn't do much on its own. But I realise there's more to come and it's good to get a basic skeleton ready.

cmd/sdk/list.go
```diff
@@ -0,0 +1,65 @@
...
+This command enumerates all local SDK volumes.
+
+Use it to inspect which SDK revisions are currently stored on the system. The
+command only reports volumes, not which workshops are using them.
+
+Notes:
+
+- For per-workshop information, use 'workshop info'.
+- Multiple entries may appear for the same SDK when several revisions are
+  present.
```
akcano:
```suggestion
This command lists all local SDK volumes.

Use it to enumerate the SDK revisions currently stored on the system.
Only volumes are reported, not the workshops that use them.

Notes:

- For per-workshop information, use 'workshop info'.
- Multiple entries may appear for a single SDK
  if several revisions are present simultaneously.
```

akcano (changes requested):

PR #502, opened 2025-10-07
Fail early if local SDK has no sdk.yaml

jonathan-conder:
# Description

Before:
```console
$ workshop launch
- Register "nav2-sdk" SDK plugs and slots (open /home/ubuntu/.local/share/workshop/id/967fc9ea/nav2/sdk/nav2-sdk/x1/meta/sdk.yaml: no such file or directory)
"nav2" launch aborted
```

After:
```console
$ workshop launch
error: cannot launch "nav2": sdk.yaml file not found for SDK "nav2-sdk"
```

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/workshopstate/request.go
+++ b/internal/overlord/workshopstate/request.go
@@ -409,7 +409,16 @@ func (s *localSdkFinder) commitRevision(wp *workshop.Workshop, w, sk, source str
 	}
 
 	target := workshop.LocalSdkDir(s.userDataDir, s.project.ProjectId, w, sk)
-	return sdk.CommitRevision(s.user, source, target, installed)
+	revision, err := sdk.CommitRevision(s.user, source, target, installed)
+	if err != nil {
+		return sdk.Revision{}, err
+	}
+
+	if !osutil.FileExists(filepath.Join(target, revision.String(), "meta", "sdk.yaml")) {
+		return sdk.Revision{}, fmt.Errorf("sdk.yaml file not found for SDK %q", sk)
+	}
+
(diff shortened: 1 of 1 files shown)


internal/overlord/workshopstate/request.go
```diff
@@ -409,7 +409,16 @@ func (s *localSdkFinder) commitRevision(wp *workshop.Workshop, w, sk, source str
 	}
 
 	target := workshop.LocalSdkDir(s.userDataDir, s.project.ProjectId, w, sk)
-	return sdk.CommitRevision(s.user, source, target, installed)
+	revision, err := sdk.CommitRevision(s.user, source, target, installed)
+	if err != nil {
+		return sdk.Revision{}, err
+	}
+
+	if !osutil.FileExists(filepath.Join(target, revision.String(), "meta", "sdk.yaml")) {
```
dmitry-lyfar, line 417: No sure I understand it fully, does it have to happen _after_ the new revision was committed?

jonathan-conder: We can check before the files are copied - then we'd have to check both sdk.yaml and meta/sdk.yaml. This could change between the check and the copy, but the only issue would be inconsistent error messages.

We could move the check inside `CommitRevision`, but then the error message won't know what the SDK is named.

dmitry-lyfar (approved):

PR #505, opened 2025-10-10
Use snap hooks to install completion scripts for bash, fish and zsh

jonathan-conder:
# Description

Snap's completion support is geared towards strictly confined snaps and only supports bash. It might be possible to adapt it to zsh by writing a zsh-style completion script which calls the bash-style one inside the snap. But I doubt it can be done in a robust way. Reimplementing the bash approach for both fish and zsh looks to be quite difficult.

Instead, since Workshop uses classic confinement, we can just install completion scripts on the host. Zsh looks for scripts in `$fpath`, which by default only has one entry outside `/usr/share`. However, bash and fish both look for completion scripts in `$XDG_DATA_DIRS`, so we can use `/var/lib/snapd/desktop` for them. This is where `snapd` itself installs its bash completion scripts, so it should be OK for us to use as well.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


snap/local/hooks/configure
```diff
@@ -1 +1,12 @@
 #!/bin/sh -e
+
```
dmitry-lyfar, line 2: It should be OK to call it from `configure` now. However, if we extend the snap with additional option, e.g. storage name, these may need to move to install / pre-refresh hooks.

dmitry-lyfar (approved):
It is a nice way to emulate an interactive session there.

snap/local/hooks/configure
jonathan-conder: I think `configure` is the only one which works when reverting the snap. I think reinstalling the completion scripts more often than necessary is better than sometimes not reinstalling them

akcano (approved):

PR #511, opened 2025-10-16
Hash Store, in-project and sketch SDKs

jonathan-conder:
# Description

We already hash SDKs published using `sdkcraft try`. This brings the other types of SDKs in line with that, with a view towards caching SDK snapshots across multiple workshops.

For Store SDKs, the GCS server only supports MD5 hashes and there's no easy way to download a specific SDK revision. Until we switch to the actual Store, SDKs will have either an MD5 or SHA3 hash. The hash is recomputed when downloading the SDK, since it might have changed since the initial query. Both the hash and SDK YAML are cached alongside the download, similar to try SDKs.

I added a validation step for all SDKs, which happens just before planning a launch or refresh. It checks that every SDK in the workshop has a hash and a valid `sdk.yaml`. Previously `sdk.yaml` was only parsed this early for Store SDKs. If the downloaded SDK differs from the plan, we validate the new version before installing it in the workshop.

For local SDKs, on launch/refresh the SDK files are copied to a temporary directory, to make them immutable. Instead of comparing this directory with others, we now compute a hash of the contents. The temporary directory is renamed to the hex digest (unless it already exists). If needed, a revision number is allocated, as a symlink (e.g. `x42 -> abcdef1234567890`). Otherwise the timestamp of the revision is bumped, using a new function `Lchtimes` to avoid following the link.

I fixed a minor bug in the previous local SDK logic: if a user deletes a file within an SDK just before it's copied to the temporary directory, Workshop now ignores the file instead of raising an error.

The hash of a directory is recursive. It accounts for filenames, contents, permissions and doesn't follow symlinks. The exact format is based on `git`:
- For regular files, the data we hash is `blob <file size>\0<file contents>`.
- For symlinks, the data is `blob <target length in bytes>\0<link target>`.
- For directories, the data is `tree <entries length>\0<entries>`, where `entries` is a sequence of items like `<file mode> <file name>\0<hash>`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/fakestore/store.go
+++ b/internal/fakestore/store.go
@@ -11,7 +11,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
 	"time"
 
@@ -19,7 +18,6 @@ import (
 	"github.com/googleapis/gax-go/v2"
 	"google.golang.org/api/option"
 
-	"github.com/canonical/workshop/internal/arch"
 	"github.com/canonical/workshop/internal/logger"
 	"github.com/canonical/workshop/internal/osutil"
 	"github.com/canonical/workshop/internal/progress"
@@ -85,25 +83,8 @@ func (c *GcsStore) SdkAction(ctx context.Context, actions []sdk.SdkAction) ([]sd
 				continue
 			}
 
-			info, err := sdk.ReadSdkInfo([]byte(s.SdkYAML), act.ProjectId, act.Workshop)
-			if err != nil {
-				actError.errors[act.Name] = err
-				continue
-			}
-			if info.Name != act.Name {
-				return nil, fmt.Errorf("SDK must be named %q (now: %q)", act.Name, info.Name)
-			}
-			if !slices.Contains([]string{"", act.Base}, info.Base) {
-				return nil, fmt.Errorf("%q SDK from %q has %q base; required: %q", act.Name, act.Channel, info.Base, act.Base)
-			}
-			if !slices.Contains([]string{"", "all", arch.DpkgArchitecture()}, info.Arch) {
-				return nil, fmt.Errorf(`%q SDK from %q has %q architecture; required: %q or "all"`, act.Name, act.Channel, info.Arch, arch.DpkgArchitecture())
-			}
-
-			info.Revision = s.Revision
-			info.Channel = s.Channel
-
-			results = append(results, sdk.SdkResult{Info: info})
+			setup := sdk.Setup{Name: s.Name, Channel: s.Channel, Revision: s.Revision}
+			results = append(results, sdk.SdkResult{Setup: setup, SdkYAML: s.SdkYAML})
 		default:
 			return nil, fmt.Errorf("unknown SDK store action")
 		}
--- a/internal/fakestore/store.go
+++ b/internal/fakestore/store.go
@@ -4,12 +4,15 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/md5"
+	"encoding/hex"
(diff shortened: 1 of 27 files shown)


internal/overlord/sdkstate/handlers.go
```diff
@@ -74,34 +74,42 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
...
-		if err = store.DownloadSdk(ctx, rec, reporter); err != nil {
-			return err
-		}
+		result, err = store.DownloadSdk(ctx, rec, reporter)
+	}
+	if err != nil {
+		return err
+	}
+
+	if result.Revision != rec.Revision {
```
dmitry-lyfar, line 119: Let's comment about the reason here. Also, why don't we validate the SDK's yaml here instead of in `doInstall`? Doing it there would contain this logic that handles appearing a new revision in  a single place.

jonathan-conder:
Marking this as ready now, but I will try to make the local SDK changes more back-compatible

internal/overlord/workshopstate/request.go
dmitry-lyfar, line 679: I'm guessing that for a new workshop there'll be no issue to create the required dir structure for the project SDKs, i.e. `<hash>/<revision>`, but it was the place you mean in the context of our conversation about snap hooks needing to remove existing project SDKs?

internal/sdk/local.go
```diff
@@ -262,12 +287,47 @@ func removeOldRevisions(path string, revisions []Revision, remove int) {
...
+	}
+	if err != nil {
+		return err
+	}
+
+	if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+
+	if strings.ContainsRune(digest, os.PathSeparator) {
```
dmitry-lyfar, line 317: Not sure I can see how can it contain a path separator?

jonathan-conder: This is just a security measure. Since we're `root` we don't want to let the user go in and do `ln -s /home/other-user x1` and have Workshop blindly delete anything on the system.

internal/overlord/workshopstate/request.go
jonathan-conder: Thanks to the snap remove hook we only need to worry about launch, not refresh. But an old directory might exist here since the snap hook doesn't currently remove it

internal/overlord/sdkstate/handlers.go
jonathan-conder: I wanted to do it here too, but we don't know the workshop's base at this point. Definitely worth a comment though

internal/fakestore/store.go
```diff
@@ -173,16 +175,27 @@ func (c *GcsStore) DownloadSdk(ctx context.Context, setup sdk.Setup, report *pro
...
 	}
 
-	if _, err = io.Copy(writer, r); err != nil {
-		return err
+	if _, err = io.Copy(io.MultiWriter(writers...), r); err != nil {
+		return nil, err
+	}
+
+	digest := hex.EncodeToString(hash.Sum(nil))
+	if err := os.WriteFile(target+".md5", []byte(digest+"\n"), 0666); err != nil {
```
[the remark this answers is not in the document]
jonathan-conder: This is no different from `os.Create`. Don't you know about `umask`?

dmitry-lyfar (approved):

dmitry-lyfar (approved):

PR #515, opened 2025-10-20
Use --machine option for systemctl instead of sudo

dmitry-lyfar:


jonathan-conder (approved):
Looks like a more robust approach

PR #522, opened 2025-10-23
Doc: Refactor starter pack SDK

akcano:


.workshop/dev.yaml
```diff
@@ -20,8 +20,21 @@ actions:
     make -C docs/ run
   docs-clean: |
     make -C docs/ clean-doc
+  docs-clean-all: |
+    make -C docs/ clean
   docs-check: |
     make -C docs/ {spelling,linkcheck,woke,lint-md}
+  docs-upgrade: |
```
dmitry-lyfar, line 27: I'm happy with the rest of PR except for this. We should work out a bit more robust approach to upgrading sphinx for a given project (and not having it in a workshop action, perhaps). 

Overall, can you introduce these changes to the starter pack SDK (https://github.com/canonical/sdks/tree/main/starter-pack) and use it here instead of having it as a project SDK to begin with?

We should explore the approach to upgrading after that in a separate PR.

akcano: Understood, closing this PR for now, then.

PR #530, opened 2025-11-07
Make backends aware of SDK semantics

jonathan-conder:
# Description

This PR reimplements state storage using bind mounts, and replaces the generic `VolumeManager` in the `Backend` with a specialised `SdkManager`. The `Backend` also has two new functions:
- `InstallSdk` replaces `Backend.AddWorkshopConfig` and `Workshop.AddSdk`.
- `UninstallSdk` replaces `Backend.RemoveWorkshopConfig` and `Workshop.RemoveSdk`.

I also removed the internal function `filterLxdInstancesByConfig` in favour of `GetInstancesWithFilter` (provided by LXD).

Overall, the `Backend` API is now a bit more high-level. It has more context about SDK-related operations, at the cost of being responsible for more of the details.

This can be a breaking change if you have a refresh waiting on error (since the state volumes have changed). Otherwise it should be compatible, as far as I know.

## Motivation

When snapshotting a workshop, it makes sense to remove SDK mounts and other devices, to avoid coupling the lifetime of the snapshot with other entities like SDK volumes. In some cases (e.g. in-project SDKs) the device's source depends on the workshop name, so there's not much point retaining the original device anyway.

It's not easy to remove LXD devices during a `CopyInstance` operation. The best we can do, for now, is set them to `{"type":"none"}`. I don't want to call `UpdateInstance` after creating a snapshot, since it's hard to handle errors in that case. When restoring a snapshot, it's easier to override the devices if they are named consistently across different workshops.

Currently device names are the same as volume names, which include revision numbers (even for local SDKs). SDKs with different revision numbers can have the same contents (e.g. `try` and Store SDKs). I want snapshots to be based on the SDK contents alone, so I'd like to remove the revision number from device names.

The goal of this PR is to decouple device names from volume names. It doesn't actually rename them yet. I think it makes sense to make the `Backend` responsible for device names, rather than adding another argument to `AttachVolume`. The approach I went with has a few other benefits:
- No need for `workshop.VolumeSetup`. Different types of volumes can be managed by different `Backend` subsystems with different function signatures. For SDKs, `sdk.Meta` replaces `workshop.VolumeSetup`.
- Installing an SDK is a single atomic operation.
- No need for `AddWorkshopConfig` and `RemoveWorkshopConfig`.

On the other hand, it makes the backend a bit more entangled with the rest of the codebase. Each backend has to determine how to translate SDK metadata into an underlying volume. This would be more of an issue if we had several different types of volumes. This PR reimplements state storage using bind mounts, so SDKs are now the only volumes, and I think the tradeoff makes sense.

I didn't touch `AddWorkshopMount` and `RemoveWorkshopMount`. These currently expose the LXD device names as arguments, which is a bit inconsistent with volumes and might not make sense for other backends. I might revisit this later on, but not yet since it doesn't impact the global SDK cache work. Currently these mounts are used for `workshopctl`, `/project` and now state storage. The first two could be reimplemented using interfaces if we had a `base` SDK. The latter could be done using a dedicated API similar to the new `SdkManager`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:


internal/workshop/lxd/lxd_backend_sdk.go
```diff
@@ -182,22 +165,22 @@ func (s *Backend) ImportVolume(ctx context.Context, info workshop.VolumeSetup, t
...
 
 	// Generate index.yaml for the volume.
-	if err = os.WriteFile(filepath.Join(dir, "index.yaml"), []byte(volumeIndexContent(info.Name)), 0644); err != nil {
+	if err = os.WriteFile(filepath.Join(dir, "index.yaml"), []byte(volumeIndexContent(name)), 0644); err != nil {
 		return err
 	}
 
 	// TODO: Remove when we can reliably fetch metadata from the store.
-	if info.Kind == "sdk" && info.Metadata == "" {
+	if meta.SdkYAML == "" {
```
dmitry-lyfar: Is it still possible to have "" here?

internal/workshop/lxd/lxd_backend_sdk.go
```diff
@@ -245,25 +228,37 @@ func (s *Backend) ImportVolume(ctx context.Context, info workshop.VolumeSetup, t
...
+	return conn.UpdateStoragePoolVolume(storagePool, "custom", name, volPut, "")
 }
 
-func (s *Backend) DetachVolume(ctx context.Context, wp, name string) error {
-	return s.RemoveWorkshopMount(ctx, wp, name)
+func sdkMetaToConfig(meta sdk.Meta) map[string]string {
+	config := map[string]string{
+		"user.kind": "sdk",
+	}
+	if meta.Sha3_384 != "" {
```
dmitry-lyfar: Is my assumption that an `sdk.Meta` with an empty hash value would not get to this point where we might need to check if it's empty? So, that when this is de-serialised from the config, there will, likewise, be no scenario with a volume does not have a hash.

To me, the lack of hash in the SDK meta == broken volume / meta data integrity that workshopd should not accomodate.

internal/overlord/sdkstate/handlers.go
```diff
@@ -125,22 +120,16 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
 		task.Set("sdk-yaml", meta.SdkYAML)
 		st.Unlock()
 	}
+	meta.SdkYAML = ""
```
dmitry-lyfar: Could you elaborate on this?

internal/overlord/sdkstate/handlers.go
```diff
@@ -161,23 +150,12 @@ func (m *SdkManager) doInstallSdk(task *state.Task, tomb *tomb.Tomb) error {
...
+	// TODO: Remove this section when we can download a specific SDK
+	// revision from the Store (this revision was validated earlier).
 	wp, err := m.backend.Workshop(ctx, w)
 	if err != nil {
 		return err
 	}
-
-	// TODO: Remove this section when we can download a specific SDK
-	// revision from the Store (this revision was validated earlier).
 	sdkYaml, err := maybeSdkYaml(task)
```
dmitry-lyfar, line 158: Same as above, are there cases when there is no `sdkYaml` but we still OK to move on here?

internal/overlord/workshopstate/handlers.go
```diff
@@ -200,7 +203,7 @@ func (m *WorkshopManager) doRemoveWorkshopStorage(task *state.Task, tomb *tomb.T
 	return errs[len(errs)-1]
```
dmitry-lyfar: Instead of logging the errors iteratively, we can:

```
return errors.Join(errs...)
```

internal/overlord/workshopstate/handlers.go
```diff
@@ -180,12 +180,15 @@ func (m *WorkshopManager) doRemoveWorkshopStorage(task *state.Task, tomb *tomb.T
 	}
 
 	var errs []error
-	if err := m.cleanupWorkshopData(user, prj.ProjectId, w); err != nil {
+	if err := m.cleanupWorkshopUserData(user, prj.ProjectId, w); err != nil {
```
dmitry-lyfar: I think that in `request.go:1280` we should create a lane per the clean up task, otherwise, if either of the three fails (e.g. `removeDirs`) the rest will be aborted. I'm commenting here as there is no change to the logic in this PR.

dmitry-lyfar (commented):
This is a very nice simplification.

internal/workshop/lxd/lxd_backend_sdk.go
jonathan-conder: Not after I remove the `meta.SdkYAML = ""` line below. Must have switched contexts after I decided to extract the YAML during the download and never came back to clean it up. Well I remember trying to clean it up at some point, and it didn't work. But I don't see any issues now. Will see if the tests pass once I do.

internal/workshop/lxd/lxd_backend_sdk.go
jonathan-conder: I think you're right, it seems *possible* but only if the user manually edits the try area.

internal/overlord/sdkstate/handlers.go
jonathan-conder: This was a shim to preserve the previous behaviour. But I don't think it's needed now that the rest of the codebase has been updated.

internal/overlord/sdkstate/handlers.go
jonathan-conder: This one is different. It's asking if there's an `sdk-yaml` attached to the `task`. That only happens very rarely, when the SDK was updated in between the Store query and download.

In the next PR I should be able to move this check into the download instead of splitting it up like this. The only issue is that the download doesn't currently know they workshop's base. Current plan is to attach the entire "plan" i.e. base image + SDKs to the `Change` instead of individual tasks.

Anyway, once we get the Store this scenario won't happen anymore.

internal/overlord/workshopstate/handlers.go
jonathan-conder: So one lane per cleanup task?

dmitry-lyfar: Yes. I would validate how it works in a test, probably, but if a task returns an error, all its lanes will be aborted. I assume we don't want to skip other clean up logic if this one has failed. Now they all belong to the same lane.

internal/workshop/lxd/lxd_backend_sdk.go
jonathan-conder: There was one outstanding issue in the tests, but I fixed it since there's no point having these workarounds in the production code just to support the tests.

internal/overlord/workshopstate/handlers.go
jonathan-conder: I tried to learn a bit more about how lanes work. Looks like a task error doesn't always cancel all tasks in the same lane. Some of those tasks might also belong to other lanes which aren't being cancelled. But in this type of case, where each task has only 1 lane, I think they will all be cancelled when any of them is. So yes we should split it into several lanes.

jonathan-conder: Another issue is what happens when a task is cancelled. As far as I can tell, since the cleanup tasks are created during a single critical section (i.e. state is not unlocked between tasks), the handlers will also be spawned in the same critical section. So by the time one task fails, it's too late to cancel the handlers.

This means there's no way to cancel `remove-state-storage` and `remove-workshop-storage`. However, `remove-workshop-stash` can be cancelled because it uses the `Tomb` to create a `context.Context`. I'm not sure we actually want that task to be cancellable though. If we don't, then we don't need to use separate lanes. I would maybe do it anyway because the behaviour may change in future.

jonathan-conder: I think I'll do this as part of the next PR. Could be wrong but I think the stash currently depends on the apt cache, so we shouldn't remove it before the stash. After decoupling them they can just run concurrently.

Another issue is with `MarkEdge`. If all the cleanup tasks are independent, we'd need a different marker for each. I suggest we only mark the `EdgeRefreshLastTaskBeforeCleanup` and drop the other mark. Then, for each pair `t1 != t2` of marked tasks we do something like `t2.HaltTasks().WaitFor(t1)`.

dmitry-lyfar: That's OK with me. Also, I think that a failed clean up logic needs to be either retried (if the error is transient) OR just logged (perhaps, even a warning). If `doRemoveWorkshopStorage` returns an error, it won't do any good to the change as it is past the time anything can be sensibly restored and it blocks other clean ups.

dmitry-lyfar: If we accept that clean up tasks do not return error as per the above, we won't need a separate lane per task.

internal/overlord/workshopstate/handlers.go
```diff
@@ -366,29 +361,41 @@ func (m *WorkshopManager) undoStashWorkshop(task *state.Task, tomb *tomb.Tomb) e
...
 func (m *WorkshopManager) doCreateStateStorage(task *state.Task, tomb *tomb.Tomb) error {
-	user, prj, w, err := UserProjectWorkshop(task)
+	username, prj, w, err := UserProjectWorkshop(task)
 	if err != nil {
 		return err
 	}
 
-	ctx, cancel := BackendContext(tomb, user, prj.ProjectId)
-	defer cancel()
+	// The workshop's root user needs write access to the state storage.
```
dmitry-lyfar, line 369: workshop's `root` or `workshop` user is meant here? I'm probably missing something.

dmitry-lyfar (approved):

internal/overlord/workshopstate/handlers.go
jonathan-conder: The save and restore hooks run as `root` inside the user namespace

PR #538, opened 2025-11-18
Remove all created SDK profiles if the change is undone

dmitry-lyfar:
# Description

If `undoSetupProfiles` does not remove all the previously created SDK profiles by `doSetupProfiles`, it results in remaining LXD profiles that were created for the new SDKs in a failed refresh change. I.e. the workshop will have an SDK profile created and assigned from an SDK that is not installed. The easiest reproducer would be a failing `setup-project` hook in a refresh change that installs new SDKs.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/overlord/ifacestate/handlers.go
+++ b/internal/overlord/ifacestate/handlers.go
@@ -858,7 +858,7 @@ func (m *InterfaceManager) doSetupProfiles(task *state.Task, tomb *tomb.Tomb) er
 
 			rev.Add(func() {
 				if err1 := backend.Remove(ctx, ref); err1 != nil {
-					logger.Noticef(`On doSetupProfiles: Failed to clean up %q SDK backend setup`, ref.ShortRef())
+					logger.Noticef(`On doSetupProfiles: Failed to clean up %q SDK backend setup (%v)`, ref.ShortRef(), err1)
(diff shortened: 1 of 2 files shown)


jonathan-conder (approved):
I think the change makes sense. I don't understand the original reason why `undoSetupProfiles` sometimes called `backend.Setup` though

internal/overlord/ifacestate/handlers.go
```diff
@@ -869,20 +869,11 @@ func (m *InterfaceManager) doSetupProfiles(task *state.Task, tomb *tomb.Tomb) er
...
 	st := task.State()
-	user, p, w, err := handlersetup.UserProjectWorkshop(task)
-	if err != nil {
-		return err
-	}
-
-	st.Lock()
-	s, err := handlersetup.Sdk(task)
-	st.Unlock()
+	user, _, _, err := handlersetup.UserProjectWorkshop(task)
```
jonathan-conder: apparently there's `handlersetup.User(task.Change())` for this purpose

PR #540, opened 2025-11-18
Refactor SDK metadata

jonathan-conder:
# Description

My aim was to introduce a struct which captures the ingredients that go into a workshop snapshot. This will be used in the `workshop.Backend` API to create, lookup and restore snapshots. Ultimately it should look something like:
```go
type Layers struct {
    Image BaseImage
    Sdks  []Sdk
}

type BaseImage {
    Name        string
    Fingerprint string
}

type Sdk {
    Name     string
    Source   string
    Revision Revision
    Hash     string
}
```

Both `Name` and `Revision` aren't strictly necessary, but make debugging easier. Also `Source` can be more coarse, e.g. there's no need to distinguish between `try` and Store SDKs. I plan to include `Name`s in the layer hash, to reduce the blast radius of hash collisions.

This PR introduces a `BaseImage` struct as above, and adjusts `sdk.Setup` to be closer to the `Sdk` struct. `sdk.Setup` now includes hashes and doesn't include the install time. It won't be easy to remove `Channel` until we migrate to the actual Store.

I also took the opportunity to rename `sdk.SdkResult` to `sdk.Meta` and consolidate some similar types used in tests. It's now just `sdk.Setup` together with the SDK YAML file.

This is likely to break existing workshops due to changes to `sdk.Setup`.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/internal/daemon/api_workshops.go
+++ b/internal/daemon/api_workshops.go
@@ -51,7 +51,7 @@ type SdkInfo struct {
 	Source      string           `json:"source,omitempty"`
 	Revision    string           `json:"revision"`
 	BuildTime   *time.Time       `json:"build-time,omitempty"`
-	InstallTime *time.Time       `json:"install-time,omitempty"`
+	InstallTime time.Time        `json:"install-time,omitempty"`
 	Health      *HealthCheckInfo `json:"health-check,omitempty"`
 	Mounts      []*Mount         `json:"mounts,omitempty"`
 	Tunnels     []*Tunnel        `json:"tunnels,omitempty"`
--- a/internal/fakestore/store.go
+++ b/internal/fakestore/store.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/md5"
+	"crypto/sha3"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -51,7 +52,7 @@ type storeSdk struct {
 	Channel  string       `json:"channel"`
 	Revision sdk.Revision `json:"revision"`
 	SdkYAML  string       `json:"sdk-yaml"`
-	MD5      string       `json:"md5"`
+	Sha3_384 string       `json:"sha3-384"`
 }
 
 type sdkReader struct {
@@ -94,7 +95,7 @@ func (c *GcsStore) SdkAction(ctx context.Context, actions []sdk.SdkAction) ([]sd
 			}
 
 			setup := sdk.Setup{Name: s.Name, Channel: s.Channel, Revision: s.Revision}
-			results = append(results, sdk.SdkResult{Setup: setup, MD5: s.MD5, SdkYAML: s.SdkYAML})
+			results = append(results, sdk.SdkResult{Setup: setup, Sha3_384: s.Sha3_384, SdkYAML: s.SdkYAML})
 		default:
 			return nil, fmt.Errorf("unknown SDK store action")
 		}
@@ -146,7 +147,7 @@ func (c *GcsStore) DownloadSdk(ctx context.Context, setup sdk.Setup, report *pro
 			return nil, err
 		}
 
-		return &sdk.SdkResult{Setup: setup, MD5: digest, SdkYAML: sdkYaml}, nil
+		return &sdk.SdkResult{Setup: setup, Sha3_384: digest, SdkYAML: sdkYaml}, nil
 	}
 
 	r, err := storeSdkReader(ctx, setup)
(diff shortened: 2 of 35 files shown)


internal/daemon/api_workshops.go
```diff
@@ -51,7 +51,7 @@ type SdkInfo struct {
 	Source      string           `json:"source,omitempty"`
 	Revision    string           `json:"revision"`
 	BuildTime   *time.Time       `json:"build-time,omitempty"`
-	InstallTime *time.Time       `json:"install-time,omitempty"`
+	InstallTime time.Time        `json:"install-time,omitempty"`
```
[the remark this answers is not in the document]
jonathan-conder: It's never zero in practice

internal/overlord/sdkstate/handlers.go
```diff
@@ -229,16 +228,7 @@ func (m *SdkManager) mountSdk(ctx context.Context, user string, project *worksho
...
-	digest, err := os.Readlink(revision)
-	if err != nil {
-		return err
-	}
-	// Ensure we only mount actual SDKs and avoid malicious symlinks.
-	if strings.ContainsRune(digest, os.PathSeparator) {
-		return &os.PathError{Op: "mount", Path: revision, Err: fmt.Errorf("invalid hash %q", digest)}
-	}
-	what := filepath.Join(sdkDir, digest)
+	what := filepath.Join(sdkDir, sdkSetup.Sha3_384)
```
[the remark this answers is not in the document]
jonathan-conder: This is sourced from `sdk.CommitRevision` and is guaranteed to be a valid hash, instead of reading it from a possibly modified link

internal/fakestore/store.go
dmitry-lyfar, line 134: Hmm, it looks like we use the revision to find out if the SDK is cached or not. We do calculate and store md5 for the SDKs, but I fail to find how `MD5` hashes are then used. E.g. we do not compare hashes obtained from the store vs. the ones existing in `target.md5` files.

internal/overlord/workshopstate/handlers.go
```diff
@@ -34,12 +33,11 @@ func (m *WorkshopManager) doDownloadBase(task *state.Task, tomb *tomb.Tomb) erro
...
-	var base, fingerprint string
+	var image workshop.BaseImage
 	st.Lock()
-	err1 := task.Get("workshop-base", &base)
-	err2 := task.Get("workshop-base-fingerprint", &fingerprint)
+	err = task.Get("workshop-base", &image)
 	st.Unlock()
-	if cmp.Or(err1, err2) != nil {
+	if err != nil {
 		return fmt.Errorf("internal error: %q workshop configuration not found (task ID: %s)", w, task.ID())
```
dmitry-lyfar: Shall we adjust the error message, e.g. workshop base image configuration not found (or just wrap the err into the return)?

internal/overlord/workshopstate/request.go
```diff
@@ -571,8 +584,8 @@ func rebuildWorkshop(st *state.State, name string, fileText string, sdkSnapshot
 	if sdkSnapshot != "" {
 		create.Set("sdk-snapshot", sdkSnapshot)
 	}
-	if fingerprint != "" {
-		create.Set("workshop-base-fingerprint", fingerprint)
+	if image.Fingerprint != "" {
```
dmitry-lyfar: Should we continue here at all? If there is no fingerprint, it would fail with internal error?

internal/overlord/workshopstate/request.go
```diff
@@ -683,13 +696,13 @@ func (w *WorkshopManager) RefreshMany(ctx context.Context, projectId string, nam
 			if err != nil {
 				return nil, fmt.Errorf("cannot refresh %q: %w", name, err)
 			}
-			setups, err := validateSdkResults(projectId, req.file, localSdks)
+			setups, err := validateSdkMeta(projectId, req.file, localSdks)
```
dmitry-lyfar:
```suggestion
			localSetups, err := validateSdkMeta(projectId, req.file, localSdks)
```

dmitry-lyfar (approved):

internal/overlord/workshopstate/handlers.go
jonathan-conder: Might as well

internal/overlord/workshopstate/request.go
jonathan-conder: The intent was to only add the attribute if it will be used, but at some point the fingerprint became nonempty here. Updated to reflect original intent

internal/fakestore/store.go
jonathan-conder: As discussed I've now gone back and confined the MD5 hashes to the Store implementation details. It converts MD5 to SHA3 by hashing the digest, with a prefix to hopefully avoid collisions with actual tarballs.

I think you're right that nothing actually uses the hashes in `sdk.Setup` yet. The plan is to use them during launch and refresh planning to pick a snapshot to restore from. This happens before the SDKs are downloaded so we can't use `sdk.Meta`. I also want the hashes to be stored in the workshop (and thus, snapshot) metadata because it seems like a bad idea to name snapshots after some opaque string that can't be reconstructed from the metadata.

PR #541, opened 2025-11-18
Update SDKcraft tutorial

jonathan-conder:
# Description

I went through this to see if I could reproduce an issue. No luck on that but I noticed a few things that could be improved:
- No need to emphasise certain parts of the `sdk.yaml` when all lines have changed.
- The `ollama.service` file is outdated.
- The note about `apt` and `sudo -u` is irrelevant here. Maybe there's somewhere else we could move it though.
- The `check-health` hook doesn't work as intended. See https://github.com/canonical/sdks/pull/76.
- The `workshop.yaml` for trying out Ollama was abbreviated before showing the full file later on. It's short so we can just show the full thing twice.
- The note about unpublished SDKs is outdated since we have `sdkcraft try` now.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Or:

### DIFF
--- a/docs/tutorial/part-4-craft-sdks.rst
+++ b/docs/tutorial/part-4-craft-sdks.rst
@@ -556,12 +556,6 @@ The resulting SDK can be accessed by |ws_markup| as follows:
 Note that the workshop :samp:`base`
 must match one of the SDK's supported :samp:`platforms`.
 
-.. note::
-
-   Currently, you can't use unpublished SDKs in a workshop.
-   However, the :ref:`sketch SDK <tut_sketch_sdks>` and in-project SDKs provide
-   a subset of |sdk_markup|'s functionality.
-
 
 Summary
 -------
--- a/docs/tutorial/part-4-craft-sdks.rst
+++ b/docs/tutorial/part-4-craft-sdks.rst
@@ -198,7 +198,7 @@ create a file named :file:`ollama.service`:
 
    [Unit]
    Description=Ollama Service
-   After=network-online.target
+   After=network.target
 
    [Service]
    ExecStart=/bin/bash -lc "ollama serve"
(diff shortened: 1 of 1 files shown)


docs/tutorial/part-4-craft-sdks.rst
```diff
@@ -327,21 +327,8 @@ named :file:`setup-base`:
...
-   For workshops,
-   :command:`apt` is configured to exclude recommended or suggested packages
-   and answer 'yes' to all confirmation prompts.
-
-   Also, the use of :command:`sudo -u workshop` here is important
-   because only the :samp:`setup-project` hook runs as a normal user by default;
-   other hooks, like :samp:`setup-base`, run as root.
-   Running commands as the non-root user
-   helps preserve the correct environment variables and file ownership,
-   and can be easier than adjusting permissions afterwards.
```
akcano, line 343: This part of the note would look better after the `check-health` hook that actually uses this mechanism.

docs/tutorial/part-4-craft-sdks.rst
```diff
@@ -327,21 +327,8 @@ named :file:`setup-base`:
 
 
 It runs when the workshop is launched or refreshed,
-installing system packages and preparing the workshop for use.
-
-.. note::
-
-   For workshops,
-   :command:`apt` is configured to exclude recommended or suggested packages
-   and answer 'yes' to all confirmation prompts.
```
akcano, line 336: I suggest keeping this here for now, because it's an important piece of info that we don't want to lose; I'll make a note to find a better place, though.

akcano (approved):
LGTM, a couple of comments.

PR #545, opened 2025-11-20
Doc: Add warning, factor starter pack SDK out into a workflow

akcano:


.github/workflows/update-starter-pack.yaml
```diff
@@ -0,0 +1,62 @@
...
+  schedule:
+    - cron: '0 0 * * 1'  # Weekly on Mondays
+
+jobs:
+  update-starter-pack:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
```
jonathan-conder:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/update-starter-pack.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Setup Python
+        uses: actions/setup-python@v4
```
jonathan-conder:
```suggestion
        uses: actions/setup-python@v6
```

.github/workflows/update-starter-pack.yaml
```diff
@@ -0,0 +1,62 @@
...
+          python-version: "3.x"
+
+      - name: Update starter pack files
+        run: |
+          # Create a temporary directory for the starter pack clone
+          TMP_DIR=$(mktemp -d)
+          trap 'rm -rf "$TMP_DIR"' EXIT
+          
+          # Clone the latest starter pack
+          git clone --depth 1 https://github.com/canonical/sphinx-docs-starter-pack "$TMP_DIR"
```
jonathan-conder: I think `actions/checkout` is more efficient than a shallow clone. Not sure if it will let you checkout to a temp dir though

.github/workflows/update-starter-pack.yaml
```diff
@@ -0,0 +1,62 @@
...
+          # Ensure the docs directory exists
+          mkdir -p docs
+          
+          # Copy the Makefile and .sphinx directory to the project's docs directory
+          cp -ar "$TMP_DIR/docs/Makefile" "$TMP_DIR/docs/.sphinx" "$TMP_DIR/docs/requirements.txt" docs/
+          
+          echo "Starter pack files updated successfully"
+
+      - name: Install documentation dependencies
+        run: make -C docs install
```
jonathan-conder, line 45: Not sure what this achieves. Maybe it should build the docs, as a sanity check? But we should probably just make the PR and make sure building the docs is part of our PR checks if it isn't already.

.workshop/tools/sdk.yaml
```diff
@@ -3,3 +3,6 @@ slots:
   fake-store:
     interface: tunnel
     endpoint: 8080
+  sphinx-build:
+    interface: tunnel
+    endpoint: 8000
```
jonathan-conder: missing a trailing newline here

jonathan-conder (approved):
I think this is a nice approach. Some minor nitpicks

.github/workflows/update-starter-pack.yaml
akcano: Turns out it does allow that.

.github/workflows/update-starter-pack.yaml
akcano: In my view, the update action updates the existing starter pack files in-place. To achieve this, the minimal step is to copy the relevant files and run this target because the state of the files as committed in the repo is what we have after 'make install'; they're not ephemeral normally. However, for a general SP SDK, this may present a viable approach.

I've reworked this logic a bit in a follow-up patch (5afd4e8); the `make html` build gate is definitely an improvement, thanks.

jonathan-conder: Ah I see, didn't realise `install` changed non-ephemeral files. TBH I don't think the `make html` was a good idea. We should make the PR so it's visible that there's a new version. If the PR fails to build then we can fix it before merging

akcano: I just realized there'll be an RTD build for the PR anyway, so it's indeed superfluous. Removing.

dmitry-lyfar (approved):

PR #557, opened 2025-11-25
Doc: update sdk channels to 24.04 and workshop base to ubuntu@24.04

locnnil:
# Description

These changes ensure that all tutorials consistently uses the Noble variant.

When following the second tutorial, I've found an inconsistency with the first one: the ollama sdk channel is set as `22.04/edge` and the workshop base is set as `ubuntu@22.04`. While on the first one, channel and base are set to `24.04`.

So, this gives this mismatched base x channel error when trying to add Jupyter and refresh:

```
error: cannot refresh "dev": "jupyter" SDK has "ubuntu@22.04" base; required: "ubuntu@24.04"
"dev" refresh aborted
```

By having this workshop.yaml file:

```yaml
name: dev
base: ubuntu@24.04
sdks:
  - name: ollama
    channel: 24.04/edge
  - name: jupyter
    channel: 22.04/edge
```

### DIFF
--- a/docs/tutorial/part-1-get-started.rst
+++ b/docs/tutorial/part-1-get-started.rst
@@ -156,15 +156,15 @@ create a workshop definition named :file:`workshop.yaml`:
    :emphasize-lines: 4,5
 
    name: dev
-   base: ubuntu@24.04
+   base: ubuntu@22.04
    sdks:
      - name: ollama
-       channel: 24.04/edge
+       channel: 22.04/edge
 
 
 Here, the SDK is referenced as :samp:`ollama`,
 and the specific version to retrieve from the SDK Store
-comes from the :samp:`24.04/edge` channel.
+comes from the :samp:`22.04/edge` channel.
 
 To confirm that |ws_markup| sees the definition,
 list the workshops in the project directory:
@@ -228,7 +228,7 @@ to see what went into your workshop:
    $ workshop info
 
      name:     dev
-     base:     ubuntu@24.04
+     base:     ubuntu@22.04
      project:  /home/user/ollama-python-project
      status:   ready
      notes:    -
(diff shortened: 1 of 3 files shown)


akcano:
Thanks for your consideration!

However, there's certain logic in this. Here in Part 1, we switch the workshop to 22.04:

https://canonical-workshop.readthedocs-hosted.com/latest/tutorial/part-1-get-started/#refresh-a-workshop

So to maintain continuity, we start Part 2 with the same workshop, hence 22.04.

However, we can definitely improve this by making the change more logical (24->22, not vice versa) and emphasizing it a bit. Would you be interested in updating this PR accordingly, or should I do this instead?

locnnil:
@akcano, Oh, I didn't notice before that the base and channel were switched, thanks for pointing this.

I think, to bring this closer to a real-world scenario, it would be good to first present the workshop definition using the 22 version, then switching it to 24. I've made that change on: f4751013 please let me know what is your take on this.

dmitry-lyfar:
@akcano do you want to update tests here or as a separate PR?

akcano:
@dmitry-lyfar I'll update the tests separately, need to wiggle them a bit anyway.

docs/tutorial/part-1-get-started.rst
```diff
@@ -323,6 +323,9 @@ are updated by their publishers.
 Alternatively,
 you may have changed the definition to switch bases,
 add and remove SDKs, or toggle their channels.
+A good example is:
+when a new Ubuntu LTS is available and by consequence,
+a new base image is published;
```
akcano:
```suggestion
A good example is when a new Ubuntu LTS version is released and,
as a result,
a new base image becomes available.
```

akcano (approved):
LGTM wit one minor fix.

PR #563, opened 2025-12-01
Drop config and devices from SDK snapshots

jonathan-conder:
# Description

- Renames SDK disk devices to `sdk.<NAME>`.
- Stores SDK metadata as `user.sdk.*` properties of these devices, removing the need for `user.workshop.sdks`.
- Removes most config options from workshop snapshots. The devices are replaced by `{"type": "none"}`. SDK devices retain a minimal amount of metadata.

This makes the snapshots behave more like base images, so I was able to merge `Restore` into `LaunchOrRebuildWorkshop` and share a little bit of code between them.

It is likely to break existing workshops.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/workshop/lxd/lxd_backend_layers.go
+++ b/internal/workshop/lxd/lxd_backend_layers.go
@@ -13,12 +13,45 @@ import (
 
 	lxd "github.com/canonical/lxd/client"
 	"github.com/canonical/lxd/shared/api"
-	"gopkg.in/yaml.v3"
 
 	"github.com/canonical/workshop/internal/logger"
 	"github.com/canonical/workshop/internal/revert"
+	"github.com/canonical/workshop/internal/sdk"
 	"github.com/canonical/workshop/internal/workshop"
 )
+
+// Classifies LXD config options into different domains which describe who is
+// responsible for them and how they (should) behave when copying an instance.
+type configDomain int
+
+const (
+	// Properties unique to a single instance. These are managed by LXD and
+	// not shared with instance copies, e.g. volatile.uuid.
+	uniqueProperty configDomain = iota
+	// Properties managed by LXD and shared with copies, e.g. image.os.
+	sharedProperty
+	// Options managed by Workshop. Includes instance config options like
+	// boot.autostart and workshop metadata like user.workshop.project-id.
+	customOption
+)
+
+// Classifies options by key. Based on LXD's InstanceIncludeWhenCopying.
+func optionDomain(key string) configDomain {
+	if strings.HasPrefix(key, "image.") {
+		return sharedProperty
+	}
+
+	suffix, found := strings.CutPrefix(key, "volatile.")
+	if !found {
+		return customOption
+	}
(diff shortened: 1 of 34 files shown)


jonathan-conder:
Drafting because the snapshots do not currently have a notion of which order the SDKs were installed. One way is to mark them with `user.sdk.index` (0, 1, 2, 3, etc.). But it's something I need to think about a bit more. Maybe `workshop.Workshop.Sdks` should become a list?

internal/workshop/lxd/lxd_backend_layers.go
```diff
@@ -241,43 +406,35 @@ func (s *Backend) layerNames(layerConn lxd.InstanceServer, pid, w string, kind s
...
+		})
+		if idx < 0 {
 			obstacles = append(obstacles, layer.Name)
+		} else {
+			names[idx] = layer.Name
 		}
 	}
 
-	return layers[idx].Name, obstacles, nil
+	if idx := slices.Index(names, ""); idx >= 0 {
```
dmitry-lyfar: No sure I understand why `names` can contain an empty string.

internal/workshop/lxd/lxd_backend_layers.go
```diff
@@ -241,43 +406,35 @@ func (s *Backend) layerNames(layerConn lxd.InstanceServer, pid, w string, kind s
 	return names, nil
 }
 
-// Find layer name for the given SDK and any subsequent SDKs.
-func (s *Backend) layerNamesAfter(layerConn lxd.InstanceServer, pid, w, sk string) (string, []string, error) {
+// Find layer name for the given SDKs and any subsequent SDKs.
+func (s *Backend) layerNamesSplit(layerConn lxd.InstanceServer, pid, w string, sdks []sdk.Id) ([]string, []string, error) {
```
dmitry-lyfar:
```suggestion
func (s *Backend) layerToRestore(layerConn lxd.InstanceServer, pid, w string, sdks []sdk.Id) (string, []string, error) {
```
Can it just return the layer's name as the first return value instead of making the caller to take the last value from the list?

internal/workshop/lxd/lxd_backend_layers.go
jonathan-conder: `names` starts as a fixed side slice of empty strings. We go through the snapshots and fill it in with their names, in some unknown order. If anything is empty at the end, it means we didn't find a layer for that SDK.

internal/workshop/lxd/lxd_backend_layers.go
jonathan-conder: Maybe I should just go back to how it was before. This will go away once the layers are hashed anyway.

dmitry-lyfar (approved):

PR #566, opened 2025-12-02
Add go proxy EOF workaround for GitHub workflows

dmitry-lyfar:


jonathan-conder (approved):
Might as well try and see how it goes

PR #570, opened 2025-12-04
Tune apt-get to be more resilient to network issues in spread tests

dmitry-lyfar:


tests/lib/utils.sh
```diff
@@ -26,6 +26,16 @@ function prepare_environment() {
     systemctl disable apt-daily.timer || true
     systemctl disable apt-daily-upgrade.timer || true
 
+    # Configure apt to retry operations to handle flaky network issues
+    # This helps with mirror sync errors and transient network failures
+    cat <<EOF >/etc/apt/apt.conf.d/80-retries
+Acquire::Retries "10";
+Acquire::http::Timeout "30";
+Acquire::https::Timeout "30";
+Acquire::http::Pipeline-Depth "0";
```
jonathan-conder: According to `apt-transport-https(1)` the `https` values are inherited from `http`:
> all options supported by `apt-transport-http(1)` are also available via `Acquire::https` and will default to the same values specified for `Acquire::http`

It also supports `Pipeline-Depth`, so I reckon we either omit `https` or add `Acquire::https::Pipeline-Depth` if we want to be explicit

jonathan-conder (approved):

PR #573, opened 2025-12-05
Handle ambiguous workshop exec and run arguments

jonathan-conder:
# Description

Using a separator (--) to split the command or action from workshop CLI arguments leaves no room for ambiguity. But users seem to prefer leaving it out, for interactive use. So we now parse arguments like this:

- `workshop exec dev -- bash`: execute "bash" in "dev"
- `workshop exec -- bash`: execute "bash" in the single workshop
- `workshop exec dev bash`: try the first case, but fall back to the second if:
  - the project has a single workshop
  - it's not called `dev`, and
  - stdout is a TTY.

The workshop CLI now ignores `--` if it's not in position 0 or 1. This is mostly to simplify the code, but also these are the only places where it has a special meaning. For reference, `sudo` only considers `--` if it's before the command.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

* [x] If needed, `docs/.coverage.yaml` updated, coverage tags added (`.. artefact`).

---

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/canonical/x-go/strutil/shlex"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"golang.org/x/sys/unix"
@@ -62,10 +63,9 @@ type ExecFlags struct {
 }
 
 type ExecArgs struct {
-	workshop string
-	implicit bool
-	command  []string
-	action   bool
+	av            []string
+	argsLenAtDash int
+	action        bool
 }
 
 var shortExecHelp = "Run a command and wait for it to complete"
@@ -90,8 +90,13 @@ To set the mode explicitly, use "-i" or "-I". If neither is supplied,
 
 
 To separate the "exec" subcommand from the command itself,
-use shell syntax such as *--*.
-This syntax is required if the workshop name is omitted.
+use a separator (*--*).
+
+If you omit the separator,
+"exec" treats its first argument as the workshop name.
+If the project has no such workshop
+and the shell is interactive,
+the argument is treated as a command to run in the default workshop.
 
 Notes:
 
@@ -140,9 +145,13 @@ To set the mode explicitly, use "-i" or "-I". If neither is supplied,
 
 
 To separate the "run" subcommand from the action and its arguments,
-use shell syntax such as *--*.
-This syntax is required if the workshop name is omitted
-and the action takes one or more arguments.
+use a separator (*--*).
+
+If you omit the separator,
+"run" treats its first argument as the workshop name.
+If the project has no such workshop
+and the shell is interactive,
+the argument is treated as an action to run in the default workshop.
 
 Notes:
 
@@ -160,26 +169,30 @@ This command enumerates all actions in the workshop, printing a YAML map.
 func (c *CmdExec) Command() *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "exec [flags] [<WORKSHOP>] [--] <COMMAND>...",
-		Args:  maybeNameAndCommand,
+		Args:  cobra.MinimumNArgs(1),
 		Short: shortExecHelp,
 		Long:  longExecHelp,
 		Example: `
 Run the "go build main.go" command under the "nimble" workshop
 in the current project directory:
-$ workshop exec nimble go build main.go
+$ workshop exec nimble -- go build main.go
 
 A similar command that sets an environment variable and the working directory:
-$ workshop exec --env GO111MODULE=off -w /project nimble go build -x
+$ workshop exec --env GO111MODULE=off -w /project nimble -- go build -x
+
+Run a command as root (the default is "workshop"):
+$ workshop exec --uid 0 nimble id
 
 Run a custom interactive shell:
 $ workshop exec -I nimble sh
 
-The name is optional if the project has only one workshop
-and a separator is provided:
-$ workshop exec -I -- sh
+If the project has only one workshop, the workshop name is optional:
+$ workshop exec -- sh
 
-Run a command as root (the default is "workshop"):
-$ workshop exec --uid 0 nimble id`,
+If the command doesn't overlap with a workshop name
+and the shell is interactive,
+the separator is also optional:
+$ workshop exec sh`,
 		RunE:              c.Run,
 		ValidArgsFunction: c.root.completeWorkshopName([]string{"Ready", "Waiting"}),
 	}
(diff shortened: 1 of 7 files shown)


docs/how-to/customize-workshops/add-actions.rst
```diff
@@ -73,7 +75,33 @@ use the :command:`workshop run` command followed by the action name:
 
 When you run an action using :command:`workshop run`,
 any additional arguments provided after the action name
-are passed directly to the action itself.
+are passed directly to the action itself:
+
+.. code-block:: console
+
+   $ workshop run dev -- lint --verbose
+
```
akcano, line 83: blank line not needed

docs/how-to/customize-workshops/add-actions.rst
```diff
@@ -73,7 +75,33 @@ use the :command:`workshop run` command followed by the action name:
...
+.. warning::
+
+   If the workshop is named :samp:`shellcheck`,
+   the above command will try to run its :samp:`scripts` action,
+   rather than running the :samp:`shellcheck` action
+   with :samp:`scripts` as an argument.
+   Omitting the separator (:samp:`--`) and the workshop name
+   can be convenient for interactive use,
+   but should be avoided in non-interactive scripts.
```
akcano:
```suggestion
```

akcano: This is definitely not how-to material.

docs/how-to/customize-workshops/add-actions.rst
```diff
@@ -73,7 +75,33 @@ use the :command:`workshop run` command followed by the action name:
...
+
+   $ workshop run dev shellcheck scripts/
+
+
+In projects with a single workshop, the workshop name is optional:
+
+.. code-block:: console
+
+   $ workshop run -- lint
+
```
akcano, line 90: blank line not needed

cmd/workshop/exec.go
```diff
@@ -89,10 +88,6 @@ To set the mode explicitly, use '-i' or '-I'. If neither is supplied,
 - Otherwise, it's non-interactive
 
 
-To separate the 'exec' subcommand from the command itself,
-use shell syntax such as *--*.
-This syntax is required if the workshop name is omitted.
```
akcano:
```suggestion
To separate "exec" from the command and its arguments,
use shell syntax such as *--*.

If you omit the separator,
"exec" tries its first argument as the workshop name.
If the project has no such workshop,
the argument is treated as the command to run in the default workshop.
```

cmd/workshop/exec.go
```diff
@@ -139,11 +134,6 @@ To set the mode explicitly, use '-i' or '-I'. If neither is supplied,
 - Otherwise, it's non-interactive
 
 
-To separate the 'run' subcommand from the action and its arguments,
-use shell syntax such as *--*.
-This syntax is required if the workshop name is omitted
-and the action takes one or more arguments.
-
```
akcano:
```suggestion
To separate "run" from the action and its arguments,
use shell syntax such as *--*.

If you omit the separator,
"run" tries its first argument as the workshop name.
If the project has no such workshop,
the argument is treated as the action to run in the default workshop.
```

docs/how-to/customize-workshops/add-actions.rst
```diff
@@ -73,7 +75,33 @@ use the :command:`workshop run` command followed by the action name:
...
+   $ workshop run shellcheck scripts
```
akcano:
```suggestion
   $ workshop run shellcheck scripts/
```

Consistency with the above example.

docs/how-to/customize-workshops/add-actions.rst
```diff
@@ -52,19 +52,21 @@ Running actions
 ---------------
 
 To execute an action,
-use the :command:`workshop run` command followed by the action name:
+use the :command:`workshop run` command
+followed by the workshop name and the action name,
```
akcano:
```suggestion
and specify the workshop and its action,
```

cmd/workshop/exec.go
akcano: Any details of a command's logic should go here first.

Also, note the double quotes.

cmd/workshop/exec.go
akcano: Any details of a command's logic should go here first.

Also, note the double quotes.

cmd/workshop/exec.go
```diff
@@ -160,26 +150,28 @@ This command enumerates all actions in the workshop, printing a YAML map.
...
 Run the 'go build main.go' command under the 'nimble' workshop
 in the current project directory:
-$ workshop exec nimble go build main.go
+$ workshop exec nimble -- go build main.go
 
 A similar command that sets an environment variable and the working directory:
-$ workshop exec --env GO111MODULE=off -w /project nimble go build -x
+$ workshop exec --env GO111MODULE=off -w /project nimble -- go build -x
+
+Run a command as root (the default is 'workshop'):
```
akcano:
```suggestion
Run a command as root (the default is "workshop"):
```

cmd/workshop/exec.go
```diff
@@ -160,26 +150,28 @@ This command enumerates all actions in the workshop, printing a YAML map.
...
 
-The name is optional if the project has only one workshop
-and a separator is provided:
-$ workshop exec -I -- sh
+The workshop name is optional if the project has only one workshop:
+$ workshop exec -- sh
 
-Run a command as root (the default is 'workshop'):
-$ workshop exec --uid 0 nimble id`,
+The separator is also optional if the command is not a workshop name:
```
akcano:
```suggestion
If the command doesn't overlap with a workshop name, the separator is also optional:
```

akcano: Conditional always goes first to simplify decision making.

cmd/workshop/exec.go
```diff
@@ -280,15 +249,13 @@ in the current project directory:
...
 
 The workshop name is optional if the project only has one workshop:
-$ workshop run build
+$ workshop run -- build
 
-Actions can accept arguments,
-if a separator or a workshop name is provided:
-$ workshop run -- build --debug
-`,
+The separator is also optional if the action is not a workshop name:
```
akcano:
```suggestion
If the action doesn't overlap with a workshop name, the separator is also optional:
```

cmd/workshop/exec.go
```diff
@@ -280,15 +249,13 @@ in the current project directory:
 $ workshop run nimble build
 
 A similar command that sets an environment variable and the working directory:
-$ workshop run --env GO111MODULE=off -w /project nimble build
+$ workshop run --env GO111MODULE=off -w /project nimble -- build
 
 The workshop name is optional if the project only has one workshop:
```
akcano:
```suggestion
If the project only has one workshop, the workshop name is optional:
```

cmd/workshop/exec.go
```diff
@@ -160,26 +150,28 @@ This command enumerates all actions in the workshop, printing a YAML map.
...
+Run a command as root (the default is 'workshop'):
+$ workshop exec --uid 0 nimble id
 
 Run a custom interactive shell:
 $ workshop exec -I nimble sh
 
-The name is optional if the project has only one workshop
-and a separator is provided:
-$ workshop exec -I -- sh
+The workshop name is optional if the project has only one workshop:
```
akcano:
```suggestion
If the project only has one workshop, the workshop name is optional:
```

akcano (changes requested):
On top of the inline comments, a few general considerations.

While I appreciate the simplicity of the new syntax, having struggled with long kebab commands myself, I am super duper concerned about the ambiguity in non-interactive scenarios.

Suppose you're reading a script or a log. The interpretation will literally depend on the layout of workshops present during a run, which requires the reader to maintain or research this for context. 

My suggestions:

- Make the decision making logic explicit in interactive mode, printing messages like 'assuming workshop %q' etc.
- Block the relaxed syntax in non-interactive mode
- Last but not least, warn about overlaps within a project at definition parsing, e.g. 'possible ambiguity: action X in workshop Y overlaps with workshop X'

I think we may need Victoria's input as well.

jonathan-conder:
> While I appreciate the simplicity of the new syntax, having struggled with long kebab commands myself, I am super duper concerned about the ambiguity in non-interactive scenarios.

Agreed, after discussing with Dmitry this morning we still want to try allowing ambiguity in interactive mode. The reason is that for users coming from `workshop launch` or `workshop info` without a workshop name, requiring one (or a `--`) for `workshop exec` and `workshop run` is unexpected.

For the same reason, it would be inconsistent to print `assuming workshop %q` for only `exec` and `run`. I think the new error messages, or the command output, should make it clear what went wrong if something unexpected happens in interactive mode.

In scripts, the output might not be checked as often, and we should optimise for reading rather than writing commands. There's also the issue that we infer what the command is likely to mean from the context. In scripts, the context can change over time as users add and remove workshops from the project. So what we decided is:
- In single-workshop projects, require either the workshop name or `--` in non-interactive mode. If neither a provided, print an error which provides an example command using `--`.
- In multi-workshop projects, the workshop name is already required. No change in behaviour based on interactive mode.

If you add a second workshop, commands which include the workshop name behave the same. Commands which use `--` to omit the workshop name will fail.

If you remove the second-to-last workshop, commands which target the remaining workshop behave the same. Commands targeting a removed workshop will fail, even without `--`.

jonathan-conder:
> * Last but not least, warn about overlaps within a project at definition parsing, e.g. 'possible ambiguity: action X in workshop Y overlaps with workshop X'

I'm not sure we need to warn here, at least I couldn't come up with a scenario which isn't covered by the changes to non-interactive mode. In any case, I don't think it's feasible to do something similar for `workshop exec` and I wouldn't add this for `run` only.

docs/how-to/customize-workshops/add-actions.rst
jonathan-conder: Good point, I wasn't thinking about that at all. I'll remove it completely since the new error message should cover this. And the reference docs will include it anyway.

I think we should try to stick to the unambiguous syntax in the docs going forward. The single workshop syntax should be in the tutorial but maybe not in how-to guides.

docs/how-to/customize-workshops/add-actions.rst
jonathan-conder: There's not much need for two commands anyway, removing.

cmd/workshop/exec.go
jonathan-conder: Great, I struggled to word this but yours is much better than anything I came up with. I made a few tweaks though.

This was based on an old version with single quotes, I'll rebase.

docs/how-to/customize-workshops/add-actions.rst
jonathan-conder: I felt this was a bit long so broke into 2 sentences

cmd/workshop/exec.go
jonathan-conder: I added "if interactive" as well. It's a bit long, would like to get your thoughts on the multi-line splitting. It will look weird in the `--help` output. Maybe we just drop the example?

akcano: looks good!

akcano: looks good!

akcano:
OK, I think I have all my concerns addressed one way or another.

dmitry-lyfar (approved):

PR #576, opened 2025-12-08
Doc: bad example

akcano:


akcano:
@copilot are you sure you're done with the review? redo it from scratch

PR #582, opened 2025-12-09
Doc: Add GH action to update SDK schema

akcano:
Used existing approach from `doc-cover.yaml` to introduce a counterpart to #580 that relies on an existing `sdkcraft` mechanism to generate the schema JSON.

### DIFF
--- a/.github/workflows/doc-update-sdk-schema.yml
+++ b/.github/workflows/doc-update-sdk-schema.yml
@@ -45,10 +45,10 @@ jobs:
             git config --global user.email 'github-actions[bot]@users.noreply.github.com'
             
             git add .
-            git commit -m "Update SDK schema"
+            git commit -m "Doc: Update SDK schema"
             git push --set-upstream origin $pr_branch
 
-            gh pr create --title "Update SDK schema" --body "This PR updates the SDK schema." --base main
+            gh pr create --title "Doc: Update SDK schema" --body "This PR updates the SDK schema." --base main
           fi
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/doc-update-sdk-schema.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ./sdkcraft
+
+      - name: Generate Schema
+        run: |
+          cd sdkcraft
+          PYTHONPATH=. python sdkcraft/models/project.py
+          mv schema.json ../docs/reference/definition-files/schema-sdk.json
```
jonathan-conder: We should use `uv` here. At first it was just a style preference but I realised recently that the schema can actually change when craft-application is updated. We should stick to using the version in `uv.lock`.

akcano: thanks!

.github/workflows/doc-update-sdk-schema.yml
```diff
@@ -0,0 +1,54 @@
...
+            git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+            
+            git add .
+            git commit -m "Doc: Update SDK schema"
+            git push --set-upstream origin $pr_branch
+
+            gh pr create --title "Doc: Update SDK schema" --body "This PR updates the SDK schema." --base main
+          fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dmitry-lyfar, line 54: What this `Subproject commit 2375f1df668c18df991cfcde51bff81b1635f6e9` one is for in sdkcraft file?

I cannot comment there for some reason.

akcano: That was a leftover from local testing; I thought I deleted it, but no.
PR updated.

dmitry-lyfar (approved):

PR #602, opened 2026-01-14
Pass workshop.Snapshot into TakeSnapshot API

jonathan-conder:
# Description

Another small step towards reusing snapshots across multiple workshops.

We need a way to identify snapshots and store that info in the snapshot itself. Technically the backend can read identifying info from the workshop itself, but this proved to be a bit awkward to implement. Also, ideally the backend should be a dumb container runtime that doesn't know anything about workshop and SDK specifics.

In practice, the backend needs to know a little bit in order to function properly. For example, snapshot names must be valid LXD instance names. The compromise we came up with is to use `workshop.Snapshot` as a high level "snapshot identifier." The backend can use this to generate hashes and snapshot names without much coupling. Ultimately the `WorkshopManager` still decides when to take snapshots and what goes into them.

To make this work, I moved the `sdk.Setup` data for the workshop under construction from various `Task`s into the launch or refresh `Change`. This makes it easy for the `snapshot-sdk` and `create-workshops` tasks to construct a `workshop.Snapshot`, without duplicating the `sdk.Setup` data and bloating the state. Other tasks like `retrieve-sdk` and `install-sdk` also use this data. However, `unregister-sdk` still stores `sdk-setup` directly, since it comes from the old workshop and isn't stored in the `Change`.

In order to surface the `sdk.Setup` data at the API layer, which constructs the `Change`, I split the launch and refresh planning into two phases. The first phase queries the Store and other SDK sources, returning a `workshopstate.Manifest` which augments the workshop definition with a specific base image and SDK revisions. For refresh, it also reconstructs the workshop's current manifest. The second phase takes these manifests and plans out the tasks to run, much the same as before.

This should make both phases easier to test. I added a bunch of unit tests for the first phase, allowing us to drop a couple of API tests, but for the second phase I'll wait until the task planning has been reworked to support global snapshots.

A few minor changes which help to enable the above:
- Add a helper type `fileText` to simplify marshalling `workshop.File`. It folds YAML marshalling errors into JSON marshalling errors.
- Relax `testutil.DirEquals` so it no longer requires directory entries to be sorted.
- Fix an issue where the fake backend was creating SDK mount directories with `000` permissions.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:


internal/overlord/workshopstate/manifest.go
```diff
@@ -0,0 +1,467 @@
...
+
+type artifactFinder struct {
+	*WorkshopManager
+	user        *user.User
+	userDataDir string
+	project     workshop.Project
+	sdkVolumes  []workshop.SdkVolume
+}
+
+func (a *artifactFinder) latestManifests(ctx context.Context, names []string, refresh bool) ([]Manifest, []Manifest, error) {
```
dmitry-lyfar: It's worth commenting the return values. From the code it says `existing` and `updated`, I don't  understand what existing means: existing but not updated? I will likely understand it as I read more of the code, but it is not obvious at the first glance.

`latestManifests` assumes that there are some not latest manifests. `latestManifests` means building a set of up to date manifests?

internal/overlord/workshopstate/manifest.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	sdks, err := sto.SdkAction(ctx, acts)
+	if err != nil {
+		return nil, err
+	}
+	return validateSdkMeta(a.project.ProjectId, file, sdks)
+}
+
+func (a *artifactFinder) findLocalSdks(ctx context.Context, existing *Manifest, file *workshop.File) ([]sdk.Setup, error) {
```
dmitry-lyfar: Same with `existing` here, why `existing` and not just `manifest`?

internal/overlord/workshopstate/manifest.go
```diff
@@ -0,0 +1,467 @@
...
+		}
+		return nil, nil
+	}
+
+	return a.commitRevision(existing, w, sdk.Sketch, sdk.SketchSource, sketchdir)
+}
+
+func (a *artifactFinder) commitRevision(existing *Manifest, w, sk string, source sdk.Source, path string) (*sdk.Meta, error) {
+	var installed sdk.Revision
+	if existing != nil {
```
dmitry-lyfar: I'm guessing `existing == nil` reads as `an existing manifest is not existing`.

internal/overlord/sdkstate/handlers.go
```diff
@@ -19,71 +19,123 @@ import (
...
-func maybeSdkYaml(task *state.Task) (string, error) {
+func sdkName(task *state.Task) (string, error) {
 	st := task.State()
 	st.Lock()
 	defer st.Unlock()
 
-	var retrieveId string
-	var sdkYaml string
+	var setup sdk.Setup
+	if err := task.Get("sdk-setup", &setup); err == nil {
```
dmitry-lyfar: Maybe also set "sdk" key in the `Unregister()` but keep this querying only "sdk"?

internal/overlord/sdkstate/handlers.go
```diff
@@ -112,13 +164,19 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
...
 		st.Lock()
-		task.Set("sdk-setup", meta.Setup)
-		// Ideally we would validate the YAML here, but we can't check
-		// the base without knowing about the workshop. And this task
-		// should remain workshop-agnostic if possible. Instead we
-		// pass it to doInstallSdk and validate there.
-		task.Set("sdk-yaml", meta.SdkYAML)
+		change := task.Change()
+		replaceSdkSetup(change, w, meta.Setup)
+		var base string
```
```suggestion
		var base workshop.BaseImage
```

internal/overlord/workshopstate/manifest.go
jonathan-conder: I made some comments and just added names to the return values.

`latestManifests` was intended to contrast with `workshopManifests`, but since it now calls `workshopManifests` it's a bit stupid. I've renamed to reflect that this is a bit of a Frankenstein function at the moment, I want to get rid of it once we make this part async.

internal/overlord/workshopstate/manifest.go
jonathan-conder: I think it's fine to use `manifest` for launch because there's only one. But `findLocalSdks` is used for both launch and refresh. The purpose of this one is to ensure we don't delete an in-project or sketch SDK revision that's currently in use.

internal/overlord/workshopstate/manifest.go
jonathan-conder: I renamed `existing -> current` and `updated -> latest` to avoid this. Hopefully it makes the meaning more clear too

internal/overlord/sdkstate/handlers.go
jonathan-conder: I'm OK with that, will revisit next week though

dmitry-lyfar (approved):

PR #609, opened 2026-01-19
Doc: Simplify requirement updates with uv

akcano:


jonathan-conder:
I don't fully understand the context here, but why do we need custom code when we could be using proper dependency management e.g. `uv`?

akcano:
@jonathan-conder Mostly because the idea never occurred to me; I'll push a simplified version now.

.github/workflows/update-starter-pack.yaml
```diff
@@ -30,15 +33,14 @@ jobs:
...
           cp -a starter-pack-temp/docs/{Makefile,.sphinx} docs/
 
-          # Merge and sort requirements.txt files
-          cat docs/requirements.txt starter-pack-temp/docs/requirements.txt | sort -u > docs/requirements.txt.tmp
-          mv docs/requirements.txt.tmp docs/requirements.txt
+          # Merge requirements.txt files using uv
+          uv pip compile --universal \
+            docs/requirements.txt \
+            starter-pack-temp/docs/requirements.txt \
+            --output-file docs/requirements.txt
```
jonathan-conder: I think we should separate the input and output files, e.g. `docs/pyproject.toml` or `docs/requirements.in` as input, `docs/requirements.txt` output

akcano: Well noted; done.

jonathan-conder (approved):
Looks great! Hopefully it solves this class of issue for good

PR #614, opened 2026-01-26
Doc: Update 'Connect VSCode' with test, fix linter

### LINKED ISSUE #612: docs: `vscode-remote: SDK not found in "24.04/stable"`
I copy-pasted the `workshop.yaml` file from the documentation, but it's failing:

```
$ cat workshop.yaml 
name: dev
base: ubuntu@24.04
sdks:
  - name: vscode-remote
    channel: 24.04/stable
$ workshop refresh
error: cannot refresh "dev": SDK store action failed:
- vscode-remote: SDK not found in "24.04/stable"
"dev" refresh aborted
```

And then it's not clear to me how to list the available channels for a given SDK.

---
*Reported from: https://canonical-workshop.readthedocs-hosted.com/stable/how-to/develop-with-workshops/connect-vscode/*

akcano:
This fixes #612.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -107,8 +107,8 @@ linkcheck: install
 pa11y: pa11y-install html
 	find $(BUILDDIR) -name *.html -print0 | xargs -n 1 -0 $(PA11Y)
 
-lint-md: pymarkdownlnt-install
-	@. $(VENV); pymarkdownlnt --config $(SPHINXDIR)/.pymarkdown.json scan --recurse --exclude=./$(SPHINXDIR)/** $(SOURCEDIR)
+lint-md: install pymarkdownlnt-install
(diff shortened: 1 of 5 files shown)


dmitry-lyfar (changes requested):
It looks like you don't use the real store in the how-to tests and it's likely why it fails. The tutorial tests use `setup_workshop true`

dmitry-lyfar (approved):
@akcano I'm going to merge this one for the fix with Markdown, but can you comment on the choice to stop the store for running the test? Instead, we should make the other how-to doc test to use the actual store (as well as the entire how-to test suite).

akcano:
To make the discussion more visible here: I will move to bring ths and other how-to tests to consistency by switching to the real Store.

PR #615, opened 2026-01-26
Build code with verbose mode on for TiCS CI workflow

dmitry-lyfar:


.github/workflows/spread.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
         run: |
           suite_name="${{ matrix.suite }}"
           sanitized_name="${suite_name%/}"
+          echo "sanitized_name=$sanitized_name" >> "$GITHUB_OUTPUT"
```
jonathan-conder: Surprising, I didn't catch that one. It's also bad practice to embed templated strings directly in bash, use an env var instead.

dmitry-lyfar: Do  you mean something like this:

```
echo "SANITIZED_NAME=$sanitized_name" >> "$GITHUB_ENV"
```
?

jonathan-conder: Something like this (untested):
```
      - name: Trim trailing slash from matrix.suite
        id: sanitize
        run: |
          echo "sanitized_name=${SUITE_NAME%/}" >> "$GITHUB_OUTPUT"
        env:
          SUITE_NAME: ${{ matrix.suite }}
```

jonathan-conder (approved):

PR #617, opened 2026-01-28
Fix and prevent future bugs in launch and refresh

jonathan-conder:
# Description

I noticed a few issues while testing the global SDK changes:
- The fake backend mounts in-project SDKs from the wrong location. I think I did this deliberately since it wasn't needed at the time, but it has bitten me a few times now, and a future test will require this to be correct.
- Previously it was possible to list the `sketch` SDK in a workshop definition without the sketch SDK existing. Workshop would just ignore it. This now causes an error.
- Refreshing was a bit too eager (maybe intentional). Now the workshop won't refresh if the system or sketch SDKs are moved around in the workshop definition, or the connections are permuted.

I also came close to breaking `retrieve-sdk` recently, due to a combination of weak typing and no test coverage for the situation where the Store is updated mid-refresh. I added a test to make sure this keeps working until we can handle it properly (using the real Store).

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/overlord/workshopstate/manifest.go
+++ b/internal/overlord/workshopstate/manifest.go
@@ -283,6 +283,17 @@ func (a *artifactFinder) findLocalSdks(ctx context.Context, current *Manifest, l
 	}
 	if meta != nil {
 		sdks = append(sdks, *meta)
+	} else {
+		// If there's no sketch SDK, it's usually fine. But in rare
+		// cases (e.g. plug binding) the user might add the sketch SDK
+		// to the workshop definition. We shouldn't just ignore it,
+		// since it makes it harder to reason about the list of SDKs.
+		idx := slices.IndexFunc(latest.Sdks, func(s workshop.SdkRecord) bool {
+			return s.Source == sdk.SketchSource
+		})
+		if idx >= 0 {
+			return nil, fmt.Errorf("%q SDK not found, but appears in workshop definition", latest.Sdks[idx].Name)
+		}
 	}
 
 	return validateSdkMeta(a.project.ProjectId, latest, sdks)
--- a/internal/overlord/workshopstate/request.go
+++ b/internal/overlord/workshopstate/request.go
@@ -219,10 +219,10 @@ func launch(st *state.State, project workshop.Project, manifest Manifest) *state
 func (w *WorkshopManager) RefreshMany(project workshop.Project, current, latest []Manifest, option conflict.RefreshOption) ([]*state.TaskSet, error) {
 	tasksets := make([]*state.TaskSet, 0, len(latest))
 
-	for i, manifest := range latest {
-		plan := resolveRefresh(current[i], manifest)
-		if option == conflict.RefreshRestore || plan.HasUpdates() {
-			tasks := refresh(w.state, project, plan, manifest.File)
+	for i := range latest {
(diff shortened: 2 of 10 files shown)


internal/overlord/workshopstate/request.go
```diff
@@ -298,8 +294,56 @@ func (p refreshPlan) InstallIntactOrRefresh() []sdk.Setup {
...
+	seen := make([]bool, len(latest.Connections))
+out:
+	for _, conn := range current.Connections {
+		for i, other := range latest.Connections {
+			if !seen[i] && conn == other {
+				seen[i] = true
+				continue out
+			}
+		}
+		return true
```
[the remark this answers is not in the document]
jonathan-conder: I'll keep this option in mind but I think it's unlikely to be a problem give our current expected size of the connections table.

internal/overlord/workshopstate/manifest.go
```diff
@@ -283,6 +283,13 @@ func (a *artifactFinder) findLocalSdks(ctx context.Context, current *Manifest, l
 	}
 	if meta != nil {
 		sdks = append(sdks, *meta)
+	} else {
```
dmitry-lyfar, line 286: We found no sketch SDK but there still can be an SDK with the SketchSource type?

dmitry-lyfar: I probably misunderstand the use case, we try to find it and if successful we return an error that it's not found? Probably a comment needed.

internal/overlord/workshopstate/request_test.go
```diff
@@ -110,6 +135,169 @@ func (s *requestSuite) launchWorkshopWithSDKs(c *check.C, ws string, sdks []work
...
+	ts, err := s.mgr.RefreshMany(s.project, current, latest, conflict.RefreshUpdate)
+	c.Assert(err, check.IsNil)
+	c.Check(ts, check.HasLen, 0)
+
+	// No updates but user requested --restore.
+	ts, err = s.mgr.RefreshMany(s.project, current, latest, conflict.RefreshRestore)
+	c.Assert(err, check.IsNil)
+	c.Check(ts, check.Not(check.HasLen), 0)
+
+	// Updated SDK.
```
dmitry-lyfar, line 204: Maybe we should also account for the channel update here?

internal/overlord/workshopstate/manifest.go
jonathan-conder: if finding it fails, and the user put `- name: sketch` in their workshop file, then it's an error that it wasn't found

internal/overlord/workshopstate/request_test.go
jonathan-conder: the channel update currently forces a refresh, but it won't in my branch

internal/overlord/workshopstate/manifest.go
jonathan-conder: oh now I see how this could be confusing. Yeah I'll add a comment

dmitry-lyfar (approved):

PR #639, opened 2026-02-16
Doc: Integrate SDK documentation

akcano:


dmitry-lyfar:
The integration looks to be working, but I doubt we should continue here, let's discuss it on a call.

PR #641, opened 2026-02-18
Update LXD Go client

dmitry-lyfar:


internal/workshop/lxd/tests/helper/helper.go
```diff
@@ -59,7 +59,7 @@ func CleanupLxdProject(c *check.C, client lxd.InstanceServer, project string) {
 			}
 		}
 
-		op, err := cli.DeleteInstance(i.Name)
+		op, err := cli.DeleteInstance(i.Name, false)
```
jonathan-conder: If we make this `true` we can remove the preceding `UpdateInstanceState`. Could even force delete the entire project since the tests should always be running a recent LXD

internal/workshop/lxd/lxd_backend.go
```diff
@@ -929,7 +929,7 @@ func (s *Backend) RemoveWorkshop(ctx context.Context, name string) (err error) {
 		}
 	}
 
-	op, err := conn.DeleteInstance(InstanceName(name, projectId))
+	op, err := conn.DeleteInstance(InstanceName(name, projectId), true)
```
jonathan-conder:
```suggestion
	op, err := conn.DeleteInstance(InstanceName(name, projectId), false)
```

jonathan-conder (changes requested):

jonathan-conder (approved):

PR #644, opened 2026-02-20
Update LXD delete project to use the new API

dmitry-lyfar:


internal/workshop/lxd/tests/helper/helper.go
```diff
@@ -62,8 +62,9 @@ func CleanupLxdProject(c *check.C, client lxd.InstanceServer, project string) {
 		c.Check(err, check.IsNil)
 	}
 
-	err = cli.DeleteProject(project, false)
+	op, err := cli.DeleteProject(project, false)
 	c.Check(err, check.IsNil)
+	c.Check(op.Wait(), check.IsNil)
```
jonathan-conder:
```suggestion
	if c.Check(err, check.IsNil) {
		c.Check(op.Wait(), check.IsNil)
	}
```

jonathan-conder (approved):

internal/workshop/lxd/tests/helper/helper.go
dmitry-lyfar: I checked the implementation, it's a no-op Operation if the server supports only sync op (as any other command LXD managed to move between sync and async modes, so it's looks safe to assume that op is never nil if err == nil.

PR #668, opened 2026-03-18
Workshop and SDK CLI redesign

jonathan-conder:
# Description

Brings formatting of other commands in line with `sdk info`:
- Capitalise table headers
- Right-align numeric columns
- Hide headers if `--no-headers` is passed in (except for `sdk info` itself, which has 2 tables)
- Hide columns if they aren't relevant (i.e. the project column for `workshop sketches` and `workshop list` without `--global`)

On the other hand, `sdk info` now repeats project paths if they appear mutliple times. We decided to continue printing repeated projects in other commands like `workshop info`.

Previously `sdk list` sorted SDKs only by name. Now SDKs with the same name are sorted by descending revision number.

Fixes a bug where `workshop tasks` can have empty output if a `Change` failed before any tasks were added (in that case we check the next most recent task).

Expands shell completion:
- Only fall back to built-in path completion if it makes sense to do so
- Complete non-boolean optional arguments: `sdk info --arch|--base`, `workshop --project` and `workshop warnings --unicode`.
- Disable built-in path completion for `sketch-sdk --eject --name` and `workshopd --http`

Completion for `workshop --project` doesn't work as well as I'd like. This is due to cobra's `bash` and `zsh` scripts expanding the `toComplete` string and escaping the returned completions (including `~`). Maybe we should just use `ShellCompDirectiveFilterDirs` and call it a day.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -103,6 +103,13 @@ type Config struct {
 
 	// UserAgent is the User-Agent header sent to the Workshop daemon.
 	UserAgent string
+
+	// RetryInterval is the time to wait between GET retries. Negative
+	// values disable retries.
+	RetryInterval time.Duration
+
+	// Timeout is the time to spend retrying the request.
+	Timeout time.Duration
 }
 
 // A Client knows how to talk to the Workshop daemon.
@@ -110,6 +117,9 @@ type Client struct {
 	baseURL   url.URL
 	doer      doer
 	userAgent string
+
+	retryInterval time.Duration
+	timeout       time.Duration
 
 	maintenance error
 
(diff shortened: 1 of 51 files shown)


dmitry-lyfar (approved):
Looks nice. I tested everything with zsh manually too, works as expected. Once the tests are fixed, we should merge.

jonathan-conder:
I think the tests are fixed now. Also added a couple of things:
- Workshop snap now installs bash/fish/zsh completion scripts for `sdk` command
- workshop CLI doesn't retry `--project` completion if the daemon isn't running (I applied this globally although I think all other completions use POST which is never retried)

dmitry-lyfar (approved):

PR #684, opened 2026-03-31
Update lxd client library

dmitry-lyfar:


internal/workshop/lxd/lxd_backend.go
```diff
@@ -255,10 +258,15 @@ func New() (*Backend, error) {
...
 			pool.Config["size"] = strconv.FormatUint(storagePoolMinimalGiB*1024*1024*1024, 10)
-			if err = conn.UpdateStoragePool(storagePool, pool.Writable(), etag); err != nil {
+			op, err = conn.UpdateStoragePool(storagePool, pool.Writable(), etag)
+			if err != nil {
+				return nil, err
+			}
+			if err := op.Wait(); err != nil {
 				logger.Noticef("On Backend.New: failed to set storage pool to the minimal size: %dGiB, %s", storagePoolMinimalGiB, err)
 				return nil, err
 			}
```
jonathan-conder, line 268: Might this suppress the logs?

jonathan-conder (approved):

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: Do you mean the first `if`? I think these errors are LXD connectivity related (or just failures occurred before op is created). We log there to make the logging clearer when the operation was created but failed due to some reasons to indicate the size we attempted to set. I think an error message there would often be like "disk out of space" without any indication of how large we need.

PR #705, opened 2026-04-21
Doc: Document 'sdkcraft test'

akcano:


dmitry-lyfar (commented):
I will be merging this but it feels like we should have a how-to. `sdkcraft init` creates a template for the tests, that we don't seem to use in docs.

dmitry-lyfar (approved):

PR #714, opened 2026-04-27
Doc: Expand completion info

akcano:


cmd/internal/doctemplates/sdk.rst
```diff
@@ -34,20 +34,69 @@ and also has a number of global flags:
...
+.. code-block:: console
+
+   $ sdk completion fish | source
+
+
+PowerShell:
+
+.. code-block:: console
+
+   $ sdk completion powershell | Out-String | Invoke-Expression
```
jonathan-conder: I wouldn't add this one until we have some test coverage (and I probably wouldn't add that unless someone requests it)

cmd/internal/doctemplates/sdk.rst
```diff
@@ -34,20 +34,69 @@ and also has a number of global flags:
...
+
+Beyond subcommand and flag names,
+the :program:`sdk` CLI completes flag values for :command:`sdk info`:
+
+- :option:`!--base` completes the supported bases
+  (for example, :samp:`ubuntu@22.04`, :samp:`ubuntu@24.04`).
+
+- :option:`!--arch` completes the allowed architectures, plus :samp:`all`.
+
+
```
jonathan-conder: Not sure we need this section

cmd/internal/doctemplates/workshop.rst
```diff
@@ -39,20 +39,80 @@ and also has a number of global flags:
...
+
+- Plugs and slots for :command:`workshop connect`
+  and :command:`workshop disconnect`:
+  the first argument completes available plugs,
+  the second completes valid slots for the chosen plug.
+
+- Recent change IDs for :command:`workshop tasks`.
+
+- Choice values for flags such as :option:`!--unicode`
+  on :command:`workshop warnings`.
```
jonathan-conder: Definitely keep the first point. Not sure we need to explain `--project` or `--unicode`. I would probably keep plugs/slots and tasks though

docs/explanation/sdks/best-practices.rst
```diff
@@ -361,8 +361,8 @@ and system package manager integration:
    PATH="$SDK/bin:\$PATH"
    EOF
 
-   "$SDK"/bin/uv generate-shell-completion bash > /etc/bash_completion.d/uv.sh
-   "$SDK"/bin/uvx --generate-shell-completion bash > /etc/bash_completion.d/uvx.sh
+   "$SDK"/bin/uv generate-shell-completion bash > /usr/share/bash-completion/completions/uv
+   "$SDK"/bin/uvx --generate-shell-completion bash > /usr/share/bash-completion/completions/uvx
```
jonathan-conder: not sure what the motivation is for this, but `/usr/` should only be modified by `apt` (with the exception of `/usr/local/`) unless there's no way to avoid it. In this case we could use `/usr/local/`, but I would stick to `/etc/` for now since that matches the actual uv SDK.

docs/tutorial/part-1-get-started.rst
```diff
@@ -190,6 +190,23 @@ list the workshops in the project directory:
...
+   to autocomplete subcommands, flags, and arguments.
+
+   Completion is context-aware:
+   each command offers only values that make sense for it.
+   For instance, :command:`workshop start` autocompletes
+   from *Stopped* workshops only,
+   :command:`workshop stop` from *Ready* ones,
+   and :command:`workshop connect` autocompletes available plugs
+   and then the matching slots.
+   See :ref:`ref_workshop__cli_completion` for the full list.
```
jonathan-conder:
```suggestion
```
I think the examples are sufficient, maintaining a full list might be hard to maintain

jonathan-conder (changes requested):
I mostly object to modifying `/usr/share/`, other comments are just suggestions

jonathan-conder (approved):

dmitry-lyfar (approved):

PR #754, opened 2026-05-15
Doc: Refactor definition reference

akcano:


dmitry-lyfar:
It looks much better on the reference page now, but there are conflicts.

docs/reference/definition-files/_interfaces/desktop.rst
```diff
@@ -0,0 +1,16 @@
+..
```
dmitry-lyfar, line 1: These references are duplicated throughout Workshop, SDK and SDKcraft project definition references

docs/reference/definition-files/_interfaces/gpu.rst
```diff
@@ -0,0 +1,16 @@
...
+   Single-sourced snippet. Included by workshop-definition.rst,
+   sdk-definition.rst, and sdkcraft-definition.rst.
+   Do not add a top-level label; the including page provides the anchor.
+
+GPU interface
+~~~~~~~~~~~~~
+
+.. @artefact GPU interface
+
+The GPU interface exposes host GPU devices via LXD's CDI extension.
```
dmitry-lyfar:
```suggestion
The GPU interface exposes host GPU devices.
```
Leaks implementation detail.

dmitry-lyfar (changes requested):
I think navigation is broken (duplicated) here for the reference (on the left) and for the sections of the reference (on the right): interfaces references are duplicated across multiple pages.

docs/reference/definition-files/_interfaces/desktop.rst
akcano: This is intentional; these sections are (almost) identical in how we want to talk about them, so it makes sense to single-source them. The snippets themselves are excluded from the build and must not appear in the navigation either way.

akcano:
@dmitry-lyfar not sure what exactly appears broken; this is what I see in the PR build:

image
image

Each instance of '* interface' on the right is specific to the definition article that single-sources it via `include::`, so we don't have a situation where clicking 'SSH interface' under 'Workshop definition' ends up in 'SDK definition'. 

Am I missing something?

PR #759, opened 2026-05-20
Add firewall check on the startup

dmitry-lyfar:
# Description

The choice to use json output of `nft` and not parse older `iptables` format is intentional. The check is run on a startup which should be sufficient enough for the socket-activated daemon. The check accounts for the forward policy and for whether an exception is made for `workshopbr0` in the user-defined chains (we do not evaluate FORWARD chain rules only for simplicity and to avoid tracing all user defined chains from the FORWARD; it may result in some false positives, e.g. accept added to the INPUT chain but not to FORWARD, but that's unlikely as these rules would come from our warning recommendation). 

I decided not to add the check to the API layer to minimize the overhead even though the check itself is seemingly light. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/workshop/lxd/firewall.go
+++ b/internal/workshop/lxd/firewall.go
@@ -1,0 +1,200 @@
+package lxdbackend
+
+import (
+	"encoding/json"
+	"os/exec"
+	"strings"
+
+	"github.com/canonical/workshop/internal/logger"
+)
+
+const firewallDocLink = "https://documentation.ubuntu.com/lxd/latest/howto/network_bridge_firewalld/"
+
+// CheckBridgeFirewall detects whether firewall rules block forwarding on the
+// workshop bridge. It returns a non-empty warning message with a proposed
+// resolution if an issue is detected, or an empty string if everything looks
+// fine.
+//
+// Detection is cause-agnostic: it checks whether the FORWARD chain policy is
+// DROP/REJECT and whether any rules ACCEPT traffic for the bridge. The
+// remediation advice is cause-aware: it suggests Docker-specific, UFW-specific,
+// or generic commands depending on what is found.
+//
+// Uses `nft -j` (JSON output) for structured, robust parsing.
+//
+// See: https://documentation.ubuntu.com/lxd/latest/howto/network_bridge_firewalld/
+func CheckBridgeFirewall(bridgeName string) string {
+	return firewallChecker(bridgeName)
+}
+
+var firewallChecker = checkBridgeFirewall
+
+func checkBridgeFirewall(bridgeName string) string {
+	nft, err := exec.LookPath("nft")
+	if err != nil {
+		return ""
+	}
+
+	out, err := exec.Command(nft, "-j", "list", "table", "ip", "filter").CombinedOutput()
+	if err != nil {
+		logger.Noticef("%s", out)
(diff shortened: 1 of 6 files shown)


internal/workshop/lxd/firewall.go
```diff
@@ -0,0 +1,200 @@
...
+	return firewallChecker(bridgeName)
+}
+
+var firewallChecker = checkBridgeFirewall
+
+func checkBridgeFirewall(bridgeName string) string {
+	nft, err := exec.LookPath("nft")
+	if err != nil {
+		return ""
+	}
```
jonathan-conder, line 36: should we embed the `nft` CLI in the snap?

tests/main/firewall-docker/task.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  # Remove LXD so Docker is the first to enable ip_forward and sets FORWARD
+  # policy to DROP.
+  snap remove lxd --purge
+  sysctl -w net.ipv4.ip_forward=0
+
+  # Install Docker CE.
+  install -m 0755 -d /etc/apt/keyrings
+  curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+  chmod a+r /etc/apt/keyrings/docker.asc
```
jonathan-conder, line 13: why not install `docker.io` from the Ubuntu archive?

akcano (approved):
The help text is OK.

internal/workshop/lxd/firewall.go
```diff
@@ -0,0 +1,200 @@
...
+
+// forwardPolicyIsDrop returns true if the FORWARD chain has a "drop" policy.
+func forwardPolicyIsDrop(ruleset nftRuleset) bool {
+	for _, obj := range ruleset.Nftables {
+		if obj.Chain != nil && obj.Chain.Name == "FORWARD" {
+			return obj.Chain.Policy == "drop"
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
dmitry-lyfar: We are only interested in the FORWARD chain policy, which is most certainly just "drop" in the scenarios of interest (docker or ufw).

tests/main/firewall-docker/task.yaml
dmitry-lyfar: docker.io does not create the same set of rules as docker-ce afaik.

internal/workshop/lxd/firewall.go
dmitry-lyfar: I would probably not. It's a classic snap, if `nft` is not on the system, I would not go any further (I see `nft` is part of the major distros for a long time already).

tests/main/firewall-docker/task.yaml
jonathan-conder: I'm fine to leave it as-is, that is how I installed Docker in the past and I imagine more users do it that way than docker.io anyway

jonathan-conder (approved):

PR #765, opened 2026-05-22
Bump LXD dependency; increase timeouts for exec

dmitry-lyfar:


internal/overlord/cmdstate/manager.go
```diff
@@ -112,9 +113,7 @@ func (m *CommandManager) waitExecution(taskID string, stop <-chan struct{}) *exe
 		default:
 		}
 
-		m.executionsMutex.Lock()
 		e := m.executions[taskID]
-		m.executionsMutex.Unlock()
```
jonathan-conder, line 117: it's not clear to me that this is safe to remove

jonathan-conder (approved):
this isn't part of the codebase I'm familiar with, but seems OK for the most part. It's nice we can remove the GCS libraries now

internal/overlord/cmdstate/manager.go
dmitry-lyfar: The access is guarded by the mutex of the condition variable above (as in other cases when the map is read or modified).

PR #774, opened 2026-05-25
Bump version to 0.9.0

dmitry-lyfar:


jonathan-conder (approved):
Maybe we should merge the snapshot cleanup ahead of this? Assuming it's ready tomorrow as planned

PR #776, opened 2026-05-26
Fix dependabot

akcano:


jonathan-conder (approved):
I think the release notes should be a separate PR, but the dependabot stuff is fine

akcano:
my bad

PR #781, opened 2026-05-26
Fix zstd-compressed SDK extraction on minimal hosts

rajannpatel:
# Description

This PR resolves a WSL2 launch issue where SDK retrievals fail with a generic `exit status 2` error.

1. **Corrected `workshopd` PATH Environment Variable:** 
   In `snap/snapcraft.yaml`, added `${SNAP}/bin` to the `workshopd` daemon environment's `PATH`. This ensures that the `zstd` utility packaged inside the snap is accessible to the `tar` command during extraction. Without this, minimal host environments (like default WSL2 instances) that do not have `zstd` installed on the host fail to extract zstd-compressed `.sdk` archives.

2. **Enhanced Error Reporting in `extractSdkYAML`:**
   Modified `extractSdkYAML` in `internal/overlord/sdkstate/handlers.go` to capture the standard error stream of the executed `tar` command via a `bytes.Buffer`. If extraction fails, the command's standard error output is bubbled up with the error message, providing immediate diagnostic clarity (e.g. `tar (child): cannot run zstd: No such file or directory`) instead of a generic `exit status 2`.

# Self-review quick check

 * [x] Divide large expressions into digestable and self-explanatory ones.

## Docs

### DIFF
--- a/internal/overlord/sdkstate/handlers.go
+++ b/internal/overlord/sdkstate/handlers.go
@@ -127,7 +127,7 @@ func (m *SdkManager) doRetrieveSdk(task *state.Task, tomb *tomb.Tomb) error {
 		return err
 	}
 
-	sdkYaml, err := extractSdkYAML(ctx, rec)
+	sdkYaml, err := extractSdkYAML(ctx, task, rec)
 	if err != nil {
 		return err
 	}
@@ -211,7 +211,7 @@ func (m *SdkManager) retrieveSdk(ctx context.Context, task *state.Task, rec sdk.
 	return nil
 }
 
-func extractSdkYAML(ctx context.Context, rec sdk.Setup) (string, error) {
+func extractSdkYAML(ctx context.Context, task *state.Task, rec sdk.Setup) (string, error) {
 	path := rec.Filepath()
 	cache := path + ".yaml"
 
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -24,7 +24,7 @@ apps:
     command: bin/workshop
   workshopd:
     environment:
-      PATH: ${SNAP}/usr/bin:${PATH}
+      PATH: ${SNAP}/bin:${SNAP}/usr/bin:${PATH}
```
jonathan-conder:
```suggestion
      PATH: ${SNAP}/bin:${PATH}
```
there's actually no `usr` in the snap

internal/overlord/sdkstate/handlers.go
```diff
@@ -220,15 +221,20 @@ func extractSdkYAML(ctx context.Context, rec sdk.Setup) (string, error) {
...
 		"--force-local",
 		"--file="+path,
 		"meta/sdk.yaml",
 	)
+	cmd.Stderr = &stderr
 	content, err = cmd.Output()
 	if err != nil {
+		if stderr.Len() > 0 {
+			return "", fmt.Errorf("tar failed: %w (stderr: %q)", err, stderr.String())
+		}
```
jonathan-conder: The `err` should already have the `Stderr` attached. If you like I can rewrite it to use that

jonathan-conder (changes requested):

jonathan-conder (approved):

PR #797, opened 2026-05-28
Doc: Expose 'workshop init'

akcano:


docs/tutorial/part-1-get-started.rst
```diff
@@ -221,6 +221,36 @@ Here, the SDK is referenced as :samp:`ollama`,
 and the specific version to retrieve from the SDK Store
 comes from the :samp:`cpu/stable` channel of the the :samp:`cpu` track.
 
+Alternatively, instead of writing the definitions by hand,
```
dmitry-lyfar: There must be either way here. Yaml implies that you can write it by hand so I'm guessing you don't need the first part. Showing the workshop yaml you are about to launch should be enough.

dmitry-lyfar:
@akcano I think 2 spaces (as before) is better by default, so I updated the init and reverted some of your changes.

akcano:
@dmitry-lyfar ok, let's see how the tests fare

dmitry-lyfar (approved):

PR #798, opened 2026-05-28
Reject unknown fields in user-facing YAMLs

### LINKED ISSUE #686: lint sdk.yaml
I just spent too much time figuring out why this sdk.yaml wasn't working:

```
name: godepdiffer
plugs:
# forward your SSH agent into the workshop, enabling "git push" inside the workshop.
ssh-agent:
  interface: ssh-agent
# EXAMPLE: expose well-known config file locations to the workshop
# vs-code-settings:
#   interface: mount
#   workshop-target: /home/workshop/.config/Code/User
slots:
# EXAMPLE: expose SDK services to the host
# dashboard:
#   interface: tunnel
#   endpoint: 8080
```

The solution was to indent the ssh-agent block. 
If possible, I think it'd be nice to give the user a warning if there are unknown keys baked into the sdk.yaml.

Snehil-Shah:
# Description

Adds validation on SDK YAMLs to catch unknown fields silently passing:

**Unknown fields**:

```yaml
name: sketch
title: tweaks
plugs:
  models:
    interface: mount
    workshop-target: /home/workshop/models
hooks:
  setup-base: |
    apt-get install -y ffmpeg
verbose: true
```

would give

```
error: sketch SDK YAML contains unknown fields: "verbose" at line 12, column 10
```

**Unsupported hook**:

```yaml
name: sketch
title: tweaks
plugs:
  models:
    interface: mount
    workshop-target: /home/workshop/models
hooks:
  setup-base: |
    apt-get install -y ffmpeg
  check-helath: |
    ffmpeg -version
```

would give

```
error: sketch SDK YAML contains unsupported hook "check-helath"; supported hooks: setup-base, setup-project, save-state, restore-state, check-health
```

Resolves #686 

# Self-review quick check

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"reflect"
 	"slices"
 	"strings"
 
@@ -230,22 +229,19 @@ func ejectSketch(project, sketchdir string, name string) (*revert.Reverter, erro
 	}
 	file, err := sdk.ParseSketchYaml(bytes.NewReader(content))
 	if err != nil {
-		return nil, err
+		return nil, parseSketchYamlUserError(err)
 	}
 
 	var document yaml.Node
 	if err := yaml.Unmarshal(content, &document); err != nil {
 		return nil, err
 	}
-	file.Name = name
+	projectFile := file
+	projectFile.Name = name
 
-	if err := sketchToProjectSdk(&document, name); err != nil {
+	if err := sketchToProjectSdk(&document, projectFile.Name); err != nil {
 		return nil, err
 	}
-	if err := validateProjectSdk(&document, &file); err != nil {
-		return nil, err
-	}
-
 	// This won't be cleaned up on failure. In most cases, the user
 	// is likely to retry the eject after fixing the underlying issue.
 	if err := os.MkdirAll(filepath.Dir(target), 0755); err != nil {
@@ -291,39 +287,15 @@ func sketchToProjectSdk(document *yaml.Node, name string) error {
 		Name  NodeRef `yaml:"name"`
 		Hooks NodeRef `yaml:"hooks"`
 	}
-	if err := document.Decode(&nodes); err != nil {
+	err := document.Decode(&nodes)
+	if err != nil {
 		return err
 	}
 
-	if nodes.Name.Node == nil {
-		return errors.New(`"sketch" SDK name not found`)
-	}
 	nodes.Name.Node.Value = name
 
 	if nodes.Hooks.Node != nil {
 		RemoveNodes(document, nodes.Hooks.Node)
-	}
-
-	return nil
-}
-
-func validateProjectSdk(document *yaml.Node, expected *sdk.SketchSDKYaml) error {
-	var actual sdk.SketchSDKYaml
-	if err := document.Decode(&actual); err != nil {
-		return err
-	}
-
-	if actual.Name != expected.Name {
-		return fmt.Errorf("internal error: sketch SDK renamed %q (expected %q)", actual.Name, expected.Name)
-	}
-	if !reflect.DeepEqual(expected.Plugs, actual.Plugs) {
-		return errors.New("internal error: sketch SDK plugs not preserved")
-	}
-	if !reflect.DeepEqual(expected.Slots, actual.Slots) {
-		return errors.New("internal error: sketch SDK slots not preserved")
-	}
-	if len(actual.Hooks) > 0 {
-		return errors.New("internal error: hooks not ejected from sketch SDK")
 	}
 
 	return nil
@@ -580,9 +552,9 @@ func editSketchSdk(sketchdir string) error {
 }
 
 // parseSketchYamlUserError converts structured [sdk.ParseSketchYaml] errors
-// into messages that tell sketch SDK users what to fix in their edited YAML.
-// Use it when reporting parse failures from the interactive sketch edit flow,
-// where raw validator errors can be too generic or expose SDK-internal terms.
+// into messages that tell sketch SDK users what to fix in their YAML. Use it
+// when reporting parse failures from sketch command flows, where raw validator
+// errors can be too generic or expose SDK-internal terms.
 func parseSketchYamlUserError(err error) error {
 	var (
 		invalidHook  sdk.InvalidSDKHookNameError
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -571,7 +571,7 @@ func parseSketchYamlUserError(err error) error {
 		return fmt.Errorf(
 			"sketch SDK YAML contains unsupported hook %q; supported hooks: %s",
 			string(invalidHook),
-			strings.Join(sdk.AllowedSketchHooks, ", "),
+			strings.Join(sdk.AllowedHooks, ", "),
 		)
 	case errors.As(err, &unknownField):
 		parts := make([]string, 0, len(unknownField.Fields))
--- a/cmd/workshop/sketch.go
+++ b/cmd/workshop/sketch.go
@@ -102,13 +102,6 @@ $ workshop sketch-sdk nimble --stash`,
 	return cmd
 }
 
-type SketchFile struct {
-	Name  string            `yaml:"name"`
-	Plugs map[string]any    `yaml:"plugs,omitempty"`
-	Slots map[string]any    `yaml:"slots,omitempty"`
-	Hooks map[string]string `yaml:"hooks,omitempty"`
-}
-
 var sketchTemplate = `# For details: https://ubuntu.com/workshop/docs/explanation/sdks/concepts/#sketch-sdk
 name: sketch
 hooks:
@@ -235,15 +228,8 @@ func ejectSketch(project, sketchdir string, name string) (*revert.Reverter, erro
 	} else if err != nil {
 		return nil, err
 	}
-	var file SketchFile
-	dec := yaml.NewDecoder(bytes.NewReader(content))
-	dec.KnownFields(true)
-	if err := dec.Decode(&file); err != nil {
-		te, ok := err.(*yaml.TypeError)
-		if ok {
-			errs := strings.Join(te.Errors, "\n")
-			return nil, fmt.Errorf("sketch SDK YAML:\n%s", errs)
-		}
+	file, err := sdk.ParseSketchYaml(bytes.NewReader(content))
+	if err != nil {
 		return nil, err
 	}
 
@@ -321,8 +307,8 @@ func sketchToProjectSdk(document *yaml.Node, name string) error {
 	return nil
 }
 
-func validateProjectSdk(document *yaml.Node, expected *SketchFile) error {
-	var actual SketchFile
+func validateProjectSdk(document *yaml.Node, expected *sdk.SketchSDKYaml) error {
+	var actual sdk.SketchSDKYaml
 	if err := document.Decode(&actual); err != nil {
 		return err
 	}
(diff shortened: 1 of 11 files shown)


internal/sdk/sdk.go
```diff
@@ -274,8 +274,14 @@ var SanitizePlugsSlots = func(snapInfo *Info) {
...
-	err := yaml.Unmarshal(yamlData, &sdkYaml)
-	if err != nil {
+	dec := yaml.NewDecoder(bytes.NewReader(yamlData))
+	dec.KnownFields(true)
+	if err := dec.Decode(&sdkYaml); err != nil {
+		te, ok := err.(*yaml.TypeError)
+		if ok {
+			errs := strings.Join(te.Errors, "\n")
+			return nil, fmt.Errorf("SDK definition YAML:\n%s", errs)
+		}
```
Snehil-Shah: @dmitry-lyfar One thing, I decided to use the error message from the decoder itself, which is pretty informative but slightly leaks the internal type names:

```
error: SDK definition YAML:
line 3: field ssh-agent not found in type sdk.sdkYaml
```

The user doesn't need to know what the type is called internally. I still left it as is to keep the code simpler. If you want, we can manually parse the error message to find the unknown field and output a more presentable error message, and possibly abstract the entire decoding step into a new metautil named `StrictUnmarshal` or smth. WDYT?

cmd/workshop/sketch.go
```diff
@@ -234,13 +235,20 @@ func ejectSketch(project, sketchdir string, name string) (*revert.Reverter, erro
 	} else if err != nil {
 		return nil, err
 	}
-	var document yaml.Node
-	if err := yaml.Unmarshal(content, &document); err != nil {
+	var file SketchFile
```
tlm: This change is potentially a little dangerous IMO.

If we look at `SketchFile`, it only works on a small subset of the fields that can actually appear in an SDK YAML file. For example:

```go
type SketchFile struct {
	Name  string            `yaml:"name"`
	Plugs map[string]any    `yaml:"plugs,omitempty"`
	Slots map[string]any    `yaml:"slots,omitempty"`
	Hooks map[string]string `yaml:"hooks,omitempty"`
}
```

versus:

```go
type sdkYaml struct {
	Name        string            `yaml:"name"`
	Base        string            `yaml:"base"`
	Arch        string            `yaml:"architecture"`
	Version     string            `yaml:"version,omitempty"`
	Title       string            `yaml:"title"`
	Summary     string            `yaml:"summary"`
	Description string            `yaml:"description"`
	License     string            `yaml:"license"`
	Type        string            `yaml:"type"`
	BuiltAt     *timeutil.TimeUTC `yaml:"sdkcraft-started-at,omitempty"`
	Plugs       map[string]any    `yaml:"plugs,omitempty"`
	Slots       map[string]any    `yaml:"slots,omitempty"`
}
```

Not all of the full set makes sense for a sketch SDK, but I think we should support as much overlap as is reasonable. My reasoning is that it allows people to copy and paste quickly and lets Workshop get out of their way. For example, supporting `title` and `description` seems reasonable.

At the moment, if we added `title: epic title`, we would end up with an error like this:

```
error: cannot sketch: sketch SDK YAML:
line 3: field title not found in type main.SketchFile
```

cmd/workshop/sketch.go
```diff
@@ -234,13 +235,20 @@ func ejectSketch(project, sketchdir string, name string) (*revert.Reverter, erro
...
+	dec := yaml.NewDecoder(bytes.NewReader(content))
+	dec.KnownFields(true)
```
tlm: I think setting `KnownFields` to true here is great. My problem is that, by going down this path, we are delegating authority over our UX to a third-party library that we do not maintain any contracts with. At the moment, we get an error similar to the following:

```
error: cannot sketch: sketch SDK YAML:
line 3: field title not found in type main.SketchFile
line 16: field ssh-agent not found in type main.SketchFile
```

This sort of works because it includes line numbers, which are a great pointer to where the problem is. However, it also references internal application types that only add confusion.

On top of this, we are relying on the library we use to keep its error messages stable and to always format them in a way that works for our UX. We cannot maintain these guarantees and should not delegate this problem.

Instead, we can place an `Unknown` member on the struct like this:

```go
Unknown map[string]yaml.Node `yaml:",inline"`
```

This allows us to capture all of the context we need and produce an error message that is in line with the UX we need to deliver. We have used a similar pattern for the workshop YAML file [here](https://github.com/canonical/workshop/blob/main/cmd/workshop/yaml.go#L23-L32).

cmd/workshop/sketch.go
```diff
@@ -586,7 +594,14 @@ func writeSketchSdk(path string, content []byte) error {
 
 func writeSketchHooks(sketchdir string, content []byte) error {
 	var file SketchFile
-	if err := yaml.Unmarshal(content, &file); err != nil {
+	dec := yaml.NewDecoder(bytes.NewReader(content))
```
tlm: By adding `KnownFields` here, we are overloading `writeSketchHooks` so that it also validates the YAML for us.

We should split the dual purpose and perform parsing and validation before calling this.

`writeSketchHooks` also shows another subtle failure path that we should consider handling as part of this PR: the hook names themselves.

The set of permissible values is:

```go
[]string{"setup-base", "setup-project", "save-state", "restore-state", "check-health"}
```

If a user mistypes one of them, we will fail silently, as we do for unknown fields.

I suggest making our sketch SDK validation include these hook names, as they are well defined and easy to reason about.

internal/workshop/workshop_file.go
```diff
@@ -289,7 +290,9 @@ func readWorkshop(path string) (*File, error) {
 		}
 		return nil, err
 	}
-	if err = yaml.Unmarshal(buf, &file); err != nil {
+	dec := yaml.NewDecoder(bytes.NewReader(buf))
```
tlm: I would probably avoid this change in this PR. We should understand the side effects and propose the change in a separate PR. That keeps this one focused on SDK handling.

tlm (commented):
Hey @Snehil-Shah,

Thank you for the contribution and the awesome work. I have left some PR feedback for you. My main request/concern is about understanding the error messages we deliver in our UX.

It is important that we not only generate errors for SDK files, but also communicate the problem to the end user in a meaningful and repeatable way.

When we are deep in the call stack, our focus should be on typed errors that convey the problem programmatically. When we reach the UX layer, we can then inspect those errors and generate user-facing error messages.

This is important as the project grows in size and features. It gives us a good design foundation for building meaningful user experiences.

I took the opportunity to provide a few commits on top of your work here: https://github.com/tlm/workshop/commits/lint-sdk/. They demonstrate some of the ideas from my PR feedback comments.

If you agree with the direction, I would suggest applying this pattern to the other parts of the PR. Feel free to apply your own style or direction. The commits are intended to demonstrate the concept rather than prescribe the exact implementation.

Please reach out if you would like help or would like to work on this together.

May I also suggest adding some QA/output examples to the PR description? It would be useful to demonstrate the errors produced by this change and the ways we now communicate problems as a result of this work.

cmd/workshop/sketch.go
Snehil-Shah: I have updated such that `writeSketchHooks` directly receives the unmarshalled file and it can do its one job it is meant to do.

Snehil-Shah:
@tlm I think you linked to a private fork that I can't access, but I went ahead and made the changes from my interpretation of your review:

- Punting the changes to workshop yaml.
- We now validate after YAML unmarshalling is done, by collecting the extra fields we don't recognize, and throwing an error for them.

Also added the updated behavior to the PR description.

cmd/workshop/sketch.go
```diff
@@ -102,10 +102,74 @@ $ workshop sketch-sdk nimble --stash`,
...
-	Name  string            `yaml:"name"`
-	Plugs map[string]any    `yaml:"plugs,omitempty"`
-	Slots map[string]any    `yaml:"slots,omitempty"`
-	Hooks map[string]string `yaml:"hooks,omitempty"`
+	Name        string               `yaml:"name"`
+	Title       string               `yaml:"title,omitempty"`
+	Summary     string               `yaml:"summary,omitempty"`
+	Description string               `yaml:"description,omitempty"`
+	License     string               `yaml:"license,omitempty"`
+	Version     string               `yaml:"version,omitempty"`
```
Snehil-Shah: @tlm These are the fields that I think wouldn't hurt to include in the sketch file (mostly metadata), just so we don't throw errors on copy-pasted SDK YAMLs. WDYT?

tlm:
@Snehil-Shah, apologies have updated my fork to be public.

Snehil-Shah:
@tlm I went ahead and cherry-picked your commits to this branch. They make more sense than the approach I was going for. I have wired the validator in and added an integration test for the same. Thanks for the solid reference!

Snehil-Shah:
There is one thing: the current implementation doesn't tell us all the errors across stages (or multiple unsupported hooks) at once. Should we keep it that way? It can be helpful to know all the issues at once. WDYT?

internal/overlord/workshopstate/manifest_test.go
```diff
@@ -898,18 +894,34 @@ func (s *manifestSuite) TestLaunchValidatesProjectSdks(c *check.C) {
...
+ssh-agent:
+  interface: ssh-agent
 `)
 
 	sdks := []workshop.SdkRecord{{Name: "test", Source: sdk.ProjectSource}}
 	s.createWFile(c, "test-1", "ubuntu@20.04", sdks)
 
 	_, err := s.manager.LaunchManifests(s.ctx, s.project, []string{"test-1"})
-	c.Assert(err, check.ErrorMatches, `cannot launch "test-1": "test" SDK has "mock32" architecture; required: "mock64" or "all"`)
+	c.Assert(err, check.ErrorMatches, `cannot launch "test-1": invalid "test" SDK: unknown SDK YAML fields: ssh-agent \(line \d+, column \d+\)`)
```
tlm:
```suggestion
	var unknown *sdk.UnknownYamlFieldsError
	ok := errors.As(err, &unknown)
	c.Assert(ok, check.Equals, true)
	c.Check(unknown.Fields, check.DeepEquals, map[string]sdk.UnknownYamlField{
		"ssh-agent": {
			Line:   3,
			Column: 3,
		},
	})
```

This explicitly checks for what we care about and makes the test less fragile when dealing with change.

tlm (approved):
Thanks @Snehil-Shah for the work on this.

Can I suggest also add this diff to your PR:

```
diff --git a/tests/main/sketch-sdk/task.yaml b/tests/main/sketch-sdk/task.yaml
index 885e8124..8c5e4031 100644
--- a/tests/main/sketch-sdk/task.yaml
+++ b/tests/main/sketch-sdk/task.yaml
@@ -16,6 +16,22 @@ execute: |
   workshop_exec info | yq '.sdks["sketch"].tracking' | MATCH '^~/.local/share/workshop/id/.*/ws-sketch/sdk/sketch'
   workshop_exec info | yq '.sdks["sketch"].installed' | MATCH '\(x1\)$'

+  echo "Reject unknown sketch SDK YAML fields"
+  ! sudo TERM=ansi EDITOR=ex -u ubuntu -- workshop sketch-sdk > error.log 2>&1 <<EOF || false
+  1/^name: sketch$/a
+  base: ubuntu@24.04
+  .
+  wq
+  EOF
+  MATCH 'cannot sketch: sketch SDK YAML contains unknown fields:' < error.log
+  MATCH '"base" at line [0-9]+, column [0-9]+' < error.log
+
+  echo "Restore valid sketch SDK YAML"
+  sudo TERM=ansi EDITOR=ex -u ubuntu -- workshop sketch-sdk <<EOF
+  /^base: ubuntu@24.04$/d
+  wq
+  EOF
+
   echo "Stash current sketch SDK"
   workshop_exec sketch-sdk --stash
   workshop_exec info | yq '.sdks["sketch"].installed' | NOMATCH '\(x1\)$'
```

This adds a spread test to assert some of the cli behaviour introduced.

internal/sdk/validate.go
```diff
@@ -15,23 +15,149 @@
...
+	// ErrorInvalidSDKName reports an invalid SDK name.
+	ErrorInvalidSDKName = errors.New("invalid SDK name")
+)
+
 var (
 	AllowedBases = []string{"ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "ubuntu@26.04"}
-	sdkName      = regexp.MustCompile(`^(?:[a-z0-9]-?)*[a-z](?:-?[a-z0-9])*$`)
+
+	// AllowedSketchHooks lists hook names accepted in sketch SDK YAML.
+	AllowedSketchHooks = []string{"setup-base", "setup-project", "save-state", "restore-state", "check-health"}
```
dmitry-lyfar:
```suggestion
	AllowedHooks = []string{"setup-base", "setup-project", "save-state", "restore-state", "check-health"}
```

tlm: Updated

tlm:
@Snehil-Shah. everything looks ready to go. I think all you need to do is sign the commits that are not signed. I also pushed a quick change to this for @dmitry-lyfar feedback.

Snehil-Shah:
@tlm done.

PR #815, opened 2026-06-02
Allow test override for snapshot cleanup cooldown

### LINKED ISSUE #799: Unused SDK snapshots remain after successful workshop remove
## Summary

Unused SDK snapshot containers can remain in the `workshop-snapshots.<user>` LXD project after a successful `workshop remove`.

I reproduced this with a normal workflow using the same project-local SDK name:

```text
launch -> refresh -> refresh -> remove
```

After waiting longer than the snapshot cleanup cooldown, the old SDK snapshot containers were still present, and `workshopd` repeatedly logged:

```text
On WorkshopManager.doDeleteUnusedSnapshots: internal error: internal error: "snapshot-check" workshop old-stash format not found (change ID: 7)
```

`change ID: 7` was the successful `Remove "snapshot-check" workshop` change.

## What I observed

The workshop operations completed successfully:

```text
Launch "snapshot-check" workshop: Done
Refresh "snapshot-check" workshop: Done
Refresh "snapshot-check" workshop: Done
Remove "snapshot-check" workshop: Done
```

Each refresh created a new SDK snapshot as expected. The running workshop moved from one `snapshot-check-base-*` snapshot to the next.

After `workshop remove snapshot-check`, there was no current `snapshot-check` container in the `workshop.<user>` LXD project. However, after waiting more than one hour, the `workshop-snapshots.<user>` project still contained the old SDK snapshots:

```text
snapshot-check-base-53ce2b4a882f4431
snapshot-check-base-3711a2463136346d
snapshot-check-base-d386f84cc368d9f8
system-2ae1192172a14642
```

## Expected behavior

After a normal `workshop remove` completes successfully, unused SDK snapshot containers should eventually be cleaned up by the automatic snapshot cleanup mechanism.

Users should not need to manually purge LXD snapshot containers for this normal workflow.

## Actual behavior

The CLI operations complete successfully, but the automatic snapshot cleanup appears to get stuck when it hits a stale `old-stash` last-used record.

The daemon repeatedly logs:

```text
"snapshot-check" workshop old-stash format not found
```

and the unused SDK snapshot containers remain.

## Notes on snapshot cleanup behavior

From reading the implementation, snapshot cleanup appears to be intentionally conservative:

* it tracks snapshot last-used state through changes/tasks
* it waits for relevant changes to complete
* it respects the cleanup cooldown
* it checks whether the backend still reports workshops using a snapshot before removing it

That behavior makes sense, and this issue is not about making snapshot deletion more aggressive.

The problem seems to be that cleanup can stop before reaching those existing safety checks. In this case, a completed remove change still has a last-used record for `old-stash`, but the old-stash metadata is no longer available. This is treated as an internal error, so cleanup stops and the unused snapshots remain.

## Proposed direction

A defensive fix could treat a missing `old-stash` last-used record as stale/absent state during snapshot cleanup, rather than as an error that blocks cleanup indefinitely.

The existing deletion safeguards should remain unchanged. The fix would only allow cleanup to skip stale `old-stash` records and continue to the existing checks, such as cooldown and backend usage checks.

I tested this direction locally by making `SnapshotLastUsed` skip stale `OldStash` records that can no longer be resolved. The regression test fails before the change with:

```text
internal error: "snapshot-check" workshop old-stash format not found
```

and passes after the defensive handling is added.

## Tests run

```text
gofmt -w internal/overlord/handlersetup/handler_setup.go   internal/overlord/handlersetup/handler_setup_test.go
go test -count=1 ./internal/overlord/handlersetup
go test -count=1 ./internal/overlord/workshopstate
go test -count=1 ./...
golangci-lint run
```

I have a small defensive fix prepared and can open a PR for this.

shizuku198411:
Fixes #799

# Description
This PR adds a guarded test-only override for the snapshot cleanup cooldown.

By default, the snapshot cleanup cooldown remains unchanged at one hour. The override is only applied when test overrides are explicitly enabled with `WORKSHOP_ENABLE_TEST_OVERRIDES=1`, and the cooldown value is provided through `WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN`.

This is intended to make end-to-end tests easier to write without waiting for the full production cooldown, while keeping production behavior unchanged unless the explicit test override flag is set.

The added unit tests cover:

* default behavior without environment variables;
* ignoring the cooldown variable unless test overrides are enabled;
* applying a valid override;
* allowing `0s` for tests;
* falling back to the default for invalid or negative values.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/overlord/workshopstate/manager.go
+++ b/internal/overlord/workshopstate/manager.go
@@ -34,21 +34,13 @@ type WorkshopManager struct {
 	firewallChecker func(string) string
 }
 
-const (
-	defaultSnapshotCooldownTime = 1 * time.Hour
-
-	testSnapshotCleanupCooldownEnv = "WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN"
-)
+const defaultSnapshotCooldownTime = 1 * time.Hour
 
 var (
 	snapshotCooldownTime = snapshotCooldownTimeFromEnvironment()
 )
 
 func snapshotCooldownTimeFromEnvironment() time.Duration {
-	if os.Getenv(testOverridesEnv) != "1" {
-		return defaultSnapshotCooldownTime
-	}
-
 	value := os.Getenv("WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN")
 	if value == "" {
 		return defaultSnapshotCooldownTime
@@ -56,12 +48,12 @@ func snapshotCooldownTimeFromEnvironment() time.Duration {
 
 	cooldown, err := time.ParseDuration(value)
 	if err != nil {
-		logger.Noticef("Ignoring invalid %s=%q: %v", testSnapshotCleanupCooldownEnv, value, err)
+		logger.Noticef("Ignoring invalid WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN=%q: %v", value, err)
 		return defaultSnapshotCooldownTime
 	}
 
 	if cooldown < 0 {
-		logger.Noticef("Ignoring negative %s=%q", testSnapshotCleanupCooldownEnv, value)
+		logger.Noticef("Ignoring negative WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN=%q", value)
 		return defaultSnapshotCooldownTime
 	}
 
--- a/internal/overlord/workshopstate/manager.go
+++ b/internal/overlord/workshopstate/manager.go
@@ -37,7 +37,6 @@ type WorkshopManager struct {
 const (
 	defaultSnapshotCooldownTime = 1 * time.Hour
 
-	testOverridesEnv               = "WORKSHOP_ENABLE_TEST_OVERRIDES"
 	testSnapshotCleanupCooldownEnv = "WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN"
 )
 
@@ -50,7 +49,7 @@ func snapshotCooldownTimeFromEnvironment() time.Duration {
 		return defaultSnapshotCooldownTime
 	}
 
-	value := os.Getenv(testSnapshotCleanupCooldownEnv)
+	value := os.Getenv("WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN")
 	if value == "" {
 		return defaultSnapshotCooldownTime
 	}
--- a/internal/overlord/workshopstate/manager.go
+++ b/internal/overlord/workshopstate/manager.go
@@ -17,6 +17,7 @@ package workshopstate
 import (
 	"context"
 	"fmt"
+	"os"
 	"slices"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/overlord/workshopstate/manager.go
```diff
@@ -33,10 +34,41 @@ type WorkshopManager struct {
 	firewallChecker func(string) string
 }
 
+const (
+	defaultSnapshotCooldownTime = 1 * time.Hour
+
+	testOverridesEnv               = "WORKSHOP_ENABLE_TEST_OVERRIDES"
```
jonathan-conder:
```suggestion
```

this doesn't make it more secure, just more involved

internal/overlord/workshopstate/manager.go
```diff
@@ -33,10 +34,41 @@ type WorkshopManager struct {
...
-	snapshotCooldownTime = 1 * time.Hour // Time to wait before deleting unused snapshots.
+	snapshotCooldownTime = snapshotCooldownTimeFromEnvironment()
 )
 
+func snapshotCooldownTimeFromEnvironment() time.Duration {
+	if os.Getenv(testOverridesEnv) != "1" {
+		return defaultSnapshotCooldownTime
+	}
+
+	value := os.Getenv(testSnapshotCleanupCooldownEnv)
```
jonathan-conder:
```suggestion
	value := os.Getenv("WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN")
```
improves locality, we don't need 2 names for the same variable

internal/overlord/workshopstate/manager_cooldown_test.go
```diff
@@ -0,0 +1,98 @@
...
+var _ = check.Suite(&snapshotCooldownSuite{})
+
+func (s *snapshotCooldownSuite) TestSnapshotCooldownTimeFromEnvironmentDefaultsToOneHour(c *check.C) {
+	defer restoreEnv(testOverridesEnv)()
+	defer restoreEnv(testSnapshotCleanupCooldownEnv)()
+
+	c.Assert(os.Unsetenv(testOverridesEnv), check.IsNil)
+	c.Assert(os.Unsetenv(testSnapshotCleanupCooldownEnv), check.IsNil)
+
+	c.Assert(snapshotCooldownTimeFromEnvironment(), check.Equals, defaultSnapshotCooldownTime)
```
jonathan-conder: The environment changes could benefit from some refactoring into `SetUpTest` and `TearDownTest`

jonathan-conder (changes requested):
I would also make this configurable using `snap set`

shizuku198411:
Thanks for the review!

I updated the PR to address the comments:

* removed the extra `WORKSHOP_ENABLE_TEST_OVERRIDES` guard;
* kept the cooldown environment variable local to the code that reads it;
* refactored the test environment setup/cleanup into `SetUpTest` and `TearDownTest`;
* wired the setting through snap config, similar to existing settings like `WORKSHOP_DEBUG`.

The snap configuration can now be set with:

```text
sudo snap set workshop workshop.test.snapshot-cleanup-cooldown=1m
sudo snap restart workshop
```

I also checked this locally with a rebuilt snap. `snap get` returns the configured value:

```text
$ sudo snap get workshop workshop.test.snapshot-cleanup-cooldown
1m
```

and the value is passed to `workshopd` as expected:

```text
WORKSHOP_TEST_SNAPSHOT_CLEANUP_COOLDOWN=1m
```

jonathan-conder:
Sorry to be a pain, but after thinking about this I don't think we need full end-to-end testing.

The issue with the unit tests is that they don't incorporate the full launch/refresh/remove planning logic. We can exercise that using the tests in `internal/daemon`, without being fully end-to-end. It should work better.

So instead of making the cooldown time customizable, would you like to incorporate some snapshot cleanup checks into the existing `apiSuite`? A few scenarios may require new tests.

shizuku198411:
That makes sense, thanks for explaining!

I agree that the daemon API tests sound like a better fit than making the cooldown configurable just for this. Since #812 is still in draft and seems to be where the intended `OldStash` / snapshot cleanup behavior is being settled, I think it would be better to wait for that direction to become clear first.

I’ll close this PR for now to avoid adding an unnecessary configuration path. Once the root fix and intended behavior are settled, I can take another look at adding coverage in `internal/daemon` / `apiSuite` if it would still be useful.

jonathan-conder:
Thanks for understanding. I wrote some API tests as well so I'll close this one now.

PR #817, opened 2026-06-03
Remember connections across refreshes

jonathan-conder:
# Description

Previously, refresh would remember `host-source` for remounted mount connections. Now, it remembers all dynamic attributes (which should have the same effect). It also remembers manual connections and disconnections.

Bound plugs are reconnected (or kept disconnected) alongside the plug they are bound to, regardless of the previous state of the plug itself. Plugs that used to be bound, but are now independent, are handled transparently, i.e. dynamic attributes come from the main connection rather than one of the bound copies.

Unlike refresh, restore resets manual connections, disconnections and remounts, as if the user removed and re-launched the workshop.

I also fixed a usability issue that isn't really a bug, but I think makes sense to do given the other changes. Previously, manually disconnecting a manual connection would not store anything in the state. I assume the reasoning was that connections are only created manually if they aren't auto-connectable. Now that manual connections are preserved by refresh, policies can change, and a manual disconnection might be undone on the next refresh. So we should always mark manual disconnections as undesired. Temporary disconnections can still be done with `disconnect --forget`.

Finally, I changed some error handling from logging into daemon degradation, otherwise the interface manager might prune some connections that we want to keep.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/overlord/handlersetup/handler_setup.go
+++ b/internal/overlord/handlersetup/handler_setup.go
@@ -27,6 +27,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/canonical/workshop/internal/overlord/conflict"
+	"github.com/canonical/workshop/internal/overlord/ifacestate/schema"
 	"github.com/canonical/workshop/internal/overlord/state"
 	"github.com/canonical/workshop/internal/sdk"
 	"github.com/canonical/workshop/internal/workshop"
@@ -299,6 +300,20 @@ func workshopBaseOnly(change *state.Change, w string, age Age) (*workshop.Snapsh
 
 	snapshot := workshop.SdkSnapshot(format, image, nil)
 	return &snapshot, nil
+}
+
+func WorkshopPreservedConns(change *state.Change, w string) (map[string]schema.PreservedConn, error) {
+	var prev map[string]schema.PreservedConn
+	if err := change.Get(WorkshopPreservedConnsKey(w), &prev); errors.Is(err, state.ErrNoState) {
+		prev = nil
+	} else if err != nil {
+		return nil, fmt.Errorf("internal error: %q workshop previous connections not found (change ID: %s)", w, change.ID())
+	}
+	return prev, nil
+}
+
+func WorkshopPreservedConnsKey(w string) string {
+	return w + "_conns"
 }
 
 // SdkVolumeLastUsed returns the most recent Task of the given Change that
--- a/internal/overlord/ifacestate/ifacemgr.go
+++ b/internal/overlord/ifacestate/ifacemgr.go
@@ -60,6 +60,7 @@ func New(s *state.State, r *state.TaskRunner) *InterfaceManager {
 	r.AddHandler("disconnect", OnDo(m.doDisconnect), OnUndo(m.undoDisconnect))
 
 	r.AddHandler("discard-conns", m.doDiscard, OnUndo(m.undoDiscard))
+	r.AddHandler("restore-conns", OnDo(m.doRestoreConns), OnUndo(m.doDiscard))
 
 	r.AddHandler("setup-profiles", OnDo(m.doSetupProfiles), OnUndo(m.undoSetupProfiles))
 	r.AddHandler("remove-profiles", OnDo(m.doRemoveProfiles), nil)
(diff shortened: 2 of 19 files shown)


jonathan-conder:
@copilot fix the linter issues within this PR (i.e. use maps.Copy)

jonathan-conder:
@copilot also add some test coverage while you're at it

jonathan-conder:
@copilot+claude-haiku-4.5 in future please don't remove the disclaimer I deliberately put in the description...

dmitry-lyfar:
We should update this too: https://canonical-workshopubuntucom--939.com.readthedocs.build/explanation/workshops/concepts/#connections-across-refresh-and-restore

internal/overlord/handlersetup/handler_setup.go
```diff
@@ -301,6 +302,20 @@ func workshopBaseOnly(change *state.Change, w string, age Age) (*workshop.Snapsh
...
 
+func WorkshopPreservedConns(change *state.Change, w string) (map[string]schema.PreservedConn, error) {
+	var prev map[string]schema.PreservedConn
+	if err := change.Get(WorkshopPreservedConnsKey(w), &prev); errors.Is(err, state.ErrNoState) {
+		prev = nil
+	} else if err != nil {
+		return nil, fmt.Errorf("internal error: %q workshop previous connections not found (change ID: %s)", w, change.ID())
+	}
+	return prev, nil
+}
```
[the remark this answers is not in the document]
jonathan-conder: this basically never happens, I'd rather stick with consistency here

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -201,6 +201,8 @@ func (m *InterfaceManager) ensureBackendInit() error {
 				continue
 			}
```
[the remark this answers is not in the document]
jonathan-conder: we can do it if it becomes useful

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -60,6 +60,7 @@ func New(s *state.State, r *state.TaskRunner) *InterfaceManager {
 	r.AddHandler("disconnect", OnDo(m.doDisconnect), OnUndo(m.undoDisconnect))
 
 	r.AddHandler("discard-conns", m.doDiscard, OnUndo(m.undoDiscard))
```
dmitry-lyfar, line 62: it's interesting that m.doDiscard is not wrapped in `OnDo` here, I could not see any issue with doing so, do you understand why not?

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -195,27 +195,22 @@ func (m *InterfaceManager) ensureBackendInit() error {
 			pctx := context.WithValue(ctx, workshop.ContextProjectId, project.ProjectId)
 			workshops, err := m.backend.ProjectWorkshops(pctx)
 			if err != nil {
-				logger.Noticef("Cannot load workshops from %q: %v", project.Path, err)
-				continue
+				return fmt.Errorf("cannot load workshops from %q: %v", project.Path, err)
```
dmitry-lyfar, line 200: This should work given we have socket activation. Otherwise, if `ensureBackendInit` breaks half-way, the repo remains with a few SDKs already added which would fail on a retry. Perhaps, worth a comment now.

dmitry-lyfar (approved):
I did additional manual testing for scenarios using `workshop restore` and `workshop sketch-sdk` that fails on refresh to ensure the manual connections are preserved. I suppose you only extended the binding spread tests assuming that the API tests provide better coverage for undesired / manual connections?

dmitry-lyfar:
@akcano please have a look into this one and if you have notes, send them as suggestions, so we can pull this change for the upcoming release (jc is off this week).

akcano:
@dmitry-lyfar sure, will do.

dmitry-lyfar (approved):

internal/overlord/ifacestate/ifacemgr.go
jonathan-conder: I wondered that as well, was meaning to ask you about it

internal/overlord/ifacestate/ifacemgr.go
jonathan-conder: yeah good point, I think we should fix it properly

PR #818, opened 2026-06-03
Add custom device interface

jonathan-conder:
# Description

Adds the `custom-device` interface, which is like the `camera` interface but allows SDK authors and Workshop users to specify a class of devices to expose inside the workshop. Currently the class is decided by the `subsystem` property, but this may expand in future.

All `custom-device` plugs connect to the unique `custom-device` slot in the system SDK. Only manual connections are allowed, for security reasons. Shell completion works well but you typically do have to spell out the slot in `workshop connect`.

The spread tests use Python's `uinput` library to create a fake input device and add it to a workshop. Have to credit copilot for its creativity here.

Now that we have a global refresh cache, adding the `custom-device` slot requires a `system` revision bump. I tested refreshing `workshop` to `edge/818` and was able to `--continue` an existing launch and `restore` a workshop still using `system-1`. This won't be possible once the `system-1` volume is cleaned up, but at that point there should be no changes in progress and no workshops that can be restored to revision `1`.

I think if `launch --wait-on-error` fails very early on, before `retrieve-sdk`, and the user doesn't have a `system-1` volume yet, then `launch --continue` might fail after refreshing the snap. But I don't think it's a significant issue.

Also removes a few places which use backticks for no good reason, since copilot was being a bit overzealous about it.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

* [x] If needed, `docs/.coverage.yaml` updated, coverage tags added (`.. artefact`).

---

Or:

### DIFF
--- a/docs/explanation/interfaces/custom-device-interface.rst
+++ b/docs/explanation/interfaces/custom-device-interface.rst
@@ -1,0 +1,144 @@
+.. _exp_custom_device_interface:
+
+.. meta::
+   :description: Documentation on the custom-device interface that enables
+                 workshops to access arbitrary host devices belonging to a
+                 given subsystem, for hardware testing, device development, and
+                 other applications that need access to non-standard devices.
+
+Custom device interface
+=======================
+
+.. @artefact custom-device interface
+
+The custom device interface
+enables access to arbitrary host devices inside the workshop,
+identified by the device *subsystem* they belong to
+(for example, :samp:`input`, :samp:`tty`, or :samp:`usb`).
+
+By using the interface,
+the SDK publisher allows the workshop to access devices
+that no dedicated interface covers,
+such as serial adapters, input devices, or other peripherals
+used for testing hardware or embedded devices.
+
+.. _exp_custom_device_plug:
+
+Custom device interface plug
+----------------------------
+
+An essential element here is the custom device interface plug,
+which is declared in the SDK definition.
+
+Its structure includes the name of the plug,
+the interface (:samp:`custom-device`),
+and a :samp:`subsystem` attribute.
+
+Defining the plug in an SDK
+allows the workshops using this SDK to connect to matching devices,
+which can unlock additional SDK functionality.
+
+
+.. _exp_determine_subsystem:
+
+Device subsystems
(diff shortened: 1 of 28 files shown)


docs/reference/definition-files/workshop-definition.rst
```diff
@@ -240,6 +240,34 @@ They have no attributes.
 The only camera interface slot is :samp:`system:camera`.
 
 
+Custom device interface
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
```suggestion
Custom device interface
~~~~~~~~~~~~~~~~~~~~~~~
```

docs/explanation/interfaces/custom-device-interface.rst
```diff
@@ -0,0 +1,144 @@
...
+Its structure includes the name of the plug,
+the interface (:samp:`custom-device`),
+and a :samp:`subsystem` attribute.
+
+Defining the plug in an SDK
+allows the workshops using this SDK to connect to matching devices,
+which can unlock additional SDK functionality.
+
+
+.. _exp_determine_subsystem:
```
akcano:
```suggestion
.. _exp_custom_device_subsystem:
```

For consistency across the entire file.

docs/reference/sdks.rst
```diff
@@ -181,6 +182,32 @@ as video capture devices.
...
 
 
+.. _ref_custom_device_interface:
+
+Custom device interface
+~~~~~~~~~~~~~~~~~~~~~~~
+
+.. @artefact custom-device interface
+
+A custom-device plug in the definition must specify the plug name, the interface, and a ``subsystem`` attribute:
```
```suggestion
A custom-device plug in the definition
must specify the plug name, the interface,
and a :samp:`subsystem` attribute:
```

Semantic roles and semantic line breaks where possible. Global.

akcano (changes requested):
LGTM with minor remarks.

dmitry-lyfar (approved):

akcano (approved):

PR #820, opened 2026-06-03
Doc: Expand plugs and slots exp, how-tos

akcano:


akcano:
Decoupled into #875, #876, #877.

PR #823, opened 2026-06-04
Fix exec summaries to show the requested command

tlm:
## Summary

Fix `workshop exec` change and task summaries so they show the user-requested command instead of Workshop’s internal `sudo` execution wrapper.

Previously, a command such as:

```sh
workshop exec -I dev -- /bin/echo hello
```

could show summaries like:

```text
Execute command "sudo"
Exec command "sudo"
```

This happened because the CLI prepended the `sudo`/login-shell wrapper directly to the command array sent to `workshopd`, and the daemon builds summaries from the first command element.

This change separates the user command from Workshop’s execution wrapper:

- `command`: the user-requested command, used for summaries and diagnostics
- `command-prefix`: Workshop-managed wrapper, applied only at the backend execution boundary

## Why not just remove the wrapper?

During investigation, we briefly tested removing the `sudo`/`bash -l` wrapper entirely. That fixed summaries, but it changed important runtime behaviour.

Without the wrapper, supplementary groups were lost. For example, `id -Gn` changed from the expected full group list:

```text
workshop adm cdrom sudo audio dip video plugdev lxd netdev render
```

to only:

```text
workshop
```

The login-shell environment was also lost. `PATH` changed from the expected workshop/login-shell path, including user and SDK paths:

```text
/home/workshop/.local/bin:/home/workshop/go/bin:/var/lib/workshop/sdk/go/bin:...
```

to a more minimal/default path:

```text
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
```

Further isolation showed:

- `bash -l` restores the login-shell `PATH`
- `sudo` restores supplementary groups
- LXD exec `User` and `Group` fields only set primary UID/GID
- LXD exec does not expose supplementary/init groups

So the wrapper is still required for correct runtime behaviour. The fix therefore keeps the wrapper, but stops treating it as the user command.

## Changes

- Add `command-prefix` support to exec API payloads.
- Keep `command` as the user-requested command for summaries and diagnostics.
- Add `CommandPrefix` and `EffectiveCommand()` to `workshop.ExecArgs`.
- Apply `CommandPrefix + Command` only at the LXD execution boundary.
- Update `workshop exec` to send the existing `sudo`/login-shell wrapper as `CommandPrefix`.
- Preserve existing runtime behaviour:
  - supplementary groups via `sudo`
  - login-shell environment/PATH via `bash -l`
  - preserved environment variables
- Sort environment keys when building the prefix so payload construction is deterministic.
- Add unit coverage for:
  - client payload encoding
  - empty/omitted command prefix handling
  - daemon summary generation
  - effective command composition
  - CLI request encoding
- Add Spread coverage for user-facing exec summaries.

## Behaviour after the fix

The API payload now keeps the requested command separate from the wrapper. For example:

```json
{
  "command": ["/bin/echo", "hello"],
  "command-prefix": [
    "sudo",
    "-u", "#1000",
    "-g", "#1000",
    "--preserve-env=...",
    "--",
    "bash",
    "-l",
    "-c",
    "exec -- \"$0\" \"$@\""
  ]
}
```

The daemon continues to build summaries from `command[0]`, so summaries now report the requested command:

```text
Execute command "/bin/echo"
Exec command "/bin/echo"
```

The LXD backend executes the effective command:

```text
command-prefix + command
```

so runtime behaviour remains unchanged.

## Validation

Passed:

```sh
go test ./cmd/workshop ./client ./internal/daemon ./internal/workshop
go test ./...
git --no-pager diff --check
git --no-pager diff --cached --check
```

Live QA passed locally after rebuilding and restarting `workshopd`:

```sh
workshop exec -I dev -- /bin/echo command-prefix-summary-check
workshop exec -I dev -- id -Gn
workshop exec -I dev -- printenv PATH
workshop exec -I --env WORKSHOP_QA_ENV=present dev -- printenv WORKSHOP_QA_ENV
```

Confirmed:

- summaries show the user command:
  ```text
  Execute command "/bin/echo"
  Exec command "/bin/echo"
  ```

- supplementary groups are preserved:
  ```text
  workshop adm cdrom sudo audio dip video plugdev lxd netdev render
  ```

- login-shell `PATH` is preserved:
  ```text
  /home/workshop/.local/bin:/home/workshop/go/bin:/var/lib/workshop/sdk/go/bin:...
  ```

- custom environment variables are preserved:
  ```text
  present
  ```

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -482,6 +482,10 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 	if args.action {
 		logger.Debugf("Running action %q", av)
 	} else {
+		// Obtain an exec session as close to a real login shell as possible.
+		// This is required as an lxd exec call is a simple namespace exec, and LXD
+		// ignores the instance configuration by design:
+		// https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#user-groups-and-working-directory
 		commandPrefix = []string{
 			"sudo",
 			"-u",
@@ -503,7 +507,7 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 			"-c",
 			`exec -- "$0" "$@"`,
 		)
-		logger.Debugf("Running command %q with prefix %q", av, commandPrefix)
+		logger.Debugf("Running command %q", av)
 	}
 
 	// Record terminal state (and restore it before we exit).
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -478,32 +478,32 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 		env[key] = value
 	}
 
+	var commandPrefix []string
 	if args.action {
 		logger.Debugf("Running action %q", av)
 	} else {
-		// Obtain an exec session as close to a real login shell as possible.
-		// This is required as an lxd exec call is a simple namespace exec, and LXD
-		// ignores the instance configuration by design:
-		// https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#user-groups-and-working-directory
-		keys := slices.Sorted(maps.Keys(env))
-		command := []string{"sudo",
+		commandPrefix = []string{
+			"sudo",
 			"-u",
 			"#" + strconv.Itoa(flags.UserId),
 			"-g",
 			"#" + strconv.Itoa(flags.GroupId),
 		}
+
+		keys := slices.Sorted(maps.Keys(env))
 		for _, k := range keys {
-			command = append(command, "--preserve-env="+k)
+			commandPrefix = append(commandPrefix, "--preserve-env="+k)
 		}
-		command = append(command,
+
+		commandPrefix = append(
+			commandPrefix,
 			"--",
 			"bash",
 			"-l",
 			"-c",
 			`exec -- "$0" "$@"`,
 		)
-		av = append(command, av...)
-		logger.Debugf("Running %q", av)
+		logger.Debugf("Running command %q with prefix %q", av, commandPrefix)
 	}
 
 	// Record terminal state (and restore it before we exit).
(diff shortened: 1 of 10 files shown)


cmd/workshop/exec.go
```diff
@@ -478,32 +478,32 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
...
 	}
 
+	var commandPrefix []string
 	if args.action {
 		logger.Debugf("Running action %q", av)
 	} else {
-		// Obtain an exec session as close to a real login shell as possible.
-		// This is required as an lxd exec call is a simple namespace exec, and LXD
-		// ignores the instance configuration by design:
-		// https://documentation.ubuntu.com/lxd/en/latest/instance-exec/#user-groups-and-working-directory
```
jonathan-conder, line 487: think we can preserve the comment, it's still relevant

cmd/workshop/exec.go
```diff
@@ -478,32 +478,32 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
...
+			commandPrefix,
 			"--",
 			"bash",
 			"-l",
 			"-c",
 			`exec -- "$0" "$@"`,
 		)
-		av = append(command, av...)
-		logger.Debugf("Running %q", av)
+		logger.Debugf("Running command %q with prefix %q", av, commandPrefix)
```
jonathan-conder: I don't see a need to surface the prefix details to the user. Not a big deal though

cmd/workshop/exec.go
```diff
@@ -478,32 +478,32 @@ func exec(root *CmdRoot, flags *ExecFlags, args *ExecArgs) error {
 		env[key] = value
 	}
 
+	var commandPrefix []string
 	if args.action {
 		logger.Debugf("Running action %q", av)
```
jonathan-conder, line 483: I was going to suggest sending the prefix for actions as well, for consistency. But it looks like it would require adding a `CommandPrefix` a few levels further down. We can probably leave it as-is for now.

cmd/workshop/exec_test.go
```diff
@@ -110,6 +113,73 @@ foo:
...
+		case fmt.Sprintf("/v1/projects/%s/workshops", m.prjId):
+			c.Check(r.Method, check.Equals, "GET")
+			w.WriteHeader(200)
+			fmt.Fprintln(w, mockSingleWorkshopSpecifyStatus("Ready"))
+		case fmt.Sprintf("/v1/projects/%s/workshops/ws/exec", m.prjId):
+			c.Check(r.Method, check.Equals, "POST")
+			body := DecodedRequestBody(c, r)
+			userId := fmt.Sprint(workshop.Uid)
+			c.Check(body["command"], check.DeepEquals, []any{"echo", "foo"})
+			commandPrefix, ok := body["command-prefix"].([]any)
```
jonathan-conder: why not use DeepEquals here too?

cmd/workshop/exec_test.go
```diff
@@ -110,6 +113,73 @@ foo:
...
+			)
+			c.Check(
+				commandPrefix[len(commandPrefix)-5:],
+				check.DeepEquals,
+				[]any{"--", "bash", "-l", "-c", `exec -- "$0" "$@"`},
+			)
+			c.Check(body["user-id"], check.Equals, json.Number(userId))
+			c.Check(body["group-id"], check.Equals, json.Number(fmt.Sprint(workshop.Gid)))
+
+			environment, ok := body["environment"].(map[string]any)
```
jonathan-conder: and here

jonathan-conder (approved):
A few minor comments but makes sense to me

cmd/workshop/exec.go
tlm: agreed. Done & Done

cmd/workshop/exec.go
tlm: agreed. Done & Done

cmd/workshop/exec.go
tlm: yeah this would be good. I'll grab it in a follow up PR to keep this one focused and small enough.

cmd/workshop/exec_test.go
tlm: 100% right. Done & Done

cmd/workshop/exec_test.go
tlm: Done & Done

PR #826, opened 2026-06-04
Retry transient HTTP network timeouts

### LINKED ISSUE #825: Retry transient network timeouts during SDK downloads
## Summary

SDK downloads can occasionally fail in GitHub-hosted Actions runners with a transient TLS handshake timeout.

I have seen this happen around three times while using Workshop from GitHub Actions via `canonical/launch-workshop@v0`:

```yaml
- name: Workshop Checkout
  uses: canonical/launch-workshop@v0
  with:
    token: ${{ secrets.WORKSHOP_TOKEN }}
    workshop: rkc-dev
```

The failure happens during `workshop launch`, while retrieving an SDK from the SDK Store:

```text
Retrieve "nim-toolchain" SDK from channel "latest/stable"

2026-06-03T22:13:15Z ERROR Get "https://canonical-bos01.cdn.snapcraftcontent.com/download-origin/canonical-lgw01/4NA7qseGJRuMJBKuGckgtEL64WqLOVkY_5.sdk?token=<TOKEN>": net/http: TLS handshake timeout
```

This is not a functional correctness issue with SDK retrieval itself. The goal is to improve CI reliability when Workshop is used from GitHub-hosted Actions runners, where transient network issues can occasionally happen.

## What I observed

The failed task was:

```text
Error    20.457s  Retrieve "nim-toolchain" SDK from channel "latest/stable"
```

The failure happens during the TLS handshake, before the SDK response body is downloaded. The SDK file is very small, so this does not appear to be related to download size.

I have not seen this locally so far, but I have hit it multiple times on GitHub-hosted Actions runners.

## Expected behavior

A transient network timeout during an SDK download should be retried when it is safe to do so.

For example, a TLS handshake timeout on a `GET` request should not immediately fail the whole `workshop launch` if the existing retry policy could recover on a later attempt.

## Actual behavior

When the SDK download request fails before an HTTP response is returned, the error is returned immediately and the `retrieve-sdk` task fails.

This can make GitHub Actions runs flaky:

```text
- Retrieve "nim-toolchain" SDK from channel "latest/stable" (... net/http: TLS handshake timeout)
"rkc-dev" launch aborted
```

## Notes on retry behavior

Workshop already retries selected HTTP status codes. This case is different because a TLS handshake timeout is a transport-level error: no HTTP response is available.

A conservative approach would be to retry only transient timeout errors for safe HTTP methods, such as `GET`, while keeping the existing timeout values unchanged.

I think it is also important not to retry explicit request cancellation or context deadline expiry, and to preserve the original timeout error if all retry attempts are exhausted.

## Proposed direction

Extend the existing HTTP retry middleware so that transport-level timeout errors are retryable for safe HTTP methods.

This would let SDK downloads use the existing retry policy when a transient TLS handshake timeout happens, without changing the timeout duration or retrying requests that may have side effects.

I tested this direction locally with unit tests for:

* retrying a network timeout for a safe method such as `GET`
* not retrying a network timeout for an unsafe method such as `POST`
* preserving the original timeout error when retry attempts are exhausted

I also tested the normal SDK download path with a locally built `workshopd` running in debug mode. The SDK download path still works as expected:

```text
DEBUG: request (method: "GET", host: "api.pkg.store", path: "/api/v1/sdks/download/YHf58kLHgbBfgip240HbCwuvHyll69Ot_6.sdk", status: "302 FOUND", duration: 354.468219ms)
DEBUG: request (method: "GET", host: "canonical-bos01.cdn.snapcraftcontent.com", path: "/download-origin/canonical-lgw01/YHf58kLHgbBfgip240HbCwuvHyll69Ot_6.sdk", status: "200 OK", duration: 1.676454592s)
```

## Tests run

```text
gofmt -w internal/https/middleware.go internal/https/middleware_test.go internal/sdkstore/http.go
go test ./internal/https ./internal/sdkstore
go test ./...
golangci-lint run
```

shizuku198411:
Fixes: #825

# Description
This PR improves HTTP retry handling for transient network timeout errors.

In GitHub-hosted Actions runners, SDK downloads from the SDK Store can occasionally fail with a TLS handshake timeout before an HTTP response is returned. Previously, those transport-level timeout errors were returned immediately and did not use the existing retry policy.

This change makes the HTTP retry middleware retry transient network timeout errors for safe HTTP methods, such as `GET`. It keeps the existing timeout values unchanged and does not retry requests that may have side effects.

The original timeout error is preserved if all retry attempts are exhausted.

Tests were added for:

* retrying a network timeout for a safe method;
* not retrying a network timeout for an unsafe method;
* preserving the original timeout error after retry exhaustion.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/https/middleware.go
+++ b/internal/https/middleware.go
@@ -230,9 +230,9 @@ func (m retryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 			if err != nil {
 				if retryable {
 					// Convert retryable transport errors into retryableErr so retry.Call
-					// treats them the same way as retryable HTTP status response. Keep the
+					// treats them the same way as retryable HTTP status responses. Keep the
 					// original error wrapped so the final error still explains what failed
-					// it all attempts are exhausted.
+					// if all attempts are exhausted.
 					return retryableErr{err: err}
 				}
 				return err
--- a/internal/https/middleware.go
+++ b/internal/https/middleware.go
@@ -250,6 +250,13 @@ func (m retryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 		Attempts: m.policy.Attempts,
 		Delay:    m.policy.Delay,
 		BackoffFunc: func(delay time.Duration, attempts int) time.Duration {
+			if res == nil {
+				// Transport errors do not have an HTTP response, so there is
+				// no Retry-After header to inspect. Fall back to the policy backoff.
+				var duration time.Duration
+				duration, backOffErr = m.clampBackoff(delay)
+				return duration
+			}
 			var duration time.Duration
 			duration, backOffErr = m.defaultBackoff(res, delay)
 			return duration
@@ -320,12 +327,6 @@ func isSafeMethod(method string) bool {
 //   - Retry-After: <http-date>
 //   - Retry-After: <delay-seconds>
 func (m retryMiddleware) defaultBackoff(resp *http.Response, backoff time.Duration) (time.Duration, error) {
-	if resp == nil {
-		// Transport errors do not have an HTTP response, so there is no
-		// Retry-After header to inspect. Fall back to the policy backoff.
-		return m.clampBackoff(backoff)
-	}
-
 	if header := resp.Header.Get("Retry-After"); header != "" {
 		// Attempt to parse the header from the request.
 		//
--- a/internal/https/middleware.go
+++ b/internal/https/middleware.go
@@ -154,7 +154,8 @@ func (lr roundTripRecorder) RoundTrip(req *http.Request) (*http.Response, error)
 }
 
 // RetryMiddleware allows retrying of certain retryable http errors.
-// This only handles very specific status codes, ones that are deemed retryable:
+// This only handles transient network timeout errors for safe requests
+// and specific status codes that are deemed retryable:
 //
 //   - 502 Bad Gateway
 //   - 503 Service Unavailable
@@ -191,10 +192,19 @@ func makeRetryMiddleware(transport http.RoundTripper, policy RetryPolicy, clock 
 	}
 }
 
-type retryableErr struct{}
+type retryableErr struct {
+	err error
+}
 
-func (retryableErr) Error() string {
+func (e retryableErr) Error() string {
+	if e.err != nil {
+		return e.err.Error()
+	}
 	return "retryable error"
+}
+
+func (e retryableErr) Unwrap() error {
+	return e.err
 }
 
 // RoundTrip defines a strategy for handling retries based on the status code.
@@ -215,8 +225,16 @@ func (m retryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
 
 			var retryable bool
 			var err error
+			res = nil
 			res, retryable, err = m.roundTrip(req) //nolint:bodyclose // Body closed by caller.
 			if err != nil {
+				if retryable {
+					// Convert retryable transport errors into retryableErr so retry.Call
+					// treats them the same way as retryable HTTP status response. Keep the
+					// original error wrapped so the final error still explains what failed
+					// it all attempts are exhausted.
+					return retryableErr{err: err}
+				}
 				return err
 			}
 			if retryable {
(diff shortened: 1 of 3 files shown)


internal/https/middleware.go
```diff
@@ -269,6 +320,12 @@ func (m retryMiddleware) roundTrip(req *http.Request) (*http.Response, bool, err
 //   - Retry-After: <http-date>
 //   - Retry-After: <delay-seconds>
 func (m retryMiddleware) defaultBackoff(resp *http.Response, backoff time.Duration) (time.Duration, error) {
+	if resp == nil {
+		// Transport errors do not have an HTTP response, so there is no
+		// Retry-After header to inspect. Fall back to the policy backoff.
+		return m.clampBackoff(backoff)
+	}
+
```
jonathan-conder: I think this makes more sense in the anonymous `BackoffFunc` rather than here

jonathan-conder (approved):
Looks good, but I must admit to not fully understanding this part of the codebase.

We share this code (mostly) with Juju: https://github.com/juju/juju/blob/main/internal/http/middleware.go. Would you like to raise a similar PR there?

internal/https/middleware.go
```diff
@@ -215,8 +225,16 @@ func (m retryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
 			var err error
+			res = nil
 			res, retryable, err = m.roundTrip(req) //nolint:bodyclose // Body closed by caller.
 			if err != nil {
+				if retryable {
+					// Convert retryable transport errors into retryableErr so retry.Call
+					// treats them the same way as retryable HTTP status response. Keep the
+					// original error wrapped so the final error still explains what failed
+					// it all attempts are exhausted.
+					return retryableErr{err: err}
```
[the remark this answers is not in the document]
jonathan-conder: this sounds reasonable but out of scope for the current PR

shizuku198411:
@jonathan-conder 
Thanks, that makes sense. I didn’t realize this middleware was shared with Juju, so thank you for pointing that out.

I haven’t looked deeply enough into the Juju codebase or its design conventions yet, so I’d like to spend a bit more time understanding that before opening a PR there. It may take me a little while, but I’m happy to look into it.

internal/https/middleware.go
shizuku198411: Good point, that makes sense! I’ll move the `resp == nil` handling into the `BackoffFunc`, since that case only happens for transport-level errors where no HTTP response exists. That keeps `defaultBackoff` focused on response-based backoff handling. I'll fix it

internal/https/middleware.go
```diff
@@ -215,8 +225,16 @@ func (m retryMiddleware) RoundTrip(req *http.Request) (*http.Response, error) {
...
 			var retryable bool
 			var err error
+			res = nil
 			res, retryable, err = m.roundTrip(req) //nolint:bodyclose // Body closed by caller.
 			if err != nil {
+				if retryable {
+					// Convert retryable transport errors into retryableErr so retry.Call
+					// treats them the same way as retryable HTTP status response. Keep the
+					// original error wrapped so the final error still explains what failed
+					// it all attempts are exhausted.
```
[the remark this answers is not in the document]
shizuku198411: Sorry It was my mistake. I’ll fix the typo and adjust the wording.

internal/https/middleware.go
```diff
@@ -191,12 +192,21 @@ func makeRetryMiddleware(transport http.RoundTripper, policy RetryPolicy, clock
...
 
-func (retryableErr) Error() string {
+func (e retryableErr) Error() string {
+	if e.err != nil {
+		return e.err.Error()
+	}
 	return "retryable error"
 }
 
+func (e retryableErr) Unwrap() error {
```
tlm, line 206: Thank you!

internal/https/middleware.go
shizuku198411: @jonathan-conder 
I moved the `resp == nil` handling into `BackoffFunc`.

I kept the separate `var duration time.Duration` assignment to avoid any accidental scoping/shadowing issue with `backOffErr`, but if that feels unnecessarily verbose I’m happy to simplify it.

dmitry-lyfar:
Nicely done, @shizuku198411. I will ensure the tests pass and pull this change in this bi-weekly release.

shizuku198411:
Thank you @dmitry-lyfar! I’m glad I could contribute :)

PR #837, opened 2026-06-09
Fix reading nil-change on panic

dmitry-lyfar:


jonathan-conder:
I did my own one here: https://github.com/canonical/workshop/pull/838

In most cases we can still show the new message. And I added a regression test

PR #847, opened 2026-06-11
handle hook Stat errors

### LINKED ISSUE #846: Malformed project SDK hook path causes workshopd panic loop
## Summary

While manually creating a project-local SDK, I accidentally created the `hooks` path as a regular file instead of an empty directory:
```bash
touch .workshop/bad-sdk/sdk/hooks
```
** This is entirely my fault.

The resulting project-local SDK layout looked like this:

```text
.
├── [drwxr-xr-x]  .workshop
│   └── [drwxrwxr-x]  bad-sdk
│       ├── [drwxrwxr-x]  sdk
│       │   └── [-rw-rw-r--]  hooks  # regular file, expected to be a directory
│       └── [-rw-rw-r--]  sdk.yaml
└── [-rw-rw-r--]  workshop.yaml
```

After running `workshop launch`, `workshopd` panicked while checking for the SDK hook path.

The more problematic part is that once this happens, `workshopd` keeps panicking on restart until the invalid SDK file is manually removed from the installed SDK copy. As a result, all normal `workshop` commands fail because the daemon socket is unavailable.

## What I observed

With the malformed project-local SDK layout, `workshop launch panic-test --verbose` progressed until installing the project SDK:

```text
✓ Retrieve "system" SDK
✓ Create "panic-test" storage directories
✓ Create new "panic-test" workshop from "system" snapshot
✓ Install "system" SDK
✓ Start "panic-test" workshop
✓ Install "bad-sdk" SDK
error: cannot communicate with workshopd: Get "http://localhost/v1/changes/103?verbose=true": dial unix /var/snap/workshop/common/workshop/workshop.socket: connect: connection refused
"panic-test" launch aborted
```

After that, even `workshop list` failed because `workshopd` was no longer reachable:

```text
error: cannot communicate with workshopd: Post "http://localhost/v1/projects": dial unix /var/snap/workshop/common/workshop/workshop.socket: connect: connection refused
```

The daemon log showed a nil pointer dereference in `hookstate.executeHook`:

```text
panic: runtime error: invalid memory address or nil pointer dereference
github.com/canonical/workshop/internal/overlord/hookstate.(*HookManager).executeHook(...)
        /root/parts/workshop/build/internal/overlord/hookstate/handlers.go:230 +0x17f
github.com/canonical/workshop/internal/overlord/hookstate.(*HookManager).doRunHook(...)
        /root/parts/workshop/build/internal/overlord/hookstate/handlers.go:171 +0x1572
```

Restarting the daemon did not recover it. The service immediately panicked again and eventually hit systemd's restart limit:

```text
snap.workshop.workshopd.service: Scheduled restart job, restart counter is at 5.
snap.workshop.workshopd.service: Start request repeated too quickly.
snap.workshop.workshopd.service: Failed with result 'exit-code'.
```

It looks like the pending hook task is resumed when `workshopd` starts again, so the same malformed SDK path is inspected again and the daemon crashes repeatedly.

I was able to recover manually by finding the installed project SDK mount source via LXD, removing the invalid `sdk/hooks` file from the installed SDK copy, and then restarting `workshopd`.

## Expected behavior

A malformed project-local SDK should cause a normal validation or task error, not a daemon panic.

In this case, if `sdk/hooks` is a regular file instead of a directory, `workshopd` should report an error such as:

```text
cannot inspect "setup-base" hook for "bad-sdk" SDK: stat .../sdk/hooks/setup-base: not a directory
```

The daemon should remain running, and other `workshop` commands should continue to work.

In particular, `workshopd` should not enter a crash loop where every restart re-runs the pending hook task and panics again. Once the daemon reaches that state, normal CLI commands cannot be used to inspect or recover the workshop state, so manual intervention in the installed SDK directory is required.

## Proposed direction

The issue appears to be in the hook existence check.

When checking whether a hook exists, `executeHook()` calls `Stat()` on a path such as:

```text
/var/lib/workshop/sdk/bad-sdk/sdk/hooks/setup-base
```

If `sdk/hooks` is a regular file, this `Stat()` call returns an error such as `ENOTDIR`. The current code appears to handle `os.ErrNotExist`, but then still dereferences `info` for other errors. In this case `info` can be nil, causing the panic.

A possible direction is to handle the `Stat()` error before using `info`:

```go
info, err := wsFs.Stat(hookPath)
wsFs.Close()

if errors.Is(err, os.ErrNotExist) {
    logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
    return nil
}
if err != nil {
    return fmt.Errorf("cannot inspect %q hook for %q SDK: %w", hook.Type(), hook.Sdk, err)
}
if !info.Mode().IsRegular() {
    logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
    return nil
}
```

This keeps the existing behavior for missing hooks, but treats unexpected path errors such as `not a directory` as normal task errors instead of panicking.

After applying a local fix, the same malformed SDK no longer crashed `workshopd`. Instead, `workshop launch` failed normally with a task error and rolled back:

```text
- Run hook "setup-base" for "bad-sdk" SDK (cannot inspect "setup-base" hook for "bad-sdk" SDK: stat /var/lib/workshop/sdk/bad-sdk/sdk/hooks/setup-base: not a directory)
"panic-test" launch aborted
```

`workshop list` also continued to work afterwards:

```text
WORKSHOP    STATUS  NOTES
panic-test  Off     -
```

shizuku198411:
Fixes: #846

# Description
Fix a `workshopd` panic when inspecting malformed SDK hook paths.

If a project-local SDK has `sdk/hooks` as a regular file instead of a directory, hook inspection can return `not a directory`. Previously this could dereference a nil `FileInfo` and crash `workshopd`. This change returns a normal task error instead, while preserving the existing behavior for missing hooks.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/overlord/hookstate/handlers.go
+++ b/internal/overlord/hookstate/handlers.go
@@ -232,7 +232,7 @@ func (h *HookManager) executeHook(ctx context.Context, task *state.Task, w strin
 		return nil
 	}
 	if err != nil {
-		return fmt.Errorf("cannot inspect %q hook for %q SDK: %w", hook.Type(), hook.Sdk, err)
+		return fmt.Errorf("cannot read hook: %w", err)
 	}
 	if !info.Mode().IsRegular() {
 		logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
--- a/internal/overlord/hookstate/handlers.go
+++ b/internal/overlord/hookstate/handlers.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"slices"
 	"sync"
+	"syscall"
 
 	"github.com/canonical/x-go/strutil"
 	"gopkg.in/tomb.v2"
@@ -227,7 +228,7 @@ func (h *HookManager) executeHook(ctx context.Context, task *state.Task, w strin
 
 	info, err := wsFs.Stat(hookPath)
 	wsFs.Close()
-	if errors.Is(err, os.ErrNotExist) {
+	if errors.Is(err, os.ErrNotExist) || errors.Is(err, syscall.ENOTDIR) {
 		logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
 		return nil
 	}
--- a/internal/overlord/hookstate/handlers.go
+++ b/internal/overlord/hookstate/handlers.go
@@ -227,7 +227,14 @@ func (h *HookManager) executeHook(ctx context.Context, task *state.Task, w strin
 
 	info, err := wsFs.Stat(hookPath)
 	wsFs.Close()
-	if errors.Is(err, os.ErrNotExist) || !info.Mode().IsRegular() {
+	if errors.Is(err, os.ErrNotExist) {
+		logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("cannot inspect %q hook for %q SDK: %w", hook.Type(), hook.Sdk, err)
+	}
+	if !info.Mode().IsRegular() {
 		logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
 		return nil
 	}
--- a/internal/overlord/hookstate/handlers_test.go
+++ b/internal/overlord/hookstate/handlers_test.go
@@ -169,7 +169,7 @@ func (s *hookSuite) TestExecHookPathError(c *check.C) {
 	err := s.backend.LaunchOrRebuildWorkshop(s.ctx, wf, snapshot)
 	c.Check(err, check.IsNil)
 
-	ws, err := s.backend.WorkshopFs(s.ctx, "ws")
+	fs, err := s.backend.WorkshopFs(s.ctx, "ws")
 	c.Check(err, check.IsNil)
 	defer ws.Close()
 	err = ws.MkdirAll(filepath.Dir(sdk.SdkHooksDir("one")), 0755)
--- a/internal/overlord/hookstate/handlers_test.go
+++ b/internal/overlord/hookstate/handlers_test.go
@@ -151,6 +151,44 @@ func (s *hookSuite) TestExecHookDoesNotExist(c *check.C) {
 	c.Check(t1.Status(), check.Equals, state.DoneStatus)
 }
 
+func (s *hookSuite) TestExecHookPathError(c *check.C) {
+	s.state.Lock()
+	defer s.state.Unlock()
+	t1 := hookstate.Hook(s.state, "one", 0, hookstate.SetupBase)
+
+	chg := s.state.NewChange("sample", "...")
+	setWorkshopProject("ws", s.project, t1)
+	chg.Set("user", "testuser")
+	chg.AddTask(t1)
+
+	// Launch a workshop with a malformed SDK layout: hooks is a regular file
+	// instead of a directory. Stat() on sdk/hooks/setup-base then returns
+	// ENOTDIR and should be reported as a task error rather than panicking.
+	wf := &workshop.File{Name: "ws", Base: "ubuntu@20.04"}
+	snapshot := workshop.BaseOnly(sdk.R(1), wf.Base, "fakeimage123")
+	err := s.backend.LaunchOrRebuildWorkshop(s.ctx, wf, snapshot)
+	c.Check(err, check.IsNil)
+
+	ws, err := s.backend.WorkshopFs(s.ctx, "ws")
+	c.Check(err, check.IsNil)
+	defer ws.Close()
+	err = ws.MkdirAll(filepath.Dir(sdk.SdkHooksDir("one")), 0755)
+	c.Check(err, check.IsNil)
+	err = ws.WriteFile(sdk.SdkHooksDir("one"), []byte("not a directory"), 0644)
+	c.Check(err, check.IsNil)
+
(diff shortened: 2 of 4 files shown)


internal/overlord/hookstate/handlers_test.go
```diff
@@ -151,6 +151,44 @@ func (s *hookSuite) TestExecHookDoesNotExist(c *check.C) {
...
+	t1 := hookstate.Hook(s.state, "one", 0, hookstate.SetupBase)
+
+	chg := s.state.NewChange("sample", "...")
+	setWorkshopProject("ws", s.project, t1)
+	chg.Set("user", "testuser")
+	chg.AddTask(t1)
+
+	// Launch a workshop with a malformed SDK layout: hooks is a regular file
+	// instead of a directory. Stat() on sdk/hooks/setup-base then returns
+	// ENOTDIR and should be reported as a task error rather than panicking.
```
jonathan-conder: I would avoid mentioning the panic here, without the context of this issue it could be confusing. Leave historical context in commit messages.

internal/overlord/hookstate/handlers_test.go
```diff
@@ -151,6 +151,44 @@ func (s *hookSuite) TestExecHookDoesNotExist(c *check.C) {
...
+	wf := &workshop.File{Name: "ws", Base: "ubuntu@20.04"}
+	snapshot := workshop.BaseOnly(sdk.R(1), wf.Base, "fakeimage123")
+	err := s.backend.LaunchOrRebuildWorkshop(s.ctx, wf, snapshot)
+	c.Check(err, check.IsNil)
+
+	ws, err := s.backend.WorkshopFs(s.ctx, "ws")
```
jonathan-conder:
```suggestion
	fs, err := s.backend.WorkshopFs(s.ctx, "ws")
```
for us `ws` typically means `workshop`

internal/overlord/hookstate/handlers.go
```diff
@@ -227,7 +227,14 @@ func (h *HookManager) executeHook(ctx context.Context, task *state.Task, w strin
 
 	info, err := wsFs.Stat(hookPath)
 	wsFs.Close()
-	if errors.Is(err, os.ErrNotExist) || !info.Mode().IsRegular() {
+	if errors.Is(err, os.ErrNotExist) {
+		logger.Debugf("%q SDK does not provide %q hook", hook.Sdk, hook.Type())
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("cannot inspect %q hook for %q SDK: %w", hook.Type(), hook.Sdk, err)
```
jonathan-conder:
```suggestion
		return fmt.Errorf("cannot read hook: %w", err)
```
this appears in the logs for the task `Run hook "setup-base" for "mysdk" SDK`, so there's no need to repeat the hook type and SDK name.

internal/overlord/hookstate/handlers.go
```diff
@@ -227,7 +227,14 @@ func (h *HookManager) executeHook(ctx context.Context, task *state.Task, w strin
 
 	info, err := wsFs.Stat(hookPath)
 	wsFs.Close()
-	if errors.Is(err, os.ErrNotExist) || !info.Mode().IsRegular() {
+	if errors.Is(err, os.ErrNotExist) {
```
jonathan-conder: I think we should also check for your specific case here; it should not be an error to leave a stray file lying around.

jonathan-conder (commented):
Another nice find, and thanks for the fix! I have some comments about the wording

shizuku198411:
@jonathan-conder 
Thanks for the review!

This was triggered by a manually-created project-local SDK, so I’ll recommend `workshop sketch-sdk` to my team members for normal usage 😅

I’ll check your comments and update the commit.

jonathan-conder:
> This was triggered by a manually-created project-local SDK, so I’ll recommend `workshop sketch-sdk` to my team members for normal usage 😅

It's also perfectly fine to make in-project SDKs by hand, whatever works for you. Bugs like this should hopefully get more rare over time.

internal/overlord/hookstate/handlers.go
shizuku198411: Thanks, that makes sense.

I’ll try updating the implementation so this case is treated like a missing hook, allowing the launch to continue even if a stray file is left at `sdk/hooks`.

shizuku198411:
I updated the implementation in this direction.

For the original own case where `hooks` is not a directory, the resulting `ENOTDIR` is now treated the same as a missing hook. This lets `launch` continue instead of failing just because a stray `hooks` file exists.

I also kept the separate `err != nil` handling, since that was the root cause of the panic. Unexpected `Stat()` errors are now handled before using the returned `FileInfo`.

I added tests for both cases:

* `ENOTDIR` from a stray `hooks` file is skipped like a missing hook.
* Other `Stat()` errors are returned as task errors instead of panicking.

To make the non-directory case detectable consistently, I also added `not a directory` normalization in the SFTP filesystem layer, following the existing handling for similar path errors.

jonathan-conder (approved):
Thanks for the update. I didn't realise this would involve parsing the SFTP error messages, thanks for adding that too!

I think for `TestExecHookStatError` it might be simpler to set up an `EPERM` or something rather than mocking the FS. But it's OK as-is.

PR #866, opened 2026-06-18
Add friendlier domain names for workshops and integrate with systemd-resolved

jonathan-conder:
# Description

- Adds a systemd service to configure `systemd-resolved` to use LXD's DNS resolver for `*.wp` domains.
- Adds aliases `<WS>.<PID>.wp` and `<WS>.<BASENAME>.pid` for `<WS>-<PID>.wp`.
- Adds `<PID>.wp` as a search domain inside a workshop (i.e. workshops within a project can ping each other by name).
- Removes `-<PID>` from the local hostname within a workshop. The LXD instance name and the hostname sent over DHCP remain unchanged.
- Exposes preferred hostname in `workshop info`, with a note if the project directory's basename was not used.

### Fixes

- Ensures LXD unit tests are run once, not twice.
- Makes `remount` spread test more idempotent.

### TODO

- Ensure YAML template values are quoted if needed

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:


internal/workshop/lxd/lxd_backend.go
```diff
@@ -422,6 +429,71 @@ func (s *Backend) launchOrRebuildFromImage(conn lxd.InstanceServer, req api.Inst
...
+	}
+
+	modified := false
+	hostname := metadata.Templates["/etc/hostname"]
+	if format.N > 3 && hostname == nil {
+		hostname = &api.ImageMetadataTemplate{
+			When:     []string{"create", "copy"},
+			Template: "hostname.tpl",
+		}
+		metadata.Templates["/etc/hostname"] = hostname
```
dmitry-lyfar, line 460: Is it guaranteed to be allocated?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -343,33 +356,27 @@ func (s *Backend) LaunchOrRebuildWorkshop(ctx context.Context, file *workshop.Fi
...
 	fs, err := s.instanceFs(conn, req.Name)
 	if err != nil {
 		return err
 	}
 	defer fs.Close()
 
-	// Workaround https://github.com/canonical/lxd/issues/17983.
-	if err := fs.RemoveIfExists("/etc/systemd/system/graphical.target.wants/udisks2.service"); err != nil {
-		return err
+	if snapshot.Format.N > 3 {
```
dmitry-lyfar, line 369: hmm, so if we launch from a snapshot, that means a snapshot has already a backed preserve_hostname?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -343,33 +356,27 @@ func (s *Backend) LaunchOrRebuildWorkshop(ctx context.Context, file *workshop.Fi
...
+		// cloud-init seems to ignore preserve_hostname in user-data, we have
+		// to bake it into the image instead.
+		cfg := strings.NewReader("preserve_hostname: true\n")
+		if err := fs.AtomicWriteTo(cfg, "/etc/cloud/cloud.cfg.d/99_preserve_hostname.cfg", 0644); err != nil {
```
dmitry-lyfar, line 373: Maybe wrap this error into `cannot write preserve_hostname config: %w`.

snap/local/hooks/configure
```diff
@@ -14,3 +16,9 @@ if [ -d /usr/local/share ] && [ -w /usr/local/share ]; then
     install_completion sdk zsh /usr/local/share/zsh/site-functions/_sdk
     install_completion workshop zsh /usr/local/share/zsh/site-functions/_workshop
 fi
+
+if [ -d /etc/systemd/system ] && [ -w /etc/systemd/system ] && command -v systemctl &>/dev/null; then
+    install_file 0644 "$SNAP/snap/lib/workshop-configure-dns.service" /etc/systemd/system
+    systemctl daemon-reload
+    systemctl enable workshop-configure-dns.service
```
dmitry-lyfar: Should we still continue if these fail?

internal/workshop/lxd/lxd_backend.go
jonathan-conder: yes, unless it's an old snapshot (which is only for `launch --continue` and `restore`)

internal/workshop/lxd/lxd_backend.go
jonathan-conder: AtomicWriteTo returns a PathError

snap/local/hooks/configure
jonathan-conder: I suppose we could, but the daemon is unlikely to work if systemd is down. Enable just makes a symlink in `/etc` so is unlikely to fail.

internal/workshop/lxd/lxd_backend.go
jonathan-conder: Nice catch, I'll fix it

internal/dns/resolved.go
```diff
@@ -0,0 +1,128 @@
...
+	return &systemdResolved{}
+}
+
+// ConfigureDNS points the given interface's DNS servers at addrs and registers
+// domain as a routing-only search domain, replacing any previously configured
+// values. It mirrors "resolvectl dns <iface> <addr>..." followed by "resolvectl
+// domain <iface> ~<domain>". The interface's kernel index is resolved once for
+// both calls.
+func (r *systemdResolved) ConfigureDNS(ctx context.Context, iface string, addrs []netip.Addr, domain string) error {
+	index, err := r.interfaceIndex(iface)
```
dmitry-lyfar, line 54: I'm not sure how likely this one is, but given we use the retry loop before to protect against d-bus events arriving earlier, maybe this index lookup should also be in the retry logic to avoid situations when it returns with "cannot find interface" and fails before the retry loop.

jonathan-conder: It's only possible if you create the device and immediately delete it. It's not possible to fail otherwise. The retry thing is, IMO a bug in systemd-resolved (it doesn't even ask the kernel if the device exists, just assumes it doesn't because it doesn't already know about it)

dmitry-lyfar: OK, I had something like LXD gets a snap refreshed, that reconfigures  the bridges but then I realised it's a managed bridge, so unless someone fiddles with it manually...

dmitry-lyfar (approved):

internal/workshop/lxd/lxd_backend.go
```diff
@@ -1157,6 +1163,20 @@ runcmd:
 	return cfg, nil
 }
 
+func mustExecute(name, text string, dot any) string {
+	t, err := template.New(name).Parse(text)
+	if err != nil {
+		panic(err)
```
dmitry-lyfar, line 1169: Panicking not erroring?

jonathan-conder: if it doesn't parse it's a logic bug, not something users will observe

jonathan-conder: it's even built in https://pkg.go.dev/text/template#Must

dmitry-lyfar (approved):

PR #829, opened 2026-06-05
Explain blocking changes during stop

tlm:
# Description

This improves the `workshop stop` UX when the stop action is blocked by another
workshop change, especially a refresh that is paused on error.

Previously, if a user ran `workshop stop dev` while a refresh was paused on
error, the CLI only received a generic daemon error. The output did not explain
which change was blocking the stop.

Before:

```text
error: cannot stop "dev": waiting on error
```

After, a refresh paused on error reports a concise, specific reason:

```text
error: cannot stop "dev": refresh change is waiting on error
```

For other (non-refresh) change conflicts, the CLI now has enough context to
point the user at the blocking change:

```text
error: cannot stop "dev": change is in progress
To view details: "workshop tasks 30"
```

## Design

This change keeps the existing project layering:

- The daemon API layer translates typed domain errors into API error responses.
- The client translates API error values into typed client errors via
  `errors.As`.
- The CLI formats the user-facing message.

The notable changes:

- `WorkshopHealth` records the change driving the current status on
  `HealthState.Cause.ChangeRef` (id, kind and status). New query helpers
  `HealthState.HasStatusIn` and `HealthStateCause.HasChangeRef` make the status
  and cause easy to inspect.
- `healthstate.CheckWorkshopHealth` now owns the decision of why an operation
  cannot proceed. When the workshop is blocked by a change waiting on error it
  returns a `healthstate.ChangeInProgressError` carrying the conflicting
  change's id and kind plus the project and workshop; any other disallowed
  status returns the matching `ErrorWorkshopHealth*` sentinel. This keeps the
  mapping in one place instead of repeating a switch in every request handler,
  so `StopMany` (and the other operations) simply call `CheckWorkshopHealth`
  with their allowed statuses and wrap the result.
- The change details come from the health cause rather than a second
  `conflict.CheckChangeConflict` call.
- The daemon workshops handler recognises `healthstate.ChangeInProgressError`
  and returns a structured `change-conflict` API error built from its fields.

Because the mapping is centralised, every blocked workshop operation (stop,
start, exec, refresh, remount, …) now yields the structured `change-conflict`
response, not just stop. What remains stop-specific is the CLI's tailored
recovery message; the equivalent `start` CLI message is handled on a separate
branch (`wsp-960-start-error-output`).

Example API error value:

```json
{
  "kind": "change-conflict",
  "message": "workshop \"dev\" has \"refresh\" change in progress",
  "value": {
    "change-id": "29",
    "change-kind": "refresh",
[...]

PR #870, opened 2026-06-19
Report a clear error when aborting with no paused change

tlm:
# Description

This improves the output of `workshop refresh --abort` / `--continue` (and the
`launch` equivalents) when there is no paused change to act on. It is part of
the WSP-960 error-output work and follows the same layering as #829 and #831.

Previously, aborting or continuing when nothing was paused surfaced an awkward,
doubled-up daemon error:

```text
error: cannot refresh "dev": cannot abort: no wait in progress
```

After this change the CLI reports a concise, command-specific message built
from its own context:

```text
error: cannot abort: no refresh in progress
error: cannot abort: no launch in progress
```

The same applies to `--continue` (`cannot continue: no refresh in progress`).

## Design

A typed error flows from the overlord up to the CLI, with each layer owning its
own representation:

- `conflict.ResumeAfterWait` returns a typed `conflict.WaitingChangeError`
  carrying the resume `Mode` and a typed `WaitingChangeReason`. Two reasons are
  defined: `no-change` (nothing in progress) and `running` (a matching change
  exists but is not paused waiting on error).
- The daemon maps `WaitingChangeError` to a structured
  `no-waiting-change-in-progress` API error whose value carries the reason
  (`{ "reason": "no-change" }`). The message field is a non-authoritative
  fallback; clients branch on the reason rather than parsing it.
- The client decodes the value into `client.WaitingChangeError` via
  `errors.As`, mirroring `client.ChangeConflictError`.
- The `refresh` and `launch` commands render the user-facing message from their
  own context: the verb from the abort/continue flag and the operation label
  ("refresh" / "launch"). Keeping the wording on the CLI is what lets each
  command name itself rather than echoing a generic daemon string.

The kind string is `no-waiting-change-in-progress` on both ends, and the value
deliberately carries only the reason — the client already knows the verb and
operation from the request it sent.

## Example API error value

```json
{
  "kind": "no-waiting-change-in-progress",
  "message": "cannot abort: no waiting change in progress",
  "value": { "reason": "no-change" }
}
```

## QA

Unit tests run:

```sh
go test ./internal/overlord/conflict
go test ./internal/daemon -check.f TestRefreshNoRefreshInProgress
go test ./internal/daemon -check.f TestLaunchWorkshopNoRefreshInProgress
go test ./client
go test ./cmd/workshop -check.f TestRefreshAbortNoWaitingChange
go test ./cmd/workshop -check.f TestLaunchAbortNoWaitingChange
```

End-to-end coverage is added to the existing per-command abort spread tests
(`tests/main/refresh-abort`, `tests/main/launch-abort`): each asserts that an
abort with no paused change reports the concise message, exercising the real
daemon, client and CLI together.

## Not included / follow-ups

- The `running` reason is wired through all layers but is not yet exercised
  end-to-end (it requires an actively-running, non-paused change at abort time,
  which is racy to provoke deterministically).
- Routing a wrong-kind blocking change to a `change-conflict` response is a
  separate follow-up.

# Self-review quick check

 * [x] Divide large expressions into digestable and self-explanatory ones.

## Docs

### DIFF
--- a/cmd/workshop/launch.go
+++ b/cmd/workshop/launch.go
@@ -99,6 +99,19 @@ $ workshop launch`,
 	return cmd
 }
 
+// waitingChangeError renders the message for an abort or continue that could
+// not be applied to a launch.
+func (c *CmdLaunch) waitingChangeError(e client.WaitingChangeError) error {
+	verb := "abort"
+	if c.Continue {
+		verb = "continue"
+	}
+	if e.Reason == client.WaitingChangeRunning {
+		return fmt.Errorf("cannot %s: launch is in progress", verb)
+	}
+	return fmt.Errorf("cannot %s: no launch in progress", verb)
+}
+
 func (c *CmdLaunch) Run(cmd *cobra.Command, av []string) error {
 	av = strutil.Deduplicate(av)
 
@@ -132,7 +145,13 @@ func (c *CmdLaunch) Run(cmd *cobra.Command, av []string) error {
 	}
 
 	changeId, err := cli.Launch(project.Id, av, mode, c.verbose)
-	if err != nil {
+
+	var waitingErr client.WaitingChangeError
+	switch {
+	case err == nil:
+	case errors.As(err, &waitingErr):
+		return c.waitingChangeError(waitingErr)
+	default:
 		return err
 	}
 
--- a/cmd/workshop/launch.go
+++ b/cmd/workshop/launch.go
@@ -99,15 +99,12 @@ $ workshop launch`,
 	return cmd
 }
 
-// waitingChangeError renders the message for an abort or continue that could
-// not be applied to a launch.
-func (c *CmdLaunch) waitingChangeError(e client.WaitingChangeError) error {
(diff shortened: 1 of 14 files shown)


cmd/workshop/launch.go
```diff
@@ -99,6 +99,19 @@ $ workshop launch`,
...
 
+// waitingChangeError renders the message for an abort or continue that could
+// not be applied to a launch.
+func (c *CmdLaunch) waitingChangeError(e client.WaitingChangeError) error {
+	verb := "abort"
+	if c.Continue {
+		verb = "continue"
+	}
+	if e.Reason == client.WaitingChangeRunning {
+		return fmt.Errorf("cannot %s: launch is in progress", verb)
```
dmitry-lyfar: What if we did this:

```
$ workshop refresh --no-wait
$ workshop launch --abort
cannot abort: launch is in progress
```
?

dmitry-lyfar (approved):
As discussed, let's make the action kind check a bit more explicit on the client and merge it.

PR #877, opened 2026-06-19
Doc: Add how-to for designing interface layout

akcano:


docs/how-to/customize-workshops/design-interface-layout.rst
```diff
@@ -0,0 +1,221 @@
+.. _how_design_interface_layout:
```
dmitry-lyfar, line 1: I think this how-to should not go as is. Let's merge the other outstanding PRs and discuss this. Mainly, the issues are duplication (e.g. wiring) and going against the pattern in the workshop's how-to where we cover interface related topics from the specific interface perspective, e.g. forward a port, add a mount (both, address adding a missing plug to an SDK; there will be another section on how to add a secret which would use the same mechanics as these two and so on). Some topics from here should become such how-to sections. E.g. share content between two SDKs to demonstrate the `connections` sections.

akcano: OK, converting this to draft, then.

Issue #788, opened 2026-05-28
workshop switch and "default" workshop


Perfect5th:
Maybe this is already a feature and I just missed it, and admittedly this is a very minor thing and just a "nice-to-have".

In a multi-workshop environment (multiple yamls under `.workshop`) it would be nice to have a way to "switch" between them and not have to specify the workshop name for every command.

e.g., something like
```
$ workshop list
WORKSHOP  STATUS  NOTES
noble     Ready   default
resolute  Ready   -

$ workshop run myaction
# runs in 'noble'

$ workshop switch resolute
resolute is now the default workshop

$ workshop list
WORKSHOP  STATUS  NOTES
noble     Ready   -
resolute  Ready   default

$ workshop run myaction
# runs in 'resolute'
```

jonathan-conder:
Agreed something like this could be useful. In my opinion it should work like a python venv, i.e. applies to the current shell session but isn't a global config value like `lxc switch` is.

PR #879, opened 2026-06-19
Improve run action completion with local resolution

### LINKED ISSUE #878: Improve action name completion performance for `workshop run`
When using `workshop run` in an actual project, I noticed that shell completion for action names can sometimes take noticeable time.

For example, given a `workshop.yaml` like this:

```yaml
actions:
  bootstrap: |
    cd /project
    ./scripts/build.sh bootstrap "$@"
  build: |
    cd /project
    ./scripts/build.sh build "$@"
　　:
  dev-setup: |
    cd /project
    ./scripts/workshop/install.sh "$@"
    ./scripts/workshop/start.sh "$@"
  dev-cleanup: |
    cd /project
    ./scripts/workshop/cleanup.sh "$@"
```

completion works as expected:

```console
workshop run -- bu<Tab>
# completes to:
workshop run -- build
```

At first, I assumed this was caused by having many action definitions in [my own project](https://github.com/shizuku198411/Raind/blob/main/workshop.yaml). However, I later noticed that completion could still take noticeable time even when the number of actions was small.

After checking the implementation, I found that action name completion currently resolves actions through the daemon path. This means completion may involve daemon communication even though the completion only needs action names from the workshop file.

Actions themselves do not directly affect the LXD container structure in the same way as operations like `workshop refresh`. They are resolved and used when `workshop run` is executed. Because of that, resolving action names from local workshop files first should be enough for the common completion case.

This would avoid unnecessary daemon/LXD-related work during shell completion and improve the interactive CLI UX.

The proposed change is:

1. Try to resolve action names from local workshop files first.
2. If local resolution fails, fall back to the existing daemon-based lookup.

This keeps the current behavior intact while making the common path faster.

shizuku198411:
Fixes: #878 

# Description

This PR improves `workshop run` action name completion by resolving actions from local workshop files first.

Previously, action completion could involve daemon communication even though action names can be read from the local workshop file. This made shell completion feel slow in some cases.

With this change, completion first tries local resolution and falls back to the existing daemon-based lookup if local resolution is unavailable, preserving the current behavior.

## Testing

- Added tests for local action completion.
- Added tests for daemon fallback behavior.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/cmd/workshop/exec.go
+++ b/cmd/workshop/exec.go
@@ -340,6 +340,14 @@ func (c *CmdRun) complete(cmd *cobra.Command, av []string, toComplete string) ([
 		return nil, cobra.ShellCompDirectiveDefault
 	}
 
+	// For action completion, prefer resolving actions from local workshop files.
+	// If local resolution fails, fall back to the existing daemon-based lookup.
+	if args.argsLenAtDash >= 0 {
+		if actions, ok := completeLocalActions(c.root.project(), args.av); ok {
+			return actions, cobra.ShellCompDirectiveNoFileComp
+		}
+	}
+
 	cli, err := c.root.noRetryClient()
 	if err != nil {
 		cobra.CompDebugln(err.Error(), false)
@@ -352,7 +360,6 @@ func (c *CmdRun) complete(cmd *cobra.Command, av []string, toComplete string) ([
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
 
-	// Argument after a separator must be an action.
 	if args.argsLenAtDash >= 0 {
 		return completeActions(cli, project, args.av)
 	}
@@ -384,6 +391,10 @@ func (c *CmdRun) complete(cmd *cobra.Command, av []string, toComplete string) ([
 }
 
 func completeActions(cli *client.Client, p *client.Project, args []string) ([]string, cobra.ShellCompDirective) {
+	if actions, ok := completeLocalActions(p.Path, args); ok {
+		return actions, cobra.ShellCompDirectiveNoFileComp
+	}
+
 	actions, err := listActions(cli, p, args)
 	if err != nil {
 		cobra.CompDebugln(err.Error(), false)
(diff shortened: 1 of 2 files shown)


jonathan-conder:
Thanks for this! I'm reluctant to duplicate this logic across the CLI and backend. Especially given that this only speeds up completion for one command.

Do you have some latency numbers to measure how big of an improvement this is? If so, is completion consistently slow or only at first? I suspect most of the latency is due to the Workshop daemon being socket-activated, it might be more useful to improve its startup speed instead of only improving one command.

shizuku198411:
Thanks for the feedback. I measured the latency, and it looks like the main issue is not the action lookup itself, but the workshop name resolution when the workshop name is omitted.

```console
$ workshop --version
0.9.2

$ lscpu | grep -E '^(Architecture|CPU\(s\)|Model name|Vendor ID)'
Architecture:                            x86_64
CPU(s):                                  8
Vendor ID:                               AuthenticAMD
Model name:                              AMD Ryzen 7 3700X 8-Core Processor
```

For example:

```console
$ workshop run -- t<Tab>

POST /v1/projects 6.081208ms 200
GET /v1/projects/a6751efd/workshops 74.525305ms 200
GET /v1/projects/a6751efd/workshops/raind-dev/actions 5.376038ms 200
```

```console
$ workshop run raind-dev -- t<Tab>

POST /v1/projects 5.984451ms 200
GET /v1/projects/a6751efd/workshops/raind-dev/actions 5.421819ms 200
```

So I agree that adding a separate CLI-side action resolution path is probably not the right approach, since it would duplicate logic between the CLI and daemon.

The workshop name lookup itself is not a huge delay in isolation. However, when completion is used frequently, the delay becomes more noticeable overall. For example, completing something like:

```console
t<Tab> -> test-u<Tab> -> test-unit
```

can trigger the workshop name lookup multiple times, so the accumulated latency can make completion feel slower.

This is just an idea, but one possible approach might be to support a default workshop name. Then, when the workshop name is omitted, completion could use the default workshop name directly instead of requesting the workshop list from the backend.

I would be interested to hear what you think about that direction.

jonathan-conder:
There's a similar open issue here: https://github.com/canonical/workshop/issues/788

I would suggest optimising the lookup before we go changing the design though. There's probably room for improvement on the backend if looking up the name takes 10x longer than creating/looking up a project. If not, we could look at caching it locally (for completion purposes only, of course).

shizuku198411:
Thanks, that sounds reasonable.

I agree that we should try to optimize the existing lookup path before changing the design or adding a separate local resolution path in the CLI.

So I’ll take a look at the backend workshop lookup path first and see why listing/resolving the workshop name is slower than project lookup or action lookup.

I’ll also check #788 and see whether this can be addressed in the same direction.
I will close this PR for now. Thank you for the review!

PR #883, opened 2026-06-23
Proxy daemon socket to a fixed path inside workshops

tlm:
# Description

SDKs with hooks could not reach their own `workshopd` when the daemon ran
with a non-default socket — most visibly under `go tool try`, which starts
each daemon with a custom `WORKSHOP_SOCKET`.

Inside a workshop, hooks and `workshopctl` talk to the daemon through an LXD
proxy device. The two ends of that proxy disagreed on the instance-side path:

- The proxy's `Listen` (instance) address derived its filename from the
  daemon's *host* socket name (`filepath.Base(dirs.SocketPath)`), so a custom
  socket produced `…/run/<custom>.untrusted`.
- `workshopctl` always connected to the fixed default name,
  `…/run/workshop.socket.untrusted`.

These coincided only when the daemon used the default socket name. With a
custom `WORKSHOP_SOCKET` the proxy exposed the socket at a path `workshopctl`
never looked at, so hooked SDKs failed to reach the right daemon.

Fix: add `dirs.WorkshopSocketPath`, the fixed in-workshop socket path, and use
it for both the proxy `Listen` address (`internal/workshop/device.go`) and the
`workshopctl` client config (`cmd/workshopctl/main.go`). The host-side
`Connect` address still follows the per-daemon `dirs.SocketPath`, so each
workshop reaches its own daemon. Adds a regression test asserting the listen
path stays fixed for a non-default daemon socket while connect follows it.

# Self-review quick check

## Docs

---

### DIFF
--- a/cmd/workshopctl/main.go
+++ b/cmd/workshopctl/main.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"syscall"
 
 	"github.com/canonical/workshop/client"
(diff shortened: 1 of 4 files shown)


jonathan-conder (approved):
Surprised it didn't already work this way! Not sure the test is really necessary but it can't hurt.

PR #875, opened 2026-06-19
Doc: Extract plugs and slots explanation

akcano:

PR #889, opened 2026-06-23
Doc: Add an auto-connection table to plugs and slots

akcano:
This follows an earlier discussion: https://github.com/canonical/workshop/pull/875#pullrequestreview-4540591369


dmitry-lyfar (approved):
Image

There is a bit of artifact in rendering, otherwise looks nice.

PR #890, opened 2026-06-23
Add vendorid and productid properties to custom-device interface

locnnil:
# Description

Add the optinal  `productid` and `vendorid` properties for custom-device interface.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```
 * [x] Put variable declaration and initialization together.

## Docs

Procedure:

* [ ] I have checked and added or updated relevant release notes. **(doesn't apply, I believe)**

Content:

---

Or:

### DIFF
--- a/docs/explanation/interfaces/custom-device-interface.rst
+++ b/docs/explanation/interfaces/custom-device-interface.rst
@@ -32,7 +32,9 @@ which is declared in the SDK definition.
 
 Its structure includes the name of the plug,
 the interface (:samp:`custom-device`),
-and a :samp:`subsystem` attribute.
+and the optional :samp:`subsystem`, :samp:`vendorid`,
+and :samp:`productid` device filters.
+At least one of these filters must be set.
 
 Defining the plug in an SDK
 allows the workshops using this SDK to connect to matching devices,
@@ -55,6 +57,29 @@ One way to query device properties is :command:`udevadm info`:
    $ udevadm info --query=property --property=SUBSYSTEM /dev/input/event0
 
      SUBSYSTEM=input
+
+
+.. _exp_custom_device_filters:
+
+Product and vendor filters
+--------------------------
+
+When a subsystem matches more devices than wanted,
+the optional :samp:`vendorid` and :samp:`productid` attributes
+narrow the selection down to devices
+reporting the given vendor and product ID.
+Both are matched against the respective identifiers reported by the kernel,
+which can be queried with :command:`udevadm info`:
+
+.. code-block:: console
+
+   $ udevadm info --query=property --property=ID_VENDOR_ID --property=ID_MODEL_ID /dev/ttyUSB0
+
+     ID_VENDOR_ID=0403
+     ID_MODEL_ID=6001
+
+Because a product ID is only meaningful within a vendor's namespace,
+setting :samp:`productid` also requires :samp:`vendorid`.
 
 
 .. _exp_custom_device_slot:
--- a/docs/reference/definition-files/_interfaces/custom-device.rst
+++ b/docs/reference/definition-files/_interfaces/custom-device.rst
@@ -11,7 +11,7 @@ Custom device interface
 The custom device interface exposes host devices
 that belong to a Linux kernel subsystem.
 
-A custom device plug is described by this attribute:
+A custom device plug is described by these attributes:
 
 .. @artefact custom-device interface attributes
 
@@ -24,10 +24,25 @@ A custom device plug is described by this attribute:
      - Value
      - Description
 
-   * - :samp:`subsystem` (required)
+   * - :samp:`subsystem` (optional)
      - string
      - The Linux kernel subsystem of the host devices to expose,
        for example :samp:`input`, :samp:`tty`, or :samp:`usb`.
+
+   * - :samp:`vendorid` (optional)
+     - string
+     - Restrict the exposed devices to those with this vendor ID,
+       quoted so it is read as a string, for example :samp:`"0403"`.
+
+   * - :samp:`productid` (optional)
+     - string
+     - Restrict the exposed devices to those with this product ID,
+       quoted so it is read as a string, for example :samp:`"6001"`.
+
+At least one of :samp:`subsystem`, :samp:`vendorid`, or :samp:`productid`
+must be set.
+Setting :samp:`productid` also requires :samp:`vendorid`,
+since a product ID is only meaningful within a vendor's namespace.
 
 Plug owner: any regular SDK; not the system SDK.
 
--- a/docs/reference/definition-files/schema-sdk.json
+++ b/docs/reference/definition-files/schema-sdk.json
@@ -20,6 +20,9 @@
       "type": "string"
     },
     "CustomDevicePlug": {
+      "$ref": "#/$defs/CustomDevicePlugModel"
+    },
+    "CustomDevicePlugModel": {
       "additionalProperties": false,
       "description": "SDKcraft project custom-device plug definition.",
       "properties": {
@@ -29,21 +32,38 @@
           "type": "string"
         },
         "subsystem": {
+          "default": "",
           "description": "Device subsystem.",
           "examples": [
             "accel",
             "usb"
           ],
-          "minLength": 1,
           "title": "Subsystem",
           "type": "string"
+        },
+        "vendorid": {
+          "$ref": "#/$defs/DeviceID",
+          "default": "",
+          "description": "Restrict the exposed devices to those with this vendor ID.",
+          "examples": [
+            "0403"
+          ],
+          "title": "Vendor ID"
+        },
+        "productid": {
+          "$ref": "#/$defs/DeviceID",
+          "default": "",
+          "description": "Restrict the exposed devices to those with this product ID.",
+          "examples": [
+            "6001"
+          ],
+          "title": "Product ID"
         }
       },
       "required": [
-        "interface",
-        "subsystem"
+        "interface"
       ],
-      "title": "CustomDevicePlug",
+      "title": "CustomDevicePlugModel",
       "type": "object"
     },
     "DesktopPlug": {
@@ -61,6 +81,10 @@
       ],
       "title": "DesktopPlug",
       "type": "object"
+    },
+    "DeviceID": {
+      "pattern": "^(?:0x)?[a-fA-F0-9]+$",
+      "type": "string"
     },
     "Endpoint": {
       "type": "string"
@@ -442,6 +466,20 @@
       ],
       "default": null,
       "title": "Source-Code"
+    },
+    "website": {
+      "anyOf": [
+        {
+          "format": "uri",
+          "minLength": 1,
+          "type": "string"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null,
+      "title": "Website"
     },
     "license": {
       "anyOf": [
--- a/internal/interfaces/builtin/custom_device.go
+++ b/internal/interfaces/builtin/custom_device.go
@@ -20,8 +20,11 @@
 package builtin
 
 import (
+	"errors"
 	"fmt"
 	"slices"
+	"strconv"
+	"strings"
 
 	"github.com/canonical/workshop/internal/interfaces"
 	"github.com/canonical/workshop/internal/interfaces/lxd_device"
@@ -51,7 +54,7 @@ const customDeviceBaseDeclarationPlugs = `
     deny-auto-connection: true
 `
 
-var knownCustomDeviceAttributes = []string{"subsystem"}
+var knownCustomDeviceAttributes = []string{"subsystem", "vendorid", "productid"}
 
 type customDeviceInterface struct{}
 
@@ -75,15 +78,64 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error 
 		}
 	}
 
-	var subsystem string
-	if err := plug.Attr("subsystem", &subsystem); err != nil {
+	// Every attribute is optional, so the map may be nil; make it writable for
+	// the parse helpers, which default absent attributes.
+	if plug.Attrs == nil {
+		plug.Attrs = make(map[string]any)
+	}
+
+	subsystem, err := parseString(plug.Attrs, "subsystem")
+	if err != nil {
 		return err
 	}
-	if subsystem == "" {
-		return fmt.Errorf(`custom-device plug "subsystem" is empty`)
+	vendorID, err := parseDeviceID(plug.Attrs, "vendorid")
+	if err != nil {
+		return err
+	}
+	productID, err := parseDeviceID(plug.Attrs, "productid")
+	if err != nil {
+		return err
+	}
+
+	// subsystem, productid, and vendorid are each optional, but the plug must
+	// narrow the exposed host devices with at least one of them.
+	if subsystem == nil && vendorID == "" && productID == "" {
+		return fmt.Errorf(`custom-device plug must contain "subsystem", "vendorid", or "productid"`)
+	}
+	// A product ID only identifies a device within a vendor's namespace.
+	if vendorID == "" && productID != "" {
+		return fmt.Errorf(`custom-device plug contains "productid" without "vendorid"`)
 	}
 
 	return nil
+}
+
+func parseString(attrs map[string]any, key string) (*string, error) {
+	object, ok := attrs[key]
+	if !ok {
+		return nil, nil
+	}
+	value, ok := object.(string)
+	if !ok {
+		return nil, fmt.Errorf("custom-device plug %q is not a string (found %T)", key, object)
+	}
+	return &value, nil
+}
+
+func parseDeviceID(attrs map[string]any, key string) (string, error) {
+	value, err := parseString(attrs, key)
+	if value == nil || err != nil {
+		return "", err
+	}
+
+	id, err := strconv.ParseUint(strings.TrimPrefix(*value, "0x"), 16, 16)
+	if err != nil {
+		return "", fmt.Errorf("custom-device plug %q must be a hexadecimal number", key)
+	}
+
+	udevForm := fmt.Sprintf("%04x", id)
+	attrs[key] = udevForm
+	return udevForm, nil
 }
 
 func (iface *customDeviceInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.SlotInfo) bool {
@@ -92,11 +144,20 @@ func (iface *customDeviceInterface) AutoConnect(plug *sdk.PlugInfo, slot *sdk.Sl
 }
 
 func (iface *customDeviceInterface) MountConnectedPlug(spec *lxd_device.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	var subsystem string
-	if err := plug.Attr("subsystem", &subsystem); err != nil {
+	device := workshop.CustomDevice{Name: plug.Name()}
+
+	var attrErr *sdk.AttributeNotFoundError
+	if err := plug.Attr("subsystem", &device.Subsystem); err != nil && !errors.As(err, &attrErr) {
 		return err
 	}
-	return spec.AddCustomDevice(workshop.CustomDevice{Name: plug.Name(), Subsystem: subsystem})
+	if err := plug.Attr("vendorid", &device.VendorID); err != nil && !errors.As(err, &attrErr) {
+		return err
+	}
+	if err := plug.Attr("productid", &device.ProductID); err != nil && !errors.As(err, &attrErr) {
+		return err
+	}
+
+	return spec.AddCustomDevice(device)
 }
 
 func init() {
--- a/internal/interfaces/builtin/custom_device_test.go
+++ b/internal/interfaces/builtin/custom_device_test.go
@@ -76,6 +76,7 @@ plugs:
     interface: custom-device
     subsystem: accel
 `, s.projectId, "ws", "consumer", "mydevice")
+	c.Assert(interfaces.BeforePreparePlug(s.iface, plug), check.IsNil)
 	connectedPlug := interfaces.NewConnectedPlug(plug, nil, nil)
 
 	slot := builtin.MockSlot(c, `name: producer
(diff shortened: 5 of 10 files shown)


internal/interfaces/builtin/custom_device.go
```diff
@@ -96,7 +96,23 @@ func (iface *customDeviceInterface) MountConnectedPlug(spec *lxd_device.Specific
...
+		if err := plug.Attr("productid", &device.ProductID); err != nil {
+			return err
+		}
+	}
+	if _, ok := plug.Lookup("vendorid"); ok {
+		if err := plug.Attr("vendorid", &device.VendorID); err != nil {
+			return err
+		}
+	}
+
```
jonathan-conder: No need for a double lookup here, you can do `errors.AsType[*sdk.AttributeNotFoundError]` or `errors.As` to detect this type of error

internal/interfaces/builtin/custom_device.go
```diff
@@ -96,7 +96,23 @@ func (iface *customDeviceInterface) MountConnectedPlug(spec *lxd_device.Specific
 	if err := plug.Attr("subsystem", &subsystem); err != nil {
```
jonathan-conder: We can make subsystem optional if a productid or vendorid is specified

internal/interfaces/builtin/custom_device.go
```diff
@@ -96,7 +96,23 @@ func (iface *customDeviceInterface) MountConnectedPlug(spec *lxd_device.Specific
...
+			return err
+		}
+	}
+	if _, ok := plug.Lookup("vendorid"); ok {
+		if err := plug.Attr("vendorid", &device.VendorID); err != nil {
+			return err
+		}
+	}
+
+	return spec.AddCustomDevice(device)
```
jonathan-conder, line 160: we should first validate the IDs (not sure exactly but I guess it's a hex-encoded uint16 or something?). Also, only allow productid if vendorid is also set. Same thing in SDKcraft

jonathan-conder (changes requested):

internal/interfaces/builtin/custom_device.go
locnnil: Agreed. Btw, should we validate the subsystem as well? 

I did a test and saw that, if I define something like
```yaml
plugs:
  dummy:
    interface: custom-device
    subsystem: foobar
```
Then do:

```shell
workshop connect hw/device:dummy hw/system:custom-device
```

It will succeed. The interface shows up as connected, but nothing happens.

We could check if the subsystem is valid by checking if it exists on `/sys/class` or `/sys/bus`.  Should I add this check as well?

jonathan-conder: It sounds like a good idea, but I would do it at connection time rather than during the basic syntax validation I would put here.

I'm also not sure if `/sys/class` is a reliable way, what if the user only plugs in the device after connecting the interface? If there's a valid list of subsystems somewhere, we could check that

locnnil: At first I thought looking at `/sys/class` and `/sys/bus` would do the job, but that only shows the subsystems currently registered on the running kernel. i.e., the ones whose modules are loaded/enabled right now.

You can have a subsystem that's perfectly valid for your kernel but whose module isn't loaded; run `modprobe` for it, and it shows up. So a subsystem can be valid without being usable, and an unusable one won't be present under `/sys/class` or `/sys/bus`. If we validate against those directories, we'd reject a config that's actually fine. or at best delay the failure.

But given the **hot-pluggable** nature of the interface, we can't refuse a subsystem just because the matching hardware (or module) isn't there yet; that's exactly the case this interface exists to handle. The device can show up at any moment, and the subsystem can register along with it.
So my suggestion: instead of treating `/sys/class` + `/sys/bus` as a hard allowlist, keep a lookup table of known-valid subsystem names and validate against that, updating it as the kernel adds new ones. Another option would be to just check against `/sys/class` + `/sys/bus`, then accept and emit a **WARN** in case nothing matches; this would prevent typos.

jonathan-conder: It sounds like too much effort for the gain

docs/explanation/interfaces/custom-device-interface.rst
```diff
@@ -57,6 +58,29 @@ One way to query device properties is :command:`udevadm info`:
...
 
+.. _exp_custom_device_filters:
+
+Product and vendor filters
+--------------------------
+
+When a subsystem matches more devices than wanted,
+the optional :samp:`productid` and :samp:`vendorid` attributes
+narrow the selection down to devices
+reporting the given product and/or vendor ID.
```
akcano:
```suggestion
reporting the given product and vendor ID.
```

docs/explanation/interfaces/custom-device-interface.rst
```diff
@@ -57,6 +58,29 @@ One way to query device properties is :command:`udevadm info`:
...
+Both are matched against the same identifiers reported by the kernel,
```
akcano:
```suggestion
Both are matched against the respective identifiers reported by the kernel,
```

docs/explanation/interfaces/custom-device-interface.rst
```diff
@@ -57,6 +58,29 @@ One way to query device properties is :command:`udevadm info`:
...
+
+.. code-block:: console
+
+   $ udevadm info --query=property --property=ID_VENDOR_ID --property=ID_MODEL_ID /dev/ttyUSB0
+
+     ID_VENDOR_ID=0403
+     ID_MODEL_ID=6001
+
+Each filter is applied only when set;
+an unset filter does not restrict the matching devices.
```
akcano: I think this is redundant.
```suggestion
```

akcano (changes requested):

internal/interfaces/builtin/custom_device.go
locnnil: I've decided to go with `errors.As` since it shows up more recurrently on the repository, according to my searches:

```shell
rg -uuu 'errors\.As\(' -g '!.git'
```
```shell
rg -uuu 'errors\.AsType' -g '!.git'
```

internal/interfaces/builtin/custom_device.go
locnnil: That's true, although I would be in favor of emitting a warning.

Regarding the vendorid and productid fields, I defined the validation as a Regex to match against `^[0-9a-f]{4}$`.

locnnil:
@jonathan-conder I moved the validation logic to `BeforePreparePlug()`, because I saw that in other interfaces that's where the validation happens. Let me know if this isn't the case for this one.

internal/interfaces/builtin/custom_device.go
```diff
@@ -51,7 +53,11 @@ const customDeviceBaseDeclarationPlugs = `
     deny-auto-connection: true
 `
 
-var knownCustomDeviceAttributes = []string{"subsystem"}
+var knownCustomDeviceAttributes = []string{"subsystem", "productid", "vendorid"}
+
+// deviceID matches a USB product or vendor ID: four lowercase hexadecimal
+// digits, the same format the LXD unix-hotplug device expects.
+var deviceID = regexp.MustCompile(`^[0-9a-f]{4}$`)
```
jonathan-conder: I wouldn't bother with a regex when you can just do `hex.DecodeString`

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +81,87 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+func optionalStringAttr(plug *sdk.PlugInfo, name string) (string, error) {
+	var value string
+	var notFound *sdk.AttributeNotFoundError
+	switch err := plug.Attr(name, &value); {
+	case err == nil, errors.As(err, &notFound):
+		return value, nil
+	default:
+		return "", err
+	}
+}
```
jonathan-conder: if we're going to have a helper function for this it should follow the pattern set by the mount interface. The comment is redundant

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +81,87 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+		return err
+	}
+
+	plugRef := plug.Ref().ShortRef()
+
+	// subsystem, productid, and vendorid are each optional, but the plug must
+	// narrow the exposed host devices with at least one of them.
+	if subsystem == "" && productID == "" && vendorID == "" {
+		return fmt.Errorf(`custom-device interface plug %q must specify at least one of "subsystem", "productid", or "vendorid"`, plugRef)
+	}
```
jonathan-conder: I also don't think there's a case where productID makes sense without a vendorID. You can override me on that though

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +81,87 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+		target *string
+	}{
+		{"subsystem", &device.Subsystem},
+		{"productid", &device.ProductID},
+		{"vendorid", &device.VendorID},
+	}
+
+	var attrErr *sdk.AttributeNotFoundError
+	for _, attr := range optional {
+		if err := plug.Attr(attr.name, attr.target); err != nil && !errors.As(err, &attrErr) {
```
jonathan-conder: BeforePreparePlug can modify plug.Attrs so we don't need to re-validate the `attrErr` here.

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +81,87 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+	device := workshop.CustomDevice{Name: plug.Name()}
+
+	optional := []struct {
+		name   string
+		target *string
+	}{
+		{"subsystem", &device.Subsystem},
+		{"productid", &device.ProductID},
+		{"vendorid", &device.VendorID},
+	}
```
jonathan-conder: Adding this loop may save a few lines now but it entangles subsystem with the other 2 attributes, making it harder to modify later

internal/interfaces/lxd_device/spec.go
```diff
@@ -286,12 +286,22 @@ func (s *Specification) AddCustomDevice(device workshop.CustomDevice) error {
...
 	name := lxdbackend.DeviceName(s.Profile.Sdk, device.Name)
-	s.devices[name] = map[string]string{
+	lxdDevice := map[string]string{
 		"type":              "unix-hotplug",
 		"subsystem":         device.Subsystem,
 		"required":          "false",
 		"ownership.inherit": "true",
 	}
+	// Only set the optional filters when provided so that an unset filter does
+	// not constrain which host devices match.
```
jonathan-conder: should subsystem be unset if empty too?

locnnil: Yes, absolutely.

internal/interfaces/builtin/custom_device.go
jonathan-conder: that's because it's old. `AsType` is the recent replacement for it. But either is fine

internal/interfaces/builtin/custom_device.go
locnnil: Yes, you're right; I added below a condition to check if productID is set but vendorID isn't. Btw, I'm considering here that: `vendorID` only is a valid setting, which makes sense, but some can argue that it might be ambiguous and lead to user mistakes. Honestly, I don't have a strong opinion. on this.

jonathan-conder: well I'm not really a hardware person so I'll defer to your judgement on this one

locnnil: Since we already have the `subsystem` field which in some cases has this effect  of "match a bunch of things," I think that allowing `vendorid` only aligns here.

docs/reference/definition-files/_interfaces/custom-device.rst
```diff
@@ -24,11 +24,26 @@ A custom device plug is described by this attribute:
...
      - Description
 
-   * - :samp:`subsystem` (required)
+   * - :samp:`subsystem` (optional)
      - string
      - The Linux kernel subsystem of the host devices to expose,
        for example :samp:`input`, :samp:`tty`, or :samp:`usb`.
 
+   * - :samp:`productid` (optional)
+     - string
```
jonathan-conder: maybe we can be more specific than just `string` here

docs/reference/definition-files/schema-sdk.json
```diff
@@ -37,6 +37,28 @@
...
+        "productid": {
+          "description": "product ID identifier",
+          "examples": [
+            "0002",
+            "0037",
+            "0402"
+          ],
+          "minLength": 1,
+          "title": "ProductID",
+          "type": "string"
```
jonathan-conder: this should be autogenerated from SDKcraft, just wanted to confirm that

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+		return "", fmt.Errorf("invalid value type %T in key %q for custom-device interface plug %q: must be a string", object, key, plugRef)
+	}
+	return value, nil
+}
+
+func parseDeviceID(attrs map[string]any, key, plugRef string) (string, error) {
+	value, err := parseString(attrs, key, plugRef)
+	if err != nil || value == "" {
+		return "", err
+	}
```
jonathan-conder, line 129: I guess we're requiring `vendorid: '1234'` in some cases. Probably a good thing rather than reinterpreting `1234` as `0x1234 = 4660`. What do you think?

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+	if err != nil {
 		return err
 	}
-	if subsystem == "" {
-		return fmt.Errorf(`custom-device plug "subsystem" is empty`)
+
+	// subsystem, productid, and vendorid are each optional, but the plug must
+	// narrow the exposed host devices with at least one of them.
+	if subsystem == "" && productID == "" && vendorID == "" {
+		return fmt.Errorf(`custom-device interface plug %q must specify at least one of "subsystem", "productid", or "vendorid"`, plugRef)
```
jonathan-conder:
```suggestion
		return fmt.Errorf(`custom-device plug must contain "subsystem", "productid", or "vendorid"`)
```
see https://github.com/canonical/workshop/blob/7fc485de1d9931cc32d71a9678b3ec5720473879/internal/interfaces/builtin/mount.go#L210

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+	}
+	// A product ID only identifies a device within a vendor's namespace.
+	if productID != "" && vendorID == "" {
+		return fmt.Errorf(`custom-device interface plug %q must specify "vendorid" when "productid" is set`, plugRef)
```
jonathan-conder:
```suggestion
		return fmt.Errorf(`custom-device plug contains "productid" without "vendorid"`)
```

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+
+func parseDeviceID(attrs map[string]any, key, plugRef string) (string, error) {
+	value, err := parseString(attrs, key, plugRef)
+	if err != nil || value == "" {
+		return "", err
+	}
+
+	id, err := hex.DecodeString(value)
+	if err != nil || len(id) != 2 {
+		return "", fmt.Errorf("invalid value %q in key %q for custom-device interface plug %q: must be four hexadecimal digits", value, key, plugRef)
```
jonathan-conder:
```suggestion
		return "", fmt.Errorf("custom-device plug %q must have four hexadecimal digits", key)
```

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
+	}
+
+	udevForm := hex.EncodeToString(id)
+	attrs[key] = udevForm
+	return udevForm, nil
```
jonathan-conder:
```suggestion
	attrs[key] = value
	return value, nil
```
not sure if there's something I'm missing here

internal/interfaces/builtin/custom_device_test.go
```diff
@@ -126,5 +127,212 @@ plugs:
...
+	c.Assert(deviceSpec.AddConnectedPlug(s.iface, connectedPlug, connectedSlot), check.IsNil)
+
+	expectedDevices := []workshop.CustomDevice{{
+		Name:      "mydevice",
+		Subsystem: "tty",
+		ProductID: "6001",
+		VendorID:  "0403",
+	}}
+	c.Assert(deviceSpec.Profile.CustomDevices, check.DeepEquals, expectedDevices)
+}
```
jonathan-conder, line 129: this should move up further, out of the `TestSanitizePlug*` section

internal/interfaces/builtin/custom_device_test.go
```diff
@@ -126,5 +127,212 @@ plugs:
...
+	plug := builtin.MockPlug(c, `name: consumer
+base: ubuntu@24.04
+plugs:
+  mydevice:
+    interface: custom-device
+    productid: "6001"
+`, s.projectId, "ws", "consumer", "mydevice")
+	err := interfaces.BeforePreparePlug(s.iface, plug)
+	c.Check(err, check.ErrorMatches, `custom-device interface plug "ws/consumer:mydevice" must specify "vendorid" when "productid" is set`)
+}
```
jonathan-conder: I would just pick one of these

internal/interfaces/builtin/custom_device_test.go
```diff
@@ -126,5 +127,212 @@ plugs:
...
+func (s *customDeviceSuite) TestSanitizePlugVendorIdOnly(c *check.C) {
+	plug := builtin.MockPlug(c, `name: consumer
+base: ubuntu@24.04
+plugs:
+  mydevice:
+    interface: custom-device
+    vendorid: "0403"
+`, s.projectId, "ws", "consumer", "mydevice")
+	c.Check(interfaces.BeforePreparePlug(s.iface, plug), check.IsNil)
+}
```
jonathan-conder: seems redundant with `TestSanitizePlugWithVendorIdFilter`

internal/workshop/lxd/lxd_backend_profile_test.go
```diff
@@ -105,6 +105,15 @@ func (f *LxdBeTests) TestLxdToSdkProfileOK(c *check.C) {
...
 		},
+		{
+			Sdk:    "sdk",
+			Mounts: map[string]workshop.Mount{},
+			CustomDevices: []workshop.CustomDevice{{
+				Name:      "device",
+				Subsystem: "tty",
+				ProductID: "6001",
+				VendorID:  "0403"}},
+		},
```
jonathan-conder, line 111: we can just extend the previous entry instead of making a new one

docs/reference/definition-files/_interfaces/custom-device.rst
```diff
@@ -24,11 +24,26 @@ A custom device plug is described by this attribute:
...
+     - Restrict the exposed devices to those with this product ID,
+       for example :samp:`6001`.
```
jonathan-conder: we should mention having to quote the ID somewhere

internal/interfaces/builtin/custom_device.go
locnnil: This is a good consideration, but I'd argue that which software/where would the user get a decimal representation and not a hex for its vendor ID/product ID? I believe it's better to keep it strict, and I'd argue hex-only is more than a style choice: it matches the whole toolchain that produces these values.

The workflow that is most likely users will follow is `lsusb` → copy the ID → paste into config. `lsusb` prints hex (`ID 1d6b:0002`), sysfs (`/sys/bus/usb/devices/*/idVendor`) stores 4 hex digits with no prefix, and `udevadm` reports hex too. 

LXD, for it 's own `usb` and `unix-hotplug` devices do the same: `vendorid`/`productid` are hex strings (e.g. `vendorid=0fce productid=51da`), and the docs for `usb` tell users to read them from `lsusb`. Expecting hex means a value copied straight from `lsusb` just works.

Accepting both representations fails in the worst way: silently. For any all-decimal-digit ID like `1234`, both interpretations parse fine, so the user gets no error, just the wrong device (`0x1234` vs `0x04D2 = 1234`). A rejected input is easy to debug; a silently-wrong device isn't.

Requiring a quoted hex string also sidesteps type coercion in the config itself: a bare `vendorid: 0451` can be read as octal or lose its leading zero, and real VIDs like `0451` (TI) or `046d` (Logitech) have leading zeros. Quoting keeps them intact, which is probably why we already see `'1234'`.

The only place I've found decimal is [the USB-IF vendor list PDF](https://usb.org/sites/default/files/vendor\_ids051920\_0.pdf), which seems to target hardware vendors, not people dealing with software and wiring up passthrough. Everything user-facing, like [devicehunt](https://devicehunt.com/vendors/usb), `lsusb`, sysfs, LXD, is hex.

So I'd keep it strict: a 4-hex-digit string, no `0x` needed. Accepting both would force us to mandate a `0x` prefix to disambiguate, which adds a formatting rule *and* still breaks copy-paste from `lsusb`. Happy to reconsider if you feel strongly, but IMHO strict-hex looks like the lower-friction and safer default.

docs/reference/definition-files/_interfaces/custom-device.rst
locnnil: Something like "4-digit-hex string"?

internal/interfaces/builtin/custom_device_test.go
locnnil: Moved to the top with the other `TestCustomDeviceInterface*` functions.

docs/reference/definition-files/schema-sdk.json
locnnil: Actually, I don't; I write it by hand.

From what I found out, by checking [this file](https://github.com/canonical/sdkcraft/blob/main/sdkcraft/models/project.py#L253-L255) and [this file](https://github.com/canonical/sdkcraft/blob/main/sdkcraft/models/metadata.py#L63-L65) I should be running those scripts right?

So, that's what I did:
```shell
      cd sdkcraft                                                                                                                                                                                                                             
      uv run python sdkcraft/models/metadata.py                                                                                                                                                                                               
      uv run python sdkcraft/models/project.py
      mv schema-sdk.json schema-sdkcraft.json ../workshop/docs/reference/definition-files/
```
Please let me know if this approach is right.

Also, since it generated some changes unrelated to the `custom-device` interface (like the `website` field), I decided to leave it on a separate commit.

docs/reference/definition-files/_interfaces/custom-device.rst
locnnil: I believe the quotes wouldn't be necessary if we agree on this: https://github.com/canonical/workshop/pull/890#discussion_r3506446398. Let me know what you think.

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +76,89 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
 
+func parseString(attrs map[string]any, key, plugRef string) (string, error) {
+	object, ok := attrs[key]
+	if !ok {
+		attrs[key] = ""
+		return "", nil
+	}
+	value, ok := object.(string)
+	if !ok {
+		return "", fmt.Errorf("invalid value type %T in key %q for custom-device interface plug %q: must be a string", object, key, plugRef)
```
jonathan-conder:
```suggestion
		return "", fmt.Errorf("custom-device plug %q is not a string (found %T)", key, object)
```
see https://github.com/canonical/workshop/blob/7fc485de1d9931cc32d71a9678b3ec5720473879/internal/interfaces/builtin/mount.go#L214

internal/interfaces/builtin/custom_device.go
jonathan-conder: yes agreed, however the code as it stands doesn't handle the unquoted 1234 case, I think. I'd rather have a hard error for now. The interface attribute parsing is quite brittle and if we rework it there's probably a clean way to "assume hex"

docs/reference/definition-files/_interfaces/custom-device.rst
locnnil: Added a sentence on the need for quotations.

docs/reference/definition-files/_interfaces/custom-device.rst
locnnil: I've changed to `Hex string` let me know what you think

jonathan-conder: let's just leave it as `string` actually. I'll do a couple of updates so we make the release in time.

docs/reference/definition-files/schema-sdk.json
jonathan-conder: yes I think we have automation for this, so it's not expected to update it with every PR. might as well do it here though

jonathan-conder (approved):

internal/interfaces/builtin/custom_device.go
```diff
@@ -75,28 +77,88 @@ func (iface *customDeviceInterface) BeforePreparePlug(plug *sdk.PlugInfo) error
...
 		return err
 	}
-	if subsystem == "" {
-		return fmt.Errorf(`custom-device plug "subsystem" is empty`)
+
+	// subsystem, productid, and vendorid are each optional, but the plug must
+	// narrow the exposed host devices with at least one of them.
+	if subsystem == nil && vendorID == "" && productID == "" {
+		return fmt.Errorf(`custom-device plug must contain "subsystem", "vendorid", or "productid"`)
+	}
```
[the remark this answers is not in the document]
jonathan-conder: LXD already handles this case

docs/reference/definition-files/schema-sdk.json
```diff
@@ -62,6 +82,10 @@
       "title": "DesktopPlug",
       "type": "object"
     },
+    "DeviceID": {
+      "pattern": "^(?:0x)?[a-fA-F0-9]+$",
+      "type": "string"
```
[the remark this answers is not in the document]
jonathan-conder: what about 00000000c? it's valid

dmitry-lyfar (approved):

jonathan-conder:
Fixes #891 (for future reference)

PR #911, opened 2026-07-03
Initial support for VMs

jonathan-conder:
# Description

Introduces VM workshops: after running `snap set workshop workshop.experimental-vms=1`, add `confinement: virtual-machine` to the workshop definition file. Only limited functionality is supported (e.g. no SDKs).

## Major chages

- Starting workshops (including during launch and refresh, but not at system boot) now relies on `workshop-waitready.service`. This is installed in workshops launched or refreshed with Workshop 0.9.5 or later. Older workshops are still supported, but will be dropped in future. After we drop support, `lxc start` can be run manually as a workaround.
- Adds support for freezing VM filesystems. LXD doesn't natively support this for VM root filesystems, but support will be added in future, possibly based on this implementation. This is required to safely snapshot a workshop VM (but snapshots aren't used anywhere yet).
- Workshop can now launch, start, stop, refresh, restore and remove stock Ubuntu VMs.

### Freeze limitations

I think the `fsfreeze` implementation is slightly better than QEMU's guest agent, but has some of the same limitations: no ZFS support and potential timeout issues in obscure cases like mounting a loop file stored on a network drive. The latter is probably not worth worrying about; the former is something we could potentially support but there's no reason to right now. LXD VMs typically use ext4 for the / partition even when the block device is backed by ZFS.

## Minor changes

- Adds a `--vm` flag to `workshop init`.
- Adds `confinement` in `workshop info` output.
- Fixes an issue where LXD client websocket goroutines can race with error handling when writing to a `stderr` buffer.
- Splits `tests/main/launch` and `tests/integration/workshop-snapshots` into multiple spread variants. This is a step towards enabling them for VMs, but we need to clean up snapshots more aggressively or the GitHub runners will run out of space.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:


jonathan-conder:
@dmitry-lyfar updated

New commits:
- 0df6c68a Fix potential data race in projectFsRoot
- 71998b56 Add temporary support for starting pre-0.9.4 workshops
- b54a7a3c Add support for freezing filesystems before taking snapshots
- 3e3dfa34 Stop VMs gracefully during refresh

Major changes:
- Added `fsfreeze` to cloud-init and updated snapshot tests (f5fc4dc1)
- Gated VMs behind `snap set workshop.experimental-vms=1` (948add58)
- Added VMs and 26.04 to `tests/main/launch` (948add58)

Minor changes:
- Resolved a merge conflict in `lxd_backend.go` (1d706912 vs fcf0d26a)
- Resolve a rebase conflict in `lxd_backend.go` (187fa1c3 vs a673c8cd)
- Updated snapshot tests again to account for GRUB config (a673c8cd)
- Fix snapshot tests after adding more whitespace to raw.idmap (46664bea)
- Two new `manifestSuite` tests: `TestRefreshRequiresContainer` and `TestLaunchRejectsVMsWithSDKs`
- Added test for `workshop init --vm` (c21dbbce)

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar, line 1415:
```suggestion
      Type=notify
      Restart=on-failure
      RestartSec=2s
```

```suggestion
      Description=Signal workshop readiness to LXD
      After=lxd-agent.service
```

internal/fsfreeze/fsfreeze.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (w *ReadyWriter) Write(data []byte) (int, error) {
+	n, err := w.buf.Write(data)
+	for {
+		line, _, found := bytes.Cut(w.buf.Bytes(), []byte{'\n'})
+		if !found {
+			break
+		}
+		if bytes.Equal(line, []byte("READY=1")) {
```
dmitry-lyfar, line 296: I was digging into this and it looks ok, however, llm suggests this to avoid potential panics on retries / possible second calls in tests.

```
w.once.Do(func() { close(w.ready) })
```

internal/workshop/lxd/lxd_backend_snapshots.go
```diff
@@ -406,6 +455,84 @@ func (s *Backend) TakeSnapshot(ctx context.Context, name string, snapshot worksh
...
+	exectx, err := s.execCommand(conn, ctx, name, args)
+	if err != nil {
+		stdin.Close()
+		return nil, nil, err
+	}
+
+	// The fsfreeze process has its own timeout. If we interrupt it we risk
+	// leaving filesystems frozen, but cancelling WaitExecution doesn't
+	// cancel the execution itself.
+	waitCtx, cancel := context.WithTimeout(ctx, freezeTimeout)
```
dmitry-lyfar:
```suggestion
	waitCtx, cancel := context.WithTimeout(context.Background(), freezeTimeout)
```
Maybe we should make it non-cancelable externally and only do so on a hard timeout.

jonathan-conder:
Summary of changes since my last comment:

- Only check for old workshops when using the LXD backend.
- Fixed shellcheck issue in the `fsfreeze` package.
- Add `agent.nic_config=true` to LXD VM config; seems to be necessary to override `eth0` config in `systemd-networkd`, but I'm not sure it's reliable yet.
- Switched back to force-stopping containers in the snapshot tests. We need to stop VMs gracefully but there's an issue where gracefully stopping containers times out.
- Use ID-mapped mounts for container root filesystems in the snapshot tests. This allows another container to modify the files, so we can share test code with the VM case.
- Ignore `/var/snap/lxd/common` when diffing snapshot filesystems.
- Check `/project` idmap in `tests/main/launch`.
- Limit `workshop.experimental-vms` gate to `workshop launch`.
- Add `confinement` field to `workshop info`.
- Split `tests/main/launch` and `tests/integration/workshop-snapshots`. VM tests are disabled until I have time to experiment with the limits of our GitHub runners.

internal/workshop/lxd/tests/integration/snapshot_test.go
```diff
@@ -131,11 +127,12 @@ var snapshotFormat []byte
 // backend-agnostic conventions are covered by `apiSuite.TestSnapshotFormat`.
 func (s *snapshotSuite) TestLxdBackendSnapshotFormat(c *check.C) {
 	var format map[string]any
+	c.Assert(format, check.HasLen, 5)
 	err := yaml.Unmarshal(snapshotFormat, &format)
 	c.Assert(err, check.IsNil)
```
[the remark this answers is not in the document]
jonathan-conder: this was a quick hack I was using to run the spread tests locally, removed now.

internal/workshop/lxd/lxd_backend_snapshots.go
```diff
@@ -406,6 +455,84 @@ func (s *Backend) TakeSnapshot(ctx context.Context, name string, snapshot worksh
...
+			err = errors.New("fsfreeze never reported ready")
+		}
+		return nil, nil, err
+	}
+}
+
+func thawFilesystems(ctx context.Context, thaw io.Closer, result <-chan error) error {
+	thaw.Close()
+
+	waitCtx, cancel := context.WithTimeout(ctx, thawTimeout)
```
dmitry-lyfar:
```suggestion
	waitCtx, cancel := context.WithTimeout(context.Background(), thawTimeout)
```
Same logic here

dmitry-lyfar (commented):
I need to finish tests review, these are the points, I've come across so far.

internal/idmap/idmapset_linux.go
```diff
@@ -0,0 +1,496 @@
...
+		case isBetween(i.Nsid, e.Nsid, e.Nsid+e.Maprange-1):
+			return true
+		case isBetween(e.Nsid+e.Maprange-1, i.Nsid, i.Nsid+i.Maprange-1):
+			return true
+		case isBetween(i.Nsid+i.Maprange-1, e.Nsid, e.Nsid+e.Maprange-1):
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
jonathan-conder: @tomponline we took this code from LXD, this comment looks relevant for you as well

internal/workshop/lxd/lxd_backend_snapshots.go
jonathan-conder: Here `ctx` is already non-cancellable, and it's an internal function so no other callsites. But no reason not to do this.

internal/workshop/lxd/lxd_backend.go
jonathan-conder: this one won't apply to old workshops, just FYI

internal/workshop/lxd/lxd_backend.go
```diff
@@ -694,44 +686,155 @@ func (s *Backend) startWorkshop(conn lxd.InstanceServer, ctx context.Context, na
...
+
 	var stderr strings.Builder
 	args := workshop.Execution{
 		ExecArgs: workshop.ExecArgs{
-			UserId:  0,
-			GroupId: 0,
-			Command: []string{
-				"bash", "-euc", startCommand,
-			},
+			Command: []string{"systemctl", "is-system-running", "--wait"},
```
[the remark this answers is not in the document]
jonathan-conder: Fixed and tested now

internal/workshop/lxd/lxd_backend.go
dmitry-lyfar: It's OK, this is hardening that I believe does not change much in regular circumstances. Remove & relaunch is cheap otherwise.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -1373,36 +1523,118 @@ runcmd:
 
 	f, err := yaml.Marshal(file)
 	if err != nil {
-		return map[string]string{}, err
+		return nil, err
+	}
+
+	idmapSet, err := workshopIdmap(file.Confinement, userid, groupid)
```
dmitry-lyfar, line 1492: Can I ask you to summarise the idmap logic and problems we address with calculating idmaps here? Links to the existing issues work too.

jonathan-conder: ### Problem it addresses

Containers and VMs have inconsistent idmaps by default. The ideal solution is to fix this problem in LXD: https://github.com/canonical/lxd/issues/18875. But for now, this workaround does the same thing with some minor caveats.

I don't know of any real world cases that require consistency, but suppose you want to limit your agent to a certain subdirectory of `/project/`. If you try that with a VM before the idmap commit, you'll see something like this:
```console
$ mkdir slop
$ chmod 777 slop
$ workshop exec -- sudo useradd -m agent
$ workshop exec -- sudo -u agent touch /project/slop/example.go
touch: cannot touch '/project/slop/example.go': Invalid argument
```
Whereas it works in containers.

I added a test, which should pass for both containers and VMs, [here](https://github.com/canonical/workshop/blob/8bbe22b037b37d219af91fb32c2188226d24dbb7/tests/main/launch/task.yaml#L24-L31).

### Summary of the logic

For containers, the outcome is the same as before (just different whitespace). The implementation differs in that we express the idmap for the `workshop` user and group as a `[]idmap.IdmapEntry` and render it using `formatIdMap`.

For VMs, the outcome is an extended `raw.idmap`, but `volatile.idmap.next` should be the same as containers.

The logic to generate `raw.idmap` for VMs is modelled on how LXD computes the container idmap [here](https://github.com/canonical/lxd/blob/6cb7ca22d2677e47fa02ffb9badde7b2d5478555/lxd/instance/drivers/driver_lxc.go#L413-L430), but:
- we skip parsing `raw.idmap` and just start with the resulting list of `IdmapEntry`s.
- LXD merges this into a copy of `state.OS.IdmapSet.Idmap`, we use `idmap.KernelDefaultMap()` instead.

When using the LXD snap, the results are identical:
- `state.OS.IdmapSet.Idmap` is constructed using `idmap.GetIdmapSet()`.
- `idmap.GetIdmapSet()` is just some logging and error handling on top of `idmap.DefaultIdmapSet("", "")`; we copy the error handling.
- `idmap.DefaultIdmapSet("", "")` returns `idmap.kernelDefaultMap()` unless it finds `/etc/subuid` or `/etc/subgid`; these don't exist in the confined LXD snap.

If we wanted to support non-snap LXD installations, the most reliable way would be to call `idmap.DefaultIdmapSet("/proc/<LXD PID>/root", "root")`, but this would make it harder to run `workshopd` as an unprivileged user. Instead we could try to detect whether LXD is a snap or not; I opted not to bother implementing this because it's not a case we support anyway.

Another difference is that LXD also tries to extend `raw.idmap` for VMs [here](https://github.com/canonical/lxd/blob/6cb7ca22d2677e47fa02ffb9badde7b2d5478555/lxd/device/device_utils_disk.go#L215), but only does it when the existing idmap doesn't handle UID 0 or GID 0. I think this is possible, but it would require running LXD in a user namespace with lots of available UIDs and few available GIDs, or vice versa; I don't think it's likely to occur in practice.

internal/idmap/idmapset_linux.go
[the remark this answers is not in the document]
tomponline: Thanks @jonathan-conder would you be able to contribute a fix to LXD for this? Thanks!

jonathan-conder:
@dmitry-lyfar rebased on main; the main changes are:

In "Add virtual-machine confinement":
- Moved launch/refresh gating in manifest.go to a later commit
- Added workaround for `actions/upload-artifact` not liking `:` in filenames (for code coverage coming from spread variants)
- Added VM snapshot format tests; removed cloud-init hash from these because the unit tests can cover it and it's annoying to update
- Split snapshot diff tests into separate functions per base and confinement
- Stop containers and VMs gracefully in these tests
- Excluded a couple more files from the diff: `/var/cache/motd-news` and `/var/lib/chrony`

In "Enable VM support":
- Check if VM support is disabled before querying base image and SDK Store
- Allow SDKs (including system SDK) if LXD is recent enough (checked during `checkServerCapabilities`)
- Disable auto-connect, restore-conns and check-health tasks for VMs (I have a more granular approach for interfaces but it will be a separate PR)
- Enables VM variants in `tests/main/` for `autostart`, `launch-multiple`, `launch`, `remove`, `ssh`, `start` and `stop`

In "Add workshop init --vm flag":
- Removed mutual exclusivity for `--vm` and `--sdks`
- Include `File.Confinement` in validation (no reason not to due to earlier changes)

dmitry-lyfar (approved):

internal/idmap/idmapset_linux.go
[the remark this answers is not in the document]
jonathan-conder: > would you be able to contribute a fix to LXD for this?

I almost forgot but it's on my TODO list now, should be able to this week

PR #923, opened 2026-07-07
Fix reloading of interface connections on restart

dmitry-lyfar:


jonathan-conder (approved):
Nice find

dmitry-lyfar:
> Nice find

I should have known before!

PR #931, opened 2026-07-09
Doc: Standardize, codify how-to prerequisites

akcano:


docs/how-to/customize-workshops/use-host-devices.rst
```diff
@@ -27,6 +27,18 @@ find the subsystem, declare a plug, connect the interface,
...
 
+Prerequisites
+-------------
+
+Before starting, ensure you have these requirements satisfied:
+
+- A host device you want to expose to the workshop,
+  such as an input device at :file:`/dev/input/event0`.
+
+- :command:`udevadm` available on the host
```
dmitry-lyfar: `udevadm` is installed by default across modern linux distros as I can tell from my research.

docs/how-to/develop-with-workshops/run-jetbrains-gateway.rst
```diff
@@ -21,14 +21,14 @@ letting you use your favorite JetBrains IDEs against |ws_markup|.
 Prerequisites
 -------------
 
-Before starting, ensure you have:
+Before starting, ensure you have these requirements satisfied:
```
dmitry-lyfar, line 24: These prereqs are dated now.

I noticed that the list ordering is broken on this page: https://canonical-workshopubuntucom--931.com.readthedocs.build/how-to/develop-with-workshops/run-jetbrains-gateway/

But also the manual appears to be outdated as we do generate and copy SSH keys into every workshop. All you need to do to connect to these workshops via SSH is to include in `~/.ssh/config`:

```
Include /var/snap/workshop/current/ssh/<UID>/config
```

I also think that adding a plug to the system SDK in the how-to to enable SSH is just incorrect, there is no slot on the other end! :-)  JetBrains gateway needs to use standard SSH port if that's possible OR workshop's opensshd needs to be reconfigured to live on 2200.

dmitry-lyfar (changes requested):

docs/how-to/customize-workshops/use-host-devices.rst
jonathan-conder: it's part of systemd which I think is already a dependency (at least for the interface system)

docs/how-to/develop-with-workshops/run-jetbrains-gateway.rst
jonathan-conder: we need to change this to `/var/snap/workshop/common` though

docs/how-to/customize-workshops/use-host-devices.rst
dmitry-lyfar: @jonathan-conder I'm going to merge this one `/var/snap/workshop/current/ssh/<ID>/config`, let's update the how-tos when you make the SSH dir change + automatic install of the config. I moved it to jira to pick up at pulse planning.

dmitry-lyfar (approved):

Issue #891, opened 2026-06-23
Error when trying to use `tty` subsystem on `custom-device` interface


locnnil:
When trying to declare an in-project SDK like

```yaml
name: device

plugs:
  serial-port:
    interface: custom-device
    subsystem: tty
```
And use it in a workshop:

```yaml
name: hw
base: ubuntu@24.04
sdks:
  - name: project-device
```

When trying to connect:

```shell
workshop connect hw/device:serial-port hw/system:custom-device
```

I got the error:

```console
 $  workshop connect hw/device:serial-port hw/system:custom-device
error: cannot perform the following tasks:
- Connect "hw/device:serial-port" to "hw/system:custom-device" (The following instances failed updating (profile change still saved):
 - Project: workshop.1049349021, Instance: hw-634648c8: Failed starting device "device_serial-port": Failed adding mount for device inside container: Failed running: /snap/lxd/current/sbin/lxd forkmount lxc-mount -- workshop.1049349021_hw-634648c8 /var/snap/lxd/common/lxd/containers /var/snap/lxd/common/lxd/logs/workshop.1049349021_hw-634648c8/lxc.conf /var/snap/lxd/common/lxd/devices/workshop.1049349021_hw-634648c8/unix.device_serial--port.dev-console /dev/console none 4096: exit status 1
)
```

But when looking into `/dev` on the workshop container, the `tty` devices are there. 
```console
workshop@hw:/project$ ls /dev
console  full     log     net   pts     stderr  tty    ttyS10  ttyS13  ttyS16  ttyS19  ttyS21  ttyS24  ttyS27  ttyS3   ttyS4  ttyS7  ttyUSB0  zfs
core     fuse     lxd     null  random  stdin   ttyS0  ttyS11  ttyS14  ttyS17  ttyS2   ttyS22  ttyS25  ttyS28  ttyS30  ttyS5  ttyS8  urandom
fd       initctl  mqueue  ptmx  shm     stdout  ttyS1  ttyS12  ttyS15  ttyS18  ttyS20  ttyS23  ttyS26  ttyS29  ttyS31  ttyS6  ttyS9  zero

```

Although the `workshop connections` command shows that it isn't connected:

```shell
workshop connections hw
INTERFACE      PLUG                   SLOT  NOTES
custom-device  hw/device:serial-port  -     -
```

jonathan-conder:
That's interesting, I thought we had solved this type of issue. Will take a look

locnnil:
[...]

PR #939, opened 2026-07-15
Doc: Update custom device how-to

akcano:
@locnnil also for your attention.

### DIFF
--- a/docs/how-to/customize-workshops/use-host-devices.rst
+++ b/docs/how-to/customize-workshops/use-host-devices.rst
@@ -2,9 +2,10 @@
 
 .. meta::
    :description: How-to guide on accessing host devices inside a workshop with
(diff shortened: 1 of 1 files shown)


locnnil:
I think it is worth mentioning (perhaps on an admonition note) that it is recommended to avoid using `subsystem: tty` **only**.

Since this would result in this error:

- https://github.com/canonical/workshop/issues/891

The addition of the fields `productid` and `vendorid` provides an **alternative** so the user doesn't hit this error. But the error is still there if you specify only the `tty`  subsystem without any `vendorid` or `productid`.

akcano:
@locnnil thanks!

dmitry-lyfar (approved):

dmitry-lyfar:
The conflict prevents merging here.

PR #942, opened 2026-07-16
Add instance readiness service

jonathan-conder:
# Description

By default, LXD returns from a `start` operation ASAP. For VMs, it happens even before the `lxd-agent` is running, so lots of other operations are unavailable, including `exec`.

Workshop has a different approach of waiting until systemd says the instance is ready. For VMs, we need a new implementation that doesn't rely on `exec`. Thankfully, LXD partially supports this scenario by distinguishing between "running" and "ready." See https://discuss.linuxcontainers.org/t/lxd-instance-ready-state/14657 for context.

This PR adds a systemd service to every workshop that handles the task of updating the instance state. The implementation is based on `systemctl is-system-running --wait`. To avoid the need for multiple `workshopctl`-type binaries, the logic is embedded in `workshopctl` itself, and activated by setting `argv[0]` to `waitready` (e.g. running it via a symlink with that name).

In future, we'll be able to rely on this service for `workshop start`, but for now the existing implementation remains in place. However, there's a workaround for a race condition where Workshop loads the instance config, LXD adds the `volatile.last_state.ready` option, and then Workshop tries to add `boot.autostart` to the config. This race only happens one way, DevLXD does a direct DB update without resetting the entire config.

## Guest bin directory

Another issue with VMs is they don't support mounting single files, so this PR moves `workshopctl` into a dedicated directory (while still supporting "all binaries in the same dir" for local development).

This is not quite a breaking change because `workshopctl` is only supposed to be used towards the end of a launch or refresh, so it doesn't really matter if it stops working in old-format workshops (and we don't support restoring or resuming refresh for old workshops).

## Minor changes

Fixes the permissions on `/var/lib/workshop/state` when it's not mounted.

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/internal/dirs/dirs.go
+++ b/internal/dirs/dirs.go
@@ -36,6 +36,9 @@ var (
 	// base directory inside a workshop
 	WorkshopBaseDir = defaultBaseDir
 
+	// Directory for mounted binaries (i.e. workshopctl)
+	WorkshopGuestBinDir = filepath.Join(WorkshopBaseDir, "bin")
+
 	// SDKs directory to install an SDK in a workshop
 	WorkshopSdksDir = filepath.Join(WorkshopBaseDir, "sdk")
 
@@ -64,8 +67,8 @@ var (
 	BaseDir string
 	// Cache directory for workshopd
 	CacheDir string
-	// Work directory
-	ExecDir string
+	// Path for workshopctl executable
+	WorkshopCtlPath string
 	// The directory to store downloaded base images and associated metadata
 	BaseDownloads string
 	// The directory to store downloaded SDKs
@@ -102,19 +105,30 @@ func getEnvPaths() (workshopdDir, cacheDir, socketPath string) {
 	return workshopdDir, cacheDir, socketPath
 }
 
-func init() {
-	var err error
-	var execPath string
-	execPath, err = os.Executable()
+func getWorkshopCtlPath() string {
+	execPath, err := os.Executable()
 	if err != nil {
-		panic("cannot get working directory")
+		panic(fmt.Errorf("cannot get executable path: %w", err))
 	}
 
-	ExecDir = filepath.Dir(execPath)
+	// Packages use a dedicated $prefix/lib/workshop/guest directory.
+	binDir := filepath.Dir(execPath)
+	workshopctl := filepath.Join(filepath.Dir(binDir), "lib", "workshop", "guest", "workshopctl")
+	if _, err := os.Stat(workshopctl); err == nil {
+		return workshopctl
+	}
+
+	// Local development often uses `go install`, which places all binaries in
+	// the same directory.
+	return filepath.Join(binDir, "workshopctl")
+}
+
+func init() {
 	XdgRuntimeDirBase = "/run/user"
 	BaseDir, CacheDir, SocketPath = getEnvPaths()
 	SetRootDir(BaseDir)
 	SetCacheDir(CacheDir)
+	WorkshopCtlPath = getWorkshopCtlPath()
 }
 
 func SetRootDir(rootdir string) {
--- a/internal/dirs/dirs.go
+++ b/internal/dirs/dirs.go
@@ -42,9 +42,6 @@ var (
 	// Base directory for the state storage
 	WorkshopStateDir = filepath.Join(WorkshopBaseDir, "state")
 
-	// Base directory for the SDK state storage
-	WorkshopSdkStateDir = filepath.Join(WorkshopStateDir, "sdk")
-
 	// Run directory inside workshop
 	WorkshopRunDir = filepath.Join(WorkshopBaseDir, "run")
 
--- a/internal/overlord/ifacestate/ifacemgr.go
+++ b/internal/overlord/ifacestate/ifacemgr.go
@@ -351,22 +351,29 @@ func (m *InterfaceManager) ResolveDisconnect(
 // invalid on the daemon restart / update. Thus, recreating it upon every
 // daemon restart makes sure it still points to the correct files.
 func (m *InterfaceManager) recreateInternalMounts(pctx context.Context, w string) error {
-	// Recreate workshopctl bind mount, this has to be done if, for example,
-	// workshopctl was updated to a new version and is shown as /deleted in a
-	// workshop.
-	workshopctl := workshop.Mount{
-		Name:     "workshop.workshopctl",
+	// Recreate the workshopctl bind mount, this has to be done if, for
+	// example, the Workshop snap was updated to a new revision.
+	mount := workshop.Mount{
+		Name:     "workshop.bin",
 		Type:     workshop.HostWorkshop,
-		What:     filepath.Join(dirs.ExecDir, "workshopctl"),
-		Where:    "/usr/bin/workshopctl",
+		What:     filepath.Dir(dirs.WorkshopCtlPath),
+		Where:    dirs.WorkshopGuestBinDir,
 		ReadOnly: true,
 	}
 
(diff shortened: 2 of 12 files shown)


internal/waitready/waitready.go
```diff
@@ -0,0 +1,123 @@
...
+		return false, err
+	}
+
+	var state string
+	if err := variant.Store(&state); err != nil {
+		return false, err
+	}
+
+	// Based on `systemctl is-system-running --wait`.
+	return state != "initializing" && state != "starting", nil
```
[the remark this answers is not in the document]
jonathan-conder: read the damn comment! the logic is the same as systemd itself

internal/waitready/waitready.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if ready {
+		return nil
+	}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-signal:
+		return nil
+	}
```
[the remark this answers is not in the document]
jonathan-conder: you'd think the AI would understand systemd better than me...

internal/workshop/lxd/lxd_backend.go
```diff
@@ -1301,15 +1312,41 @@ write_files:
...
+	var cloudConfig strings.Builder
+	dot := struct {
+		GuestBin string
+	}{
+		GuestBin: filepath.Base(dirs.GuestBin),
+	}
+	t := template.Must(template.New("cloud-config").Parse(cloudConfigTemplate))
+	if err := t.Execute(&cloudConfig, dot); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
jonathan-conder: the string is constant...

tests/main/start/task.yaml
```diff
@@ -14,6 +14,15 @@ execute: |
...
   }
 
+  # TODO: remove this when `workshop start` begins to rely on LXD events.
+  function check_lxd_ready() {
+    echo "Check LXD itself reports the instance as Ready"
+    instance="ws-start-$(< .workshop.lock)"
+    # Allow some time for LXD to process the status update.
+    sleep 0.1s
+    lxc list --format csv --columns s --project workshop.ubuntu "name=$instance" | MATCH "^READY$"
+  }
```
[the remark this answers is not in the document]
jonathan-conder: normally I would agree, but this test is just a sentinel to ensure the logic works correctly before we begin to rely on it. if it takes longer I want to know about it

jonathan-conder:
@dmitry-lyfar I'm not totally confident about this one after this test failure: https://github.com/canonical/workshop/actions/runs/29546866249/job/87781543072. Will see if I can reproduce locally

tests/main/start/task.yaml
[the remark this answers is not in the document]
jonathan-conder: changed my mind about this after seeing https://github.com/canonical/workshop/actions/runs/29546866249/job/87781543072. Can't reproduce locally so it looks like 0.1s is simply not enough.

internal/dirs/dirs.go
```diff
@@ -67,8 +70,8 @@ var (
 	BaseDir string
 	// Cache directory for workshopd
 	CacheDir string
-	// Work directory
-	ExecDir string
+	// Full path for workshopctl
+	GuestBin string
```
dmitry-lyfar: `WorkshopCtlBin` or `CtlBin` which resides in the directory for mounted binaries named `WorkshopGuestBinDir`?

internal/overlord/ifacestate/ifacemgr.go
```diff
@@ -351,23 +351,30 @@ func (m *InterfaceManager) ResolveDisconnect(
 // invalid on the daemon restart / update. Thus, recreating it upon every
 // daemon restart makes sure it still points to the correct files.
 func (m *InterfaceManager) recreateInternalMounts(pctx context.Context, w string) error {
-	// Recreate workshopctl bind mount, this has to be done if, for example,
-	// workshopctl was updated to a new version and is shown as /deleted in a
-	// workshop.
-	workshopctl := workshop.Mount{
-		Name:     "workshop.workshopctl",
+	// Recreate the workshopctl bind mount, this has to be done if, for example,
```
dmitry-lyfar: I think this comment is not relevant anymore as we bind mount the directory and even if `workshopctl` will be updated on the next workshop snap refresh this mount does not have to be recreated on the daemon restart. I even suppose we don't have to have it in the `recreateInternalMounts` anymore.

dmitry-lyfar (approved):
I think some name adjustment makes it more readable, see suggested. Otherwise, checking readiness logic is pretty solid to me.

internal/overlord/ifacestate/ifacemgr.go
jonathan-conder: I think it does have to be recreated because it will be `/snap/workshop/123` or something

internal/dirs/dirs.go
jonathan-conder: in integration tests it's not always literally "workshopctl," but I haven't implemented that part yet

jonathan-conder: I went with `WorkshopCtlPath` to be similar to `SocketPath`

internal/overlord/ifacestate/ifacemgr.go
jonathan-conder: I still adjusted the comment to reflect it better

jonathan-conder:
Apart from the naming and comment changes, I removed the `.keep` files. The workshop state dir doesn't need to exist until the first refresh, so we just create it with cloud-init. The guest `bin` directory is always mounted, so the permissions are only observable in very early VM boot.

Also added some escaping when using templated paths inside the cloud-config, and changed `waitready` to relative symlink.

See https://github.com/canonical/workshop/compare/433ae5f3fd5a90de945445daf4b0f446da9cba28..f7ecbd874c49e660c0d3d161b5be540e23dd2214 and https://github.com/canonical/workshop/compare/f7ecbd874c49e660c0d3d161b5be540e23dd2214..c18239f05b1674e40531f2925475eec99957a8ff

internal/dirs/dirs.go
```diff
@@ -36,6 +36,9 @@ var (
 	// base directory inside a workshop
 	WorkshopBaseDir = defaultBaseDir
 
+	// Directory for mounted binaries (i.e. workshopctl)
+	WorkshopGuestBinDir = filepath.Join(WorkshopBaseDir, "bin")
```
dmitry-lyfar, line 40: Down in this file (49) we have `WorkshopSdkStateDir`, maybe drop it too here, seems unused?

dmitry-lyfar:
> Apart from the naming and comment changes, I removed the `.keep` files. The workshop state dir doesn't need to exist until the first refresh, so we just create it with cloud-init. The guest `bin` directory is always mounted, so the permissions are only observable in very early VM boot.
> 
> Also added some escaping when using templated paths inside the cloud-config, and changed `waitready` to relative symlink.
> 
> See https://github.com/canonical/workshop/compare/433ae5f3fd5a90de945445daf4b0f446da9cba28..f7ecbd874c49e660c0d3d161b5be540e23dd2214 and https://github.com/canonical/workshop/compare/f7ecbd874c49e660c0d3d161b5be540e23dd2214..c18239f05b1674e40531f2925475eec99957a8ff

Yep, the cloud-init approach is much nicer now.

PR #944, opened 2026-07-16
Doc: Update 'In this documentation'

akcano:


dmitry-lyfar (approved):
I'm curious about the category navigation, should not it be coming from the sphinx shared repo?

PR #956, opened 2026-07-21
Add Btrfs fallback for other Linux distros

### LINKED ISSUE #742: workshop does not work on non-Ubuntu systems
Hi,

I cannot use workshop on an openSUSE Tumbleweed host which is supported by snapd.

```
maj 13 11:05:04 carrack systemd[1]: snap.workshop.workshopd.service: Scheduled restart job, restart counter is at 4.
maj 13 11:05:04 carrack systemd[1]: Starting Service for snap application workshop.workshopd...
maj 13 11:05:04 carrack workshop.workshopd[87635]: Acquiring state lock file
maj 13 11:05:04 carrack workshop.workshopd[87635]: Acquired state lock file
maj 13 11:05:04 carrack workshop.workshopd[87635]: cannot run workshop: Error loading "zfs" module: Failed to run: modprobe -b zfs: exit status 1 (modprobe: FATAL: Module zfs not found in directory /usr/lib/modules/7.0.3-1-default)
maj 13 11:05:04 carrack systemd[1]: snap.workshop.workshopd.service: Main process exited, code=exited, status=1/FAILURE
maj 13 11:05:04 carrack systemd[1]: snap.workshop.workshopd.service: Failed with result 'exit-code'.
```

I can create LXD container of VM instances just fine. It seems like it's assuming that zfs is always available, and can used for storage which isn't true. If workshop is published through the snap store it should either try to support other systems where snapd can run or should include some sort of disclaimer that only Ubuntu is supported.

```
$ snap list workshop
Name      Version          Rev  Tracking       Publisher  Notes
workshop  0.1.30-a8fb6ac3  126  latest/stable  dlyfar     classic
```

locnnil:
# Description

Enables Btrfs fallback for other Linux distributions.

ZFS is the preferred storage driver, but it isn't available everywhere: WSL, SUSE, Arch, Fedora, and other distributions may ship without it. Keying the fallback off the distribution name is too coarse; it assumes every non-Ubuntu host lacks ZFS and every Ubuntu-derived one has it, neither of which holds for Debian and its derivatives.

Instead, detect whether the host provides a usable ZFS kernel module, already loaded  (present at `/dev/zfs` **AND** `/sys/module/zfs` ) or loadable for the running kernel per [modules.dep(5)](https://man7.org/linux/man-pages/man5/modules.dep.5.html) (`/lib/modules/$(uname -r)/modules.dep`) or modules.builtin (`/lib/modules/$(uname -r)/modules.builtin`) and keep ZFS only then, falling back to btrfs otherwise. Only the kernel module is checked: LXD ships as a snap that bundles the zpool/zfs userspace tools, so a loaded or loadable module is what actually determines usability from the host side, [tested on Arch Linux](https://pastebin.canonical.com/p/xTBBkd9vRT/) and [openSUSE tumbleweed](https://pastebin.canonical.com/p/tn2Hk2vPMG/).

- Closes: https://github.com/canonical/workshop/issues/742

# Self-review quick check

 * [ ] Delete dead code and redundant comments. **TODO**: check if it is okay to delete `isWSL()`
 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:


jonathan-conder:
It looks to me like the way LXD does it is just trying to load the zfs module; maybe we should follow their lead on that

jonathan-conder:
I think we also need to be more careful about where we use the `storagePoolDriver` variable. Currently we can assume it's accurate, since whether the host is WSL or not is unlikely to change. The availability of the ZFS module can easily change, though; we need a separate mechanism to indicate "what's the current driver of the `workshop` pool."

locnnil:
> I think we also need to be more careful about where we use the `storagePoolDriver` variable. Currently we can assume it's accurate, since whether the host is WSL or not is unlikely to change. The availability of the ZFS module can easily change, though; we need a separate mechanism to indicate "what's the current driver of the `workshop` pool."

So, considering the case where ZFS was available when the pool was created but later stops being present, most likely because the user installed a different kernel that doesn't ship the ZFS module. What should we do with the now-unusable ZFS pool? 

I can think of those options:

- Put the daemon into a degraded state and add an actionable error message like "**restore the ZFS-capable kernel, or remove the pool to recreate it as Btrfs**" 
- Warn the user with an actionable message like "Old ZFS pool detected on system without support for ZFS on path `<path-to-zfs-pool>`"  but create a new Btrfs pool under a different name.

I believe that the first option is better, right?

jonathan-conder:
> I believe that the first option is better, right?

I agree, but we should not suggest removing the pool, it will probably break the workshops using it (if it's even possible to remove it). Instead maybe it's best to reinstall the Workshop snap.

We just need to test what actually happens in that scenario, maybe LXD will fail to start and we probably handle that already.

locnnil:
@jonathan-conder I've implemented the requested change in [`54bf6dd`](https://github.com/canonical/workshop/pull/956/commits/54bf6dd01e12bef640096bdec1bb8e08090c95ee), which stops assuming a fixed state for ZFS availability. I then tested the degraded-state behavior for the scenario where ZFS is present at first but later vanishes.

I ran this on an **Ubuntu Resolute LXD VM**, reproducing a `workshop` pool created on ZFS whose kernel module later becomes unavailable after a kernel change.

**Reproduction**

1. Started with ZFS available, so the `workshop` pool was created on the `zfs` driver.
2. Built and installed a `.deb` for a newer kernel (`7.2.0-rc7`) that doesn't ship the ZFS module.
3. Removed the previous kernel and its ZFS module package (`linux-main-modules-zfs-7.0.0-29-generic`).
4. Rebooted into the new kernel and ran `workshop list`.

**Result**

Instead of surfacing a raw LXD error, the daemon goes into degraded mode with an actionable message:

```console
$ workshop list
error: system is not healthy: cannot use the "workshop" storage pool, its storage driver may be unavailable (for example a ZFS pool after booting a kernel without the ZFS module): Error loading "zfs" module: Failed running: modprobe -b zfs
: exit status 1 (modprobe: FATAL: Module zfs not found in directory /lib/modules/7.2.0-rc7)
Boot into a kernel that provides the pool's storage driver, or reinstall Workshop to recreate the pool on an available driver
```

**Recovery**

The ZFS pool can't be deleted through LXD while the module is missing (`lxc storage delete` loads the driver, which fails), so the recovery is to **reinstall Workshop**: this recreates the `workshop` pool on an available driver (Btrfs) for a host that no longer provides ZFS.

**On adding a Spread test**

Let me know if you think we should add a Spread test covering this case. Two caveats:

- For my manual test I had to build and install an upstream kernel, since simply uninstalling `linux-main-modules-zfs-7.0.0-xx-generic` isn't enough to reproduce a realistic scenario. For an automated test we could instead force the broken state, making it reproducible by just forcefully removing a package.
- The bigger drawback is that this would require Spread to run on **LXD VMs** instead of containers. That would be a significant change to the current testing infrastructure.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -59,6 +57,11 @@ const (
 	storagePool           = "workshop"
 	storagePoolMinimalGiB = 5
 
+	// ZFS is workshop's preferred storage driver; Btrfs is the fallback used
+	// on hosts where LXD reports ZFS as unavailable.
+	storageDriverZFS   = "zfs"
+	storageDriverBtrfs = "btrfs"
+
```
jonathan-conder: I don't see what we gain from these constants

internal/workshop/lxd/lxd_backend.go
```diff
@@ -251,19 +164,56 @@ func checkVersion(version string) error {
...
+// mode. When the pool does not exist yet, it ensures a supported driver is
+// available to create it with.
+func checkStoragePool(conn lxd.InstanceServer, supported []api.ServerStorageDriverInfo) error {
+	_, _, err := conn.GetStoragePool(storagePool)
+	if err == nil {
 		return nil
 	}
+	if !api.StatusErrorCheck(err, http.StatusNotFound) {
+		return fmt.Errorf(`cannot use the %q storage pool, its storage driver may be unavailable (for example a ZFS pool after booting a kernel without the ZFS module): %w
+Boot into a kernel that provides the pool's storage driver, or reinstall Workshop to recreate the pool on an available driver`, storagePool, err)
```
jonathan-conder, line 230: is the error returned from `GetStoragePool` related to this explanation?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -251,19 +164,56 @@ func checkVersion(version string) error {
...
+// checkStoragePool verifies the workshop storage pool is usable. When the pool
+// exists but cannot be loaded, most likely because its storage driver's kernel
+// module is gone (e.g. a ZFS pool after booting a kernel without the ZFS
+// module), it returns an actionable error that puts the daemon into degraded
+// mode. When the pool does not exist yet, it ensures a supported driver is
+// available to create it with.
+func checkStoragePool(conn lxd.InstanceServer, supported []api.ServerStorageDriverInfo) error {
+	_, _, err := conn.GetStoragePool(storagePool)
+	if err == nil {
 		return nil
```
jonathan-conder, line 226: do we need to check if the storage pool driver is supported here?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -356,15 +306,22 @@ func ensureBackendReady() error {
 	}
 	defer conn.Disconnect()
 
-	// Create LXD storage pool if it doesn't exist.
-	pools, err := conn.GetStoragePools()
+	// Create LXD storage pool if it doesn't exist. GetStoragePoolNames avoids
+	// loading each pool's driver, so an existing but currently-unloadable pool
+	// (e.g. a ZFS pool whose module is gone) is reported by checkStoragePool
+	// rather than failing here.
+	names, err := conn.GetStoragePoolNames()
```
jonathan-conder: do we need to list all the storage pools here? what stops us querying `workshop` directly?

internal/workshop/lxd/lxd_backend_snapshots.go
```diff
@@ -346,7 +346,11 @@ func (s *Backend) TakeSnapshot(ctx context.Context, name string, snapshot worksh
 	if inst.Devices == nil {
 		inst.Devices = map[string]map[string]string{}
 	}
-	if err := mergeDevices(inst.Devices, snapshot.Sdks, name); err != nil {
+	usesZFS, err := poolUsesZFS(conn)
```
jonathan-conder, line 349: this adds an additional API call to these operations, do we have any info on the impact it has on latency? we could alternatively cache the storage pool driver at launch

jonathan-conder:
> The ZFS pool can't be deleted through LXD while the module is missing (`lxc storage delete` loads the driver, which fails), so the recovery is to **reinstall Workshop**: this recreates the `workshop` pool on an available driver (Btrfs) for a host that no longer provides ZFS.

I don't see how reinstalling Workshop will help. If LXD is unable to delete the pool then it will still be there after the reinstall.

> For my manual test I had to build and install an upstream kernel, since simply uninstalling `linux-main-modules-zfs-7.0.0-xx-generic` isn't enough to reproduce a realistic scenario. For an automated test we could instead force the broken state, making it reproducible by just forcefully removing a package.

It's not reasonable to do that in a test. It's a rare enough case that I think some manual recovery steps are OK.

> The bigger drawback is that this would require Spread to run on **LXD VMs** instead of containers. That would be a significant change to the current testing infrastructure.

we already do this, it's not a change at all

internal/workshop/lxd/lxd_backend.go
locnnil: Yes, from my tests, this is what you get:

```console
$ workshop list
error: system is not healthy: cannot use the "workshop" storage pool, its storage driver may be unavailable (for example a ZFS pool after booting a kernel without the ZFS module): Error loading "zfs" module: Failed running: modprobe -b zfs
: exit status 1 (modprobe: FATAL: Module zfs not found in directory /lib/modules/7.2.0-rc7)
Boot into a kernel that provides the pool's storage driver, or reinstall Workshop to recreate the pool on an available driver

```

So the error returned by `GetStoragePool` is:

`Error loading "zfs" module: Failed running: modprobe -b zfs
: exit status 1 (modprobe: FATAL: Module zfs not found in directory /lib/modules/7.2.0-rc7)`

Perhaps the explanation should be less verbose? Since the error is already quite informative.

jonathan-conder: That's interesting and a bit unexpected. I'd rather not rely on specific LXD error messages if possible, it's likely not something they have tests for. How about checking `info.Environment.StorageSupportedDrivers` like you did below?

The error message is useful though, does it appear in the LXD logs? Maybe we can direct users there

locnnil: >  I'd rather not rely on specific LXD error messages if possible

Good point, I'll keep my explanation rather than trimming it in favor of LXD's message.

> How about checking info.Environment.StorageSupportedDrivers like you did below?

On using `info.Environment.StorageSupportedDrivers` for this check, though, I ran into a wall: to know whether the *existing* `workshop` pool is affected, we need its driver, and every call that returns it loads that driver's module. `GetStoragePool` (and `GetStoragePools` with recursion) goes through LXD's `LoadByName` -> `drivers.Load` -> `modprobe`, which is the exact call that fails once the module is gone (that's why my test showed `GetStoragePool` erroring). `GetStoragePoolNames` avoids the load, but only returns names, not drivers.

So `StorageSupportedDrivers` works for *creation* (picking a driver we already know is available, as below), but on its own it can't tell us whether the *already-created* pool's driver is still supported, since we can't read that pool's driver without triggering the failing load.

On the logs: LXD only logs the "driver not available" line at `Debug` level (in `SupportedDrivers`), so it won't appear in default logs, directing users that there wouldn't be reliability unless they first raise LXD's log level.

A couple of options that I could think of:

1. **Keep the status-based detection.** We branch on the 404 vs. non-404 status from `GetStoragePool`, not on the message text; the LXD error is only surfaced through `%w`. I'd just keep our wording neutral.

2. **Persist the pool's driver at creation**, then the health check compares that stored value against `StorageSupportedDrivers` without ever calling `GetStoragePool`. This would also address your caching comment on the snapshot path.

I prefer (1) for this PR as the smaller change, but (2) is cleaner if we want to avoid depending on the LXD error at all. What do you think?

jonathan-conder: We should definitely keep branching on 404. I would do the following:

1. Call `GetStoragePool("workshop")` (pseudocode). If that fails for any reason other than 404, report the error and give up.
2. If it succeeds, record its driver.
3. If it 404s, get the `info` and pick a driver from `info.Environment.StorageSupportedDrivers`.

internal/workshop/lxd/lxd_backend_snapshots.go
locnnil: With this change https://github.com/canonical/workshop/pull/956/commits/cbe0b61169b13779c90da6ecf84f4b3a6501a909 `poolUsesZFS` reads from a cached value.

locnnil:
> I don't see how reinstalling Workshop will help. If LXD is unable to delete the pool then it will still be there after the reinstall.

Yeah, I think in this case you would need to reinstall LXD and Workshop, but this seems too drastic. Anyway, I've tested, and when the ZFS-capable kernel is restored, the `workshop` will be working again.

docs/explanation/architecture/components.rst
```diff
@@ -258,8 +258,9 @@ used for temporary storage during workshop rebuild operations.
 Storage backends
 ~~~~~~~~~~~~~~~~
 
-|ws_markup| uses ZFS for storage on Linux,
-with automatic Btrfs fallback on Windows Subsystem for Linux (WSL).
+|ws_markup| uses ZFS for storage where it detects that ZFS is available,
+with automatic Btrfs fallback otherwise,
```
jonathan-conder, line 262:
```suggestion
with automatic btrfs fallback otherwise,
```

It looks like upstream alternates between BTRFS and btrfs, what do you think @akcano?

internal/workshop/lxd/lxd_backend.go
```diff
@@ -164,19 +160,77 @@ func checkVersion(version string) error {
...
+}
+
+// preferredDriver picks the driver for a new workshop pool: ZFS when LXD
+// reports it as supported, otherwise Btrfs.
+func preferredDriver(supported []api.ServerStorageDriverInfo) string {
+	if driverSupported(supported, "zfs") {
+		return "zfs"
+	}
+	return "btrfs"
+}
```
jonathan-conder, line 179: instead of factoring out a function with a single callsite, I would turn `preferredDrivers` into a list and loop over it, in a single function. If none of the preferred drivers are supported we should return an error.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -164,19 +160,77 @@ func checkVersion(version string) error {
...
+
+// poolDriverCache stores the driver of the workshop storage pool once known.
+// ensureBackendReady records it, and pool operations reuse it instead of
+// re-querying LXD (which reloads the driver's module). The pool's driver is
+// fixed at creation, so the cached value stays accurate for the daemon's run.
+var poolDriverCache struct {
+	sync.RWMutex
```
jonathan-conder, line 186: do we need a mutex here? I think anything that looks at the storage driver should run after the value has been initialised.

internal/workshop/lxd/lxd_backend.go
```diff
@@ -164,19 +160,77 @@ func checkVersion(version string) error {
...
+		pool, _, err := conn.GetStoragePool(storagePool)
+		if err != nil {
+			return false, err
+		}
+		driver = pool.Driver
+		setPoolDriver(driver)
 	}
-	if slices.ContainsFunc(drivers, hasDriver) {
+	return driver == "zfs", nil
+}
```
jonathan-conder, line 216: I also don't see a need for lazy initialisation

internal/workshop/lxd/lxd_backend.go
```diff
@@ -164,19 +160,77 @@ func checkVersion(version string) error {
...
+
+// checkStoragePool verifies the workshop storage pool is usable. When the pool
+// exists but cannot be loaded, most likely because its storage driver's kernel
+// module is gone (e.g. a ZFS pool after booting a kernel without the ZFS
+// module), it returns an actionable error that puts the daemon into degraded
+// mode. A missing pool is not an error here: ensureBackendReady creates it.
```
jonathan-conder, line 222: what is the point of this comment?

jonathan-conder (changes requested):

PR #957, opened 2026-07-22
Doc: Handle orphaned workshops

akcano:


dmitry-lyfar (approved):
This should be addressed in the product and not in documentation. But as we don't have a solution to it implemented yet, e.g. allowing `workshop remove dev.project.wp` or surfacing full name like `dev-45fg23df`  and allowing `workshop remove dev-45fg23df`, I'll merge it for now.

jonathan-conder:
I think we should just allow `workshop remove --project /path dev` even if `/path` doesn't exist

PR #962, opened 2026-07-24
Send workshop machine ID with workshopctl requests

tlm:
# Description

Forward the Workshop instance ID from workshopctl to workshopd in the workshop-instance-id HTTP header.

The value currently comes from /etc/machine-id, which Workshop populates from the LXD instance UUID. The source is an implementation detail; the HTTP contract identifies the Workshop instance that originated the request.

This enables follow-up secrets work: workshopctl and workshopd need a shared Workshop instance identity to communicate and validate which Workshop a secret request originates from.

This PR only transports the identifier through the existing private container-to-host Unix-socket path. It does not validate the identifier or authorise secret access; those checks will be added in follow-up secrets PRs. The identifier is not written to daemon logs.

# Self-review quick check

## Docs


jonathan-conder:
just FYI from `machine-id(5)`:

> This ID uniquely identifies the host. It should be considered "confidential", and must not be exposed in untrusted environments, in particular on the network. If a stable unique identifier that is tied to the machine is needed for some application, the machine ID or any part of it must not be used directly. Instead the machine ID should be hashed with a cryptographic, keyed hash function, using a fixed, application-specific key. That way the ID will be properly unique, and derived in a constant way from the machine ID but there will be no way to retrieve the original machine ID from the application-specific one. The sd_id128_get_machine_app_specific(3) API provides an implementation of such an algorithm.

I don't think the socket necessarily counts as the "network," but it seems like a good idea regardless.

tlm:
@jonathan-conder good spot.

I looked into the available options. For Workshop LXD containers,
`/etc/machine-id` is derived from the instance’s `volatile.uuid`. The
host-side LXD API already exposes that authoritative value to `workshopd`,
which Workshop uses when configuring the instance template used during
creation and restore.

In this request path, `workshopctl` is not disclosing the value to a new
network or untrusted service: the container already has the machine ID, and
`workshopd` already has access to the corresponding LXD instance UUID. The
request travels over the existing private Unix-socket proxy. I have also
removed it from daemon debug logging so it is not retained there.

`workshopd` already has the authoritative raw value through LXD, so deriving
an HMAC would not hide information from the request recipient. It would only
replace one representation of an identifier that `workshopd` can already
derive with another representation.

A client-side fixed HMAC key also does not add an authentication boundary:
processes in the workshop can read the machine ID and inspect the
`workshopctl` binary containing the key. `workshopd` could recompute the
expected HMAC and compare it, but could not use it to recover a machine ID it
did not already obtain from LXD.

I am inclined to retain the raw value for this private host/container
validation path. If a future secrets flow sends this identity beyond that
boundary, we should introduce an application-specific derived identifier at
that point. I am open to stricter alternatives or a different threat model.

dmitry-lyfar (approved):

PR #976, opened 2026-08-12
Move workshopd run directory to SNAP_COMMON

dmitry-lyfar:
# Description

Having the run directory in `SNAP_DATA` is problematic because it gets snap revision in its path. Thus, artifacts like Xauthority become unavailable after `snap refresh` cleans up the old snap's revisions. It breaks workshops that connect a desktop interface plug as the mount cannot find the source. Moving it to `SNAP_COMMON` keeps the data shared among the snap instances.

The `run` directory maintains locks for SDKs too. I could not identify any issues with these being moved to the common location either. Locks that are held at the refresh time will be cleared by the kernel and recaptured if needed on the daemon restart. The only cosmetic effect will be that the lock files will not be cleared on the system's reboot automatically as it would be expected from a run location.

## Known problems

Existing workshops using the desktop interfaces will have to be re-launched. 

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```

## Docs

Procedure:

Content:

---

Or:

### DIFF
--- a/snap/local/hooks/post-refresh
+++ b/snap/local/hooks/post-refresh
@@ -2,3 +2,4 @@
 
 # TODO: remove a one-off data -> common migration cleanup
 rm -rf "${SNAP_DATA}/ssh"
+rm -rf "${SNAP_DATA}/run"
--- a/internal/dirs/dirs.go
+++ b/internal/dirs/dirs.go
@@ -152,8 +152,6 @@ func setDataDir(dataDir string) {
 	DataDir = dataDir
 
 	WorkshopStateLockFile = filepath.Join(DataDir, "state.lock")
-	WorkshopdRunDir = filepath.Join(DataDir, "run", "workshopd")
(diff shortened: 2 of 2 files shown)


snap/local/hooks/post-refresh
```diff
@@ -2,3 +2,4 @@
 
 # TODO: remove a one-off data -> common migration cleanup
 rm -rf "${SNAP_DATA}/ssh"
+rm -rf "${SNAP_DATA}/run"
```
tlm, line 5: Does the TODO above need to be dealt with ?

tlm (approved):
lgtm

snap/local/hooks/post-refresh
dmitry-lyfar: Yes, TODO is for both of these. It was only just introduced in the recent PR, still have to be there.

PR #978, opened 2026-08-12
Update copilot instructions; add knowledge base

dmitry-lyfar:
# Description

Aiming two goals here (WSP-1054):

- Start a knowledge base that describes invariants we keep in mind as maintainers but agents have to rediscover every time as these are implicit. Keep it to the minimal, any invariant has to earn its place in the KB. I imagine it as an additional context for planning, writing tests, discovery, experiments. The daemon was the first candidate that had some of these. The LXD backend is another source of implicit knowledge and gotchas.
- Update copilot-instructions to leave only high-signal instructions (based on my and llm's review).

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,106 +1,27 @@
 # GitHub Copilot Instructions
 
-This file provides general project context for GitHub Copilot. For review-specific guidance, see:
+For review-specific guidance, see:
+
 - **Code Review**: `.github/agents/code-review.agent.md`
 - **Documentation Review**: `.github/skills/documentation-review/SKILL.md`
 
 ## Project Overview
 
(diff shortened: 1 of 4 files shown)


tlm (commented):
My Two Cents:

- I think delete the copilot instructions and move to a root `AGENTS.md`
- Would it be better to move things like daemon.md to another AGENTS.md in the respective file. Or package it up as developer docs that we can point an agent to about overall design?

Other then that I like it I can see the intent that this is trying to move towards. I like removing things like the tech stack etc. They're low value and describe properties where it's best to describe use and how to drive for which the properties can be assumed off of.

akcano (approved):
I like this direction very much; one thing I'd suggest adding, perhaps as a follow-up, is a skill that will update the knowledge base with items of expertise and wikify it for progressive disclosure.

PR #982, opened 2026-08-14
Skip tics reporting for PR test runs

dmitry-lyfar:


.github/workflows/cover.yaml
jonathan-conder: this looks unnecessary now

.github/workflows/cover.yaml
```diff
@@ -36,6 +36,10 @@ jobs:
 
   code-coverage:
     needs: [spread, unit-tests]
+    # only update the coverage report on main branch pushes, not PRs or
+    # scheduled runs as reporting takes up to 10 mins per run and does not
+    # provide much benefit for PRs.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
tlm: would it be worth while just pulling this job out into it's own file and then the if statement can go away as the on clause correctly states what will trigger it?

It would also mean that the spread jobs can have their own cadence.

tlm (approved):

.github/workflows/cover.yaml
dmitry-lyfar: Yes, I switched it to run on schedule after our periodic tests run and renamed the workflow file, I think it should be `automatic-tests.yaml` instead of `cover.yaml`.

PR #988, opened 2026-08-17
lxd: install secret socket systemd units on workshop creation

tlm:
## Summary

First building block of the secrets interface spec: install the systemd
socket that SDK units will use to consume secrets via `LoadCredential`.

When a workshop is created, cloud-init now installs two units:

- `workshop-secret.socket` — listens on
  `/var/lib/workshop/run/workshop.socket.secret` (mode 0660, `Accept=yes`),
  wanted by `sockets.target`.
- `workshop-secret@.service` — a per-connection socket-activated resolver,
  running `workshopctl get-secret --systemd`.

The socket is enabled and started on first boot. Each credential request is
accepted by systemd and passed to a resolver instance on standard input and
output. Resolver diagnostics go to the journal so only credential bytes are
returned on the connection.

The `get-secret` subcommand itself is added in a follow-up; nothing connects
to the socket yet, so this change is inert until then.

## Refactoring

The cloud-init template was a ~100-line string literal embedded in
`workshopConfig`, which made the function hard to read and the template hard
to review in isolation. It is now:

- extracted into a dedicated `cloudconfig.go`,
- parsed once via `sync.OnceValues` instead of on every workshop creation,
- returned as an error rather than panicking via `template.Must`,
- parameterised by a `cloudConfigVars` struct so paths come from the `dirs`
  package instead of being hard-coded in the template text.

`dirs.WorkshopSecretSocketPath` is defined as the single canonical source for
the socket path, which the upcoming daemon-side secret resolution will also
need.

## Snapshot compatibility

The LXD snapshot format revision is bumped to 12 because cloud-init now
creates persistent secret socket and resolver unit files in the workshop
rootfs. The volatile runtime secret socket is excluded from snapshot rootfs
comparisons, consistent with the existing untrusted socket.

## Testing

- `TestWorkshopConfigSecretSocketUnits` — asserts the units and the enable
  command land in generated cloud-init user-data.
- `TestCloudConfigTemplateParses` — the embedded template compiles.
- `TestCloudConfigTemplateRendersVars` — template variables are interpolated
  into rendered output.
- `TestCloudConfigTemplateCached` — the template is parsed once and reused.
- `go test ./internal/workshop/lxd`

## Follow-ups

- `workshopctl get-secret` subcommand (plain and `--systemd` modes).
- Daemon-side secret resolution.
- `SDK_SYSTEMD_SECRET_SOCKET` env var for SDK hooks.

### DIFF
--- a/internal/dirs/dirs.go
+++ b/internal/dirs/dirs.go
@@ -54,6 +54,10 @@ var (
 	// socket name (which varies, e.g. under "go tool try").
 	WorkshopSocketPath = filepath.Join(WorkshopRunDir, "workshop.socket")
 
+	// Path to the unix socket inside a workshop that systemd units connect
+	// to for secret resolution (via LoadCredential).
+	WorkshopSecretSocketPath = filepath.Join(WorkshopRunDir, "workshop.socket.secret")
+
 	// Directory for actions inside workshop
 	WorkshopActionsDir = filepath.Join(WorkshopRunDir, "actions")
 
--- a/internal/workshop/lxd/cloudconfig.go
+++ b/internal/workshop/lxd/cloudconfig.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2026 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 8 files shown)


jonathan-conder:
I like the refactor but it would be easier to review if split into its own commit

dmitry-lyfar:
@tlm the test seem to be failing, otherwise looks good. agree, we should have it as a separate commit if possible,

jonathan-conder:
I guess the test is indicating that the `FormatRevision` needs updating

dmitry-lyfar (approved):

PR #1002, opened 2026-08-24
Give duck-typed interface hooks named contracts

tlm:
# Description

Groundwork for the secrets work: the secrets spec introduces a new
`secret` interface whose plugs and slots are resolved at request time
through the workshop secret socket and daemon, not by configuring the
container. It will be the first built-in interface with no LXD device
backend hooks, which the existing interface tests made impossible to
register: TestEachInterfaceImplementsSomeBackendMethods required
every interface to implement at least one backend hook, and the only
ways through were a misleading no-op hook or an exclusion mechanism.

Inspecting those tests showed they were low value beyond that
roadblock:

- TestEachInterfaceImplementsSomeBackendMethods only required one
  recognised hook to match, so an interface expected to provide
  several hooks still passed with a typo in one of them. Its
  recognised-hook list was also a hand-maintained copy of the
  backend's anonymous interface types and had already drifted,
  blessing two permanent hooks that the backend never dispatched.
- TestUnexpectedSpecSignatures never asserted anything: it composed
  expected method names from the security system name, producing
  lookups like "lxd-devicePermanentPlug" that cannot match a Go
  method. It passed for its entire life without checking a single
  signature.

Both are removed. In their place the hook contracts become real,
named types with a single source of truth:

- Export the hook interfaces from lxd_device, which owns the hook
  vocabulary (ConnectedPlugMounter, ConnectedSlotMounter,
  PermanentPlugMounter, PermanentSlotMounter), and dispatch the
  Specification Add* methods through them, so the published contract
  and the dispatch are the same type. This also wires up the
  permanent plug and slot hooks, which were previously ignored
  no-ops; nothing implements them today, so behaviour is unchanged.
- Export a StaticInfoProvider from interfaces, asserted against in
  StaticInfoOf.

A contract conformance suite was added and then removed within this
PR: asserting each interface's method set restates the implementation
without exercising behaviour, which each interface's own behaviour
tests already cover by invoking the backend dispatch and checking the
resulting specification.

# Self-review quick check

## Docs


jonathan-conder:
Want to get Dmitry's opinion on this. If we're going to make core changes to the interface system I also want to start using `encoding/json` to unmarshal the plug and slot attributes.

internal/interfaces/builtin/contract_test.go
```diff
@@ -0,0 +1,127 @@
...
+// TestCameraImplementsContract verifies that the camera interface
+// satisfies its expected contracts.
+func (s *contractSuite) TestCameraImplementsContract(c *check.C) {
+	// The full set of contracts the camera interface must satisfy.
+	type cameraContract interface {
+		interfaces.Interface
+		interfaces.StaticInfoProvider
+		lxd_device.MountConnectedPlugDefiner
+	}
+	var iface cameraContract = &cameraInterface{}
```
jonathan-conder: I think this can be arranged to fail at compile time, no need for a test

internal/interfaces/lxd_device/definer.go
```diff
@@ -0,0 +1,62 @@
...
+// to a workshop's LXD device configuration when a plug of the interface
+// is connected. The backend detects the method by type assertion; an
+// interface that does not implement it contributes nothing.
+type MountConnectedPlugDefiner interface {
+	MountConnectedPlug(
+		spec *Specification,
+		plug *interfaces.ConnectedPlug,
+		slot *interfaces.ConnectedSlot,
+	) error
+}
```
jonathan-conder: isn't there some place this can be used, similar to what you did with StaticInfoProvider?

internal/interfaces/lxd_device/definer.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"github.com/canonical/workshop/internal/interfaces"
+	"github.com/canonical/workshop/internal/sdk"
+)
+
+// MountConnectedPlugDefiner is implemented by interfaces that contribute
+// to a workshop's LXD device configuration when a plug of the interface
+// is connected. The backend detects the method by type assertion; an
+// interface that does not implement it contributes nothing.
+type MountConnectedPlugDefiner interface {
```
jonathan-conder:
```suggestion
type ConnectedPlugMounter interface {
```

internal/interfaces/builtin/contract_test.go
tlm: Yeah. The thinking was to get the failure down to a named test so purpose could be explained. After speaking with @dmitry-lyfar we have just removed the tests all together. They are very low value and don't make a great deal of sense.

PR updated.

internal/interfaces/lxd_device/definer.go
tlm: Yep done & done

internal/interfaces/lxd_device/definer.go
tlm: Good suggestion. Have updated all of them. Please have a look.

internal/interfaces/lxd_device/spec.go
```diff
@@ -93,32 +93,31 @@ type Specification struct {
 
 // AddPermanentSlot records side-effects of having a slot.
 func (s *Specification) AddPermanentSlot(iface interfaces.Interface, slot *sdk.SlotInfo) error {
+	if iface, ok := iface.(PermanentSlotMounter); ok {
+		return iface.MountPermanentSlot(s, slot)
+	}
```
jonathan-conder, line 98: I guess this is dead code? Given that it already was, I don't mind too much, but would be good to know what your reasoning is for adding these.

jonathan-conder (approved):
Looks good now. I haven't seen some of those tests before, makes a lot of sense to remove them

dmitry-lyfar (approved):

internal/interfaces/lxd_device/spec.go
tlm: Yeah it is. I was just avoiding making the decision on it as it has been implemented on purpose at some stage.

PR #1004, opened 2026-08-26
Add the builtin secret interface

tlm:
# Description

First step of the secrets spec on the workshop side: register a
builtin secret interface so workshopd accepts SDK definitions and
workshop files declaring secret plugs and slots.

The interface behaviour follows the spec:

- Plugs may only be declared by regular SDKs, expressing an SDK's
  need for a secret. Slots may only be declared by the system SDK, as
  they carry the host keyring lookup details for the workshop.
- Secrets must always be attached by an explicit user decision: both
  the code-level AutoConnect veto and the base declaration deny
  auto-connection. Established connections are preserved across
  refreshes by the existing connection stickiness machinery.
- The interface has no backend hooks. Secret values are resolved at
  request time via the workshop secret socket and are never
  configured into the container.

A spread test launches a workshop declaring both sides, checks
neither auto-connects, connects and disconnects manually, and checks
the connection survives a refresh. Both sides are declared in the
workshop definition rather than an SDK: sdkcraft does not yet know
the secret interface and rejects it when packing, so extending
sdkcraft is a prerequisite for the SDK-side examples in the spec.

Not yet covered, coming next:

- Slot attributes (`collection`, `attributes`) are not yet accepted:
  the interface has no BeforePrepareSlot sanitizer, so attribute
  validation and the spec's keyring lookup details are follow-up
  work.
- Interface reference documentation will land with the feature when
  it merges to main.

Also includes a small cleanup replacing the byIfaceName sort.Interface
boilerplate with a CompareByName comparison function.

# Self-review quick check

## Docs

### DIFF
--- a/internal/interfaces/policy/basedeclaration_test.go
+++ b/internal/interfaces/policy/basedeclaration_test.go
@@ -167,6 +167,7 @@ func (s *baseDeclSuite) TestAutoConnectPlugSlot(c *check.C) {
 	// these have more complex or in flux policies and have their
 	// own separate tests
 	snowflakes := map[string]bool{
+		"secret": true,
 		"tunnel": true,
 	}
 
--- a/internal/interfaces/policy/basedeclaration_test.go
+++ b/internal/interfaces/policy/basedeclaration_test.go
@@ -167,7 +167,6 @@ func (s *baseDeclSuite) TestAutoConnectPlugSlot(c *check.C) {
 	// these have more complex or in flux policies and have their
 	// own separate tests
 	snowflakes := map[string]bool{
-		"secret": true,
 		"tunnel": true,
 	}
 
--- a/internal/interfaces/builtin/all.go
+++ b/internal/interfaces/builtin/all.go
@@ -21,7 +21,8 @@ package builtin
 
 import (
 	"fmt"
-	"sort"
+	"maps"
+	"slices"
 
 	"github.com/canonical/workshop/internal/interfaces"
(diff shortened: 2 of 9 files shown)


internal/interfaces/policy/basedeclaration_test.go
```diff
@@ -167,6 +167,7 @@ func (s *baseDeclSuite) TestAutoConnectPlugSlot(c *check.C) {
 	// these have more complex or in flux policies and have their
 	// own separate tests
 	snowflakes := map[string]bool{
+		"secret": true,
```
dmitry-lyfar: What is the reason here? I think the desktop interface has a very similar policy but it's not in the list.

tlm: it's just the way this test is written. It's testing auto connection. I was close to re-doing the testing so it's better but left it alone for the moment. Happy to put up another PR to fix this testing.

dmitry-lyfar: Ah. No, it's OK to leave it as is for now but AutoConnect needs to return true. I think it tests that the `AutoConnect` implementation not overriding the interface's policy and, thus, it should return true for the `secret` interface (e.g. the desktop interface). If you return true there, it does not make the interface auto-connectable.

tlm: All updated.

dmitry-lyfar (approved):

PR #1008, opened 2026-08-30
Rework release workflow to run on signed version tags

dmitry-lyfar:
# Description

- Extract the reference CLI documentation as a separate workflow.
- Run the release workflow on a release PR merge which follows by a signed version tag on main. The release PR includes release notes and a version bump.
- Run tag check for the signed version tags.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,9 +1,41 @@
 name: Release
 on:
-  workflow_dispatch:
+  push:
+    tags:
+      - 'v[0-9]*'
+
+concurrency:
+  group: release
+  cancel-in-progress: false
 
 jobs:
+  verify-tag:
+    environment: release
+    permissions:
+      contents: read
+    runs-on:
+      - self-hosted
+      - Linux
+      - noble
+      - X64
+      - medium
+    steps:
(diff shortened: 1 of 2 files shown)


jonathan-conder (approved):

.github/workflows/release.yaml
```diff
@@ -1,9 +1,40 @@
...
+      - X64
+      - medium
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Verify signed tag
+        env:
+          MAINTAINERS_SSH_PUBKEYS: ${{ vars.MAINTAINERS_SSH_PUBKEYS }}
```
jonathan-conder, line 29: It looks like I can modify this one, in case you wanted to keep it restricted to yourself

dmitry-lyfar: hmm, that's weird, do you mean you can modify the variable in the repo's settings?

jonathan-conder: I think so, didn't actually try it though

dmitry-lyfar: Good catch, variables are modifiable by "Write:" but variables in an environment -- by "Admin:", updated.

PR #1014, opened 2026-09-01
Fix create github release workflow

dmitry-lyfar:
`gh release create` fails if executed outside of the checked out repository, apparently.


jonathan-conder:
can't you just pass `--repo` to `gh`? commit is unsigned btw

jonathan-conder (approved):

PR #1017, opened 2026-09-03
Add workshop identity to workshopctl requests

tlm:
# Description

Lay the groundwork for associating workshopctl requests with the workshop instance from which they originated.

Add user-scoped project lookup to workshop backends and expose the LXD instance UUID on loaded workshop metadata. The UUID is normalized to match the value written to `/etc/machine-id`.

Add request middleware that copies the optional `workshop-instance-id` header into the request context, and apply it to the workshopctl endpoint.

This change only propagates workshop identity. Validation of the supplied instance ID will be added separately.

## LXD integration test identity

Update the LXD integration fixture to use the test runner's numeric UID and GID rather than assuming both are `1000`. Workshop instance configuration uses these values in `raw.idmap` to map the owning host user to the fixed workshop user inside the container.

The previous fixture worked only on hosts where the test runner happened to use UID and GID `1000`. On other hosts, those IDs can belong to different accounts or groups, causing LXD container creation to fail with `operation not permitted`. Using `user.Current()` makes the mapping represent the user actually running the test while retaining the synthetic `testuser` name and isolated LXD project. This is a test portability fix and does not change production identity handling.

## Testing

- `go test ./internal/workshop/fakebackend ./internal/workshop/lxd ./internal/overlord/workshopstate`
- `go test ./internal/daemon -check.f TestWithWorkshopInstanceID`
- `go test -v -tags integration ./internal/workshop/lxd/tests/integration -check.v -check.f TestLxdBackendWorkshopInstanceID`

## Docs

- [x] I confirm the PR has no implications for documentation.

### DIFF
--- a/internal/workshop/lxd/lxd_backend_project.go
+++ b/internal/workshop/lxd/lxd_backend_project.go
@@ -169,7 +169,7 @@ func (s *Backend) CreateOrLoadProject(ctx context.Context, path string) (*worksh
 
 func (s *Backend) Projects(ctx context.Context) (map[string][]workshop.Project, error) {
 	if user, ok := ctx.Value(workshop.ContextUser).(string); ok {
-		projects, err := s.userProjects(ctx)
+		projects, err := s.UserProjects(ctx)
 		if err != nil {
 			return nil, err
 		}
@@ -203,7 +203,7 @@ func (s *Backend) Projects(ctx context.Context) (map[string][]workshop.Project, 
 		}
 
 		prjctx := context.WithValue(ctx, workshop.ContextUser, username)
-		projects, err := s.userProjects(prjctx)
+		projects, err := s.UserProjects(prjctx)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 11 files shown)


dmitry-lyfar:
@tlm I can review it, but I think the tests are failing

internal/workshop/lxd/lxd_backend_project.go
```diff
@@ -213,7 +213,11 @@ func (s *Backend) Projects(ctx context.Context) (map[string][]workshop.Project,
 	return allProjects, nil
 }
 
-func (s *Backend) userProjects(ctx context.Context) ([]workshop.Project, error) {
+func (s *Backend) UserProjects(ctx context.Context) ([]workshop.Project, error) {
+	if _, ok := ctx.Value(workshop.ContextUser).(string); !ok {
+		return nil, nil
```
dmitry-lyfar, line 218:
```suggestion
		return []workshop.Project{}, nil
```
The interface's comment says it would return an empty slice, though, it's probably equivalent behavior here. I'm guessing the difference will show up if this would make it as a return value for `/v1/projects` (not quite likely either given it would be transformed into an API level struct).

dmitry-lyfar (approved):

internal/workshop/lxd/lxd_backend_project.go
tlm: I would be inclined to leave it as is. `len(nil)` still returns 0 and append operations still work. i.e nill is safe value for slices in go.

PR #1019, opened 2026-09-04
Workshopctl instance context

tlm:
# Description

This PR is stacked on #1017, which adds the authenticated user and workshop instance ID to the HTTP request context.

It propagates that request context through `workshopctl` command execution and resolves an appropriate hook context for both hook-driven and ordinary workshop requests. This prepares `get-secret` to determine securely which workshop made a request without coupling command execution directly to the HTTP layer.

## Context propagation

The internal command interface now accepts a standard `context.Context`, and the workshopctl endpoint passes `r.Context()` into command execution.

Request identity belongs in `context.Context` because it is request-scoped metadata. Passing it this way avoids adding workshop identity parameters to every command and allows future commands to consume request metadata without further changing the command interface.

`go-flags` only invokes commands implementing `Execute([]string) error`. Since that signature cannot carry a context, the parser now uses a command handler that calls the project's context-aware command interface explicitly.

## Hook context resolution

Workshopctl commands still require a `hookstate.Context`, but requests can originate through two different paths.

When a request supplies a hook cookie, the handler resolves the existing active hook context. This preserves the established behavior for commands executed during a hook, including access to the hook task and its context data.

Ordinary SDK wrapper calls do not execute inside a hook and therefore do not have a hook cookie. For these requests, the handler uses the workshop instance ID propagated by #1017.

The instance ID is not trusted by itself. Before creating a context, the workshop manager verifies that the authenticated user owns a workshop with that instance ID. This prevents a caller from presenting another workshop's identifier and using it to obtain that workshop's context.

After ownership is established, the hook manager creates a taskless ephemeral context. This provides the interface expected by workshopctl commands without fabricating a hook task or persisting context data beyond the request.

The cookie and instance-ID paths are kept in separate helpers so their different trust models and lifecycle requirements remain explicit.

## Error handling

Unknown or unowned instance IDs are rejected without creating a context.

Backend and state lookup failures are logged server-side, while callers receive a generic internal error response. Returning the underlying error could expose project names, filesystem details, state contents, or other implementation information through an untrusted endpoint.

## Tests

Tests cover:

- Propagation of request-scoped values into command execution.
- Resolution of existing cookie-backed hook contexts.
- Cookie-less resolution using an owned workshop instance ID.
- Rejection of missing, unknown, and unowned instance IDs.
- Workshop ownership lookup across the user's projects.
- Creation of taskless ephemeral contexts.
- Logging and sanitization of internal lookup failures.
- Help requests relying on the propagated workshop identity.

# Self-review quick check

## Docs

### DIFF
--- a/internal/daemon/api_workshopctl.go
+++ b/internal/daemon/api_workshopctl.go
@@ -62,15 +62,15 @@ func v1PostWorkshopCtl(c *Command, r *http.Request, _ *userState) Response {
 
 	// Ignore missing context error to allow 'workshopctl -h' without a context;
 	// Actual context is validated later by get/set.
-	context, _ := c.d.overlord.HookManager().Context(reqData.ContextID)
+	hookContext, _ := c.d.overlord.HookManager().Context(reqData.ContextID)
 
 	if reqData.Stdin != nil {
-		context.Lock()
-		context.Set("stdin", reqData.Stdin)
-		context.Unlock()
+		hookContext.Lock()
+		hookContext.Set("stdin", reqData.Stdin)
+		hookContext.Unlock()
 	}
 
-	stdout, stderr, err := ctlcmd.Run(context, reqData.Args, uid)
+	stdout, stderr, err := ctlcmd.Run(r.Context(), hookContext, reqData.Args, uid)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


dmitry-lyfar (approved):
Looks good, there are a couple of linter issues

dmitry-lyfar (approved):

PR #1029, opened 2026-09-09
Add task-backed secret retrieval for workshopctl

tlm:
## Summary

Add task-backed secret retrieval for `workshopctl get-secret`, establishing
the path from the daemon-side command through the overlord to the response.

- Register a secret manager and a `get-secret` task handler.
- Add `secretstate.GetSecret` to validate arguments, schedule a task, and
  wait for its result.
- Pass secret values through the non-persistent state cache, removing them
  after consumption or cancellation.
- Route the command through this flow and provide specific identifier
  validation errors without echoing the supplied value.
- Start the overlord loop in the existing daemon test so its request can
  complete.

## Why

Secret retrieval needs to run through the overlord without persisting
secret values in task or change data. This establishes task scheduling,
result delivery, and request cancellation before introducing real secret
providers and workshop identity resolution.

## Current limitations

This is deliberately an incremental implementation:

- The command supplies placeholder project and workshop identity.
- The task handler returns `workshop-placeholder-secret`.
- Restart recovery and handling tasks whose original caller is no longer
  waiting are deferred.

## Validation

- Unit tests cover the secret manager, scheduling, command integration,
  and overlord registration.
- Race checks passed for the secretstate and command packages.
- The existing daemon instance-ID test passes with task execution enabled.
- Manually confirmed that `workshopctl get-secret` inside a workshop
  returns the placeholder value.

### DIFF
--- a/internal/overlord/secretstate/getsecret.go
+++ b/internal/overlord/secretstate/getsecret.go
@@ -1,0 +1,149 @@
+// Copyright (c) 2026 Canonical Ltd
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+package secretstate
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/canonical/workshop/internal/overlord/state"
+	"github.com/canonical/workshop/internal/sdk"
+	"github.com/canonical/workshop/internal/workshop"
+)
+
+// GetSecret schedules retrieval for a previously authorised secret consumer
+// and waits for its result. The caller must not hold the state lock, and the
+// state engine must be running. Results are consumed from the in-memory cache.
+// Cancellation aborts the change and discards any result.
+//
+// The following errors may be expected:
+//   - [context.Canceled]: the request was cancelled.
+//   - [context.DeadlineExceeded]: the request deadline expired.
+func GetSecret(
+	ctx context.Context,
+	st *state.State,
+	project workshop.Project,
+	ref sdk.PlugRef,
+) ([]byte, error) {
+	err := ctx.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	user, ok := ctx.Value(workshop.ContextUser).(string)
+	if !ok || user == "" {
+		return nil, errors.New("secret request has no user")
+	}
+
+	err = validateGetSecretArgs(project, ref)
+	if err != nil {
+		return nil, fmt.Errorf("validating get secret request arguments: %w", err)
+	}
+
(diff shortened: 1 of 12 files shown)


internal/overlord/secretstate/manager.go
```diff
@@ -0,0 +1,130 @@
...
+	st.Lock()
+	defer st.Unlock()
+
+	// Check under the publication lock: an abort may precede tomb cancellation.
+	switch status := task.Status(); status {
+	case state.DoingStatus:
+		st.Cache(secretResultKey(task.ID()), value)
+		return nil
+
+	case state.AbortStatus:
```
dmitry-lyfar: I think the best place to clear the cache state is the undo handler. If a task is aborted in flight, it will be waited until Done and then an undo handler is executed. See taskrunner.go:281.

So for us, once you got the secret, there is no other errors except for external cancellation / abortion. This would trigger the undo handler that should clear the cache. As opposed to handling task statuses in the handler, we should rely on the taskrunner orchestration.

internal/overlord/secretstate/manager.go
```diff
@@ -0,0 +1,130 @@
...
+) ([]byte, error) {
+	err := ctx.Err()
+	if err != nil {
+		return nil, err
+	}
+	return []byte("workshop-placeholder-secret"), nil
+}
+
+// New creates a secret manager and registers its task handlers.
+func New(runner *state.TaskRunner) SecretManager {
```
dmitry-lyfar, line 116: I suspect it would receive the interface repository as a dependency here to query interfaces?

internal/overlord/secretstate/getsecret.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, errors.New("secret request has no user")
+	}
+
+	err = validateGetSecretArgs(project, ref)
+	if err != nil {
+		return nil, fmt.Errorf("validating get secret request arguments: %w", err)
+	}
+
+	st.Lock()
+	task := st.NewTask("get-secret", "Retrieve a workshop secret")
```
dmitry-lyfar: We put an exact plug or slot name in the description when connect or disconnect an interface. We should put the full secret name in the summary here too.

internal/overlord/secretstate/getsecret.go
```diff
@@ -0,0 +1,149 @@
...
+	case <-ctx.Done():
+	}
+
+	resultKey := secretResultKey(task.ID())
+
+	st.Lock()
+	defer st.Unlock()
+	// Remove the secret on every return path so results do not outlive their
+	// request. This defer runs before the state is unlocked.
+	defer st.Cache(resultKey, nil)
```
dmitry-lyfar, line 83: NB: I suppose you considered using `AddCleanup` for this kind of change so that it could make sure the cache is clean on the next ensure pass?

tlm: I considered `AddCleanup`, but change readiness does not guarantee the waiting caller has consumed the result. A cleanup handler could run before `GetSecret` reacquires the state lock and delete the secret prematurely.

Caller-side deferred cleanup handles consumption and cancellation, while the undo handler added in the local follow-up commit removes results published after an abort. I have retained that ownership rather than adding a competing cleanup path. The follow-up commits have not been pushed yet.

dmitry-lyfar:
Looks good, the test's issue with autostart should be fixed in main, I think it's just wasn't rebased yet?

dmitry-lyfar (approved):

PR #1037, opened 2026-09-14
Resolve workshop secrets through SDK providers

tlm:
## Summary

Connect the existing get-secret task flow to an SDK-keyed provider resolver,
keeping provider selection separate from task scheduling and SDK-specific
retrieval. The overlord assembles and injects the built-in system provider.

- Validate the workshop, installed SDK and requested secret plug, then resolve
  its connected provider slot. Reject missing or ambiguous connections.
- Read typed system slot configuration through a narrow repository interface,
  defaulting omitted collections to "default" and returning independent
  attribute maps.
- Preserve a single-consumption Secret through resolution, task caching and
  retrieval. Transfer ownership on success; close abandoned cached results,
  including results published after cancellation and cleaned up by task undo.
- Consume the secret only when copying it into workshopctl stdout. Reads clear
  consumed bytes, Close clears any remainder, and fmt formatting suppresses
  accidental output. Copies share consumption state; concurrent use is not
  supported. Ordinary stdout and transport buffers are outside this control.

## Scope

This establishes the provider integration and ownership path, not real Secret
Service retrieval. The system provider still returns a hard-coded placeholder;
D-Bus integration is deferred. The command also still uses placeholder project
and workshop identities, which must be replaced before enabling real providers.

## Validation

- go test ./... passed on the rebased branch.
- git diff --check against secrets-implementation passed.
- Added focused coverage for provider selection, slot validation, identity
  scoping, ownership transfer, cancellation/undo cleanup and command output.


internal/overlord/secretstate/manager.go
```diff
@@ -96,31 +124,95 @@ func (m SecretManager) Ensure() error {
...
+	if err != nil {
+		return secrets.Secret{}, fmt.Errorf("resolving secret: %w", err)
+	}
+	return secret, nil
 }
 
 // New creates a secret manager and registers its task handlers.
-func New(runner *state.TaskRunner) SecretManager {
-	manager := SecretManager{}
+func New(
```
dmitry-lyfar: What is the reason to be inconsistent with how other state managers do it by accepting the task runner and / or a backend interface?

internal/sdk/system/secretprovider.go
```diff
@@ -0,0 +1,56 @@
...
+
+// SecretProvider returns a placeholder for system SDK secret slots until
+// retrieval from the user's Secret Service over D-Bus is implemented.
+type SecretProvider struct {
+	slots SecretSlotLookup
+}
+
+// NewSecretProvider creates a system SDK secret provider using slots to read
+// typed slot configuration.
+func NewSecretProvider(slots SecretSlotLookup) SecretProvider {
```
dmitry-lyfar: The provider should be decoupled from the system SDK. This providers is a secret service provider. It implements the interface from secrets. Similarly, to how lxd backend implements the workshop interface.

internal/sdk/system/secretslot.go
```diff
@@ -0,0 +1,146 @@
...
+type SecretSlotLookup interface {
+	// Lookup returns validated, independently owned configuration for the
+	// referenced slot, or an error if the slot is missing or invalid.
+	// Implementations must release any
+	// repository or state locks before returning.
+	Lookup(context.Context, sdk.SlotRef) (SecretSlotConfig, error)
+}
+
+// SlotRepository looks up registered SDK slots.
+type SlotRepository interface {
```
dmitry-lyfar: These interfaces are agnostic of the system SDK, they are designed to work with any SDK's slot, they are rather related to the secrets package. So, it should also be decoupled from the system SDK.

PR #1040, opened 2026-09-15
Add initial system SDK Secret Service client

tlm:
> [!WARNING]
> DO NOT MERGE. This draft captures the initial implementation and a verified cross-user D-Bus blocker that must be resolved first.

# Description

Adds an initial freedesktop Secret Service client for the system SDK. The implementation resolves default and named collections, rejects ambiguous item matches, opens a plain Secret Service session, and returns secret material through the existing Workshop secret type.

The implementation is intentionally isolated and is not wired into the system SDK secret provider.

## Known blocker: user session bus authentication

workshopd runs as root, while the Secret Service is available on the requesting users session bus. The current connector opens /run/user/<uid>/bus as root and supplies the target UID to D-Bus EXTERNAL authentication.

This does not work on the tested host because EXTERNAL authentication does not change the Unix peer credentials of the connecting process.

Observed with UID 1001:

- A probe running as UID 1001 authenticated successfully and called Secret Service ReadAlias.
- The same probe running through sudo while claiming UID 1001 failed during connection with a broken pipe.
- sudo busctl connecting to the same user bus failed because the transport was disconnected.

The hackathon branch tlm/workshop:secrets-hackathon worked around this by temporarily changing the UID of a locked OS thread before connecting. That demonstrates feasibility, but raw credential switching inside the multithreaded daemon has restoration and isolation risks.

The proposed production direction is to launch a short-lived helper process with the target UID and GID. That helper would perform the complete D-Bus operation and return the result over private pipes. The daemon must not temporarily change its own credentials.

## Additional follow-up before merge

- Prevent uncleared duplicate secret buffers created by godbus Call.Store.
- Make the initial Unix socket dial context-aware.
- Map Secret Service locked errors that race with the collection lock check.
- Validate the plain session response envelope.
- Add tests for named collections and all stable error paths.
- Complete function documentation and ordering cleanup.

## Validation

- go test ./internal/sdk/system/...
- go vet ./internal/sdk/system/...

# Self-review quick check

## Docs

* [x] I confirm the PR has no implications for documentation because the implementation is internal and not wired into user-facing behaviour.


cmd/workshopsyssecret/doc.go
```diff
@@ -0,0 +1,44 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+// Command workshopsyssecret retrieves one secret from the host Secret Service.
```
dmitry-lyfar: `secret-service-tool` or `secret-service-helper`?

tlm: secret-service-tool I think. Should we prefix it with workshop? I know it's long.

PR #1044, opened 2026-09-18
Auto-connect the mount interface for VMs

jonathan-conder:
# Description

# Self-review quick check

 ```
 // one way to handle errors
 if err := f(); err != nil {
    ...
 }
 
 val, err := f()
 if err != nil {
    ...
 }
 ...
 ```
 * [ ] Follow the [style guide](https://github.com/canonical/workshop/tree/main/docs/coding-style-guide.md#error-handling) for new error messages.

## Docs

Procedure:

Content:

---

Or:


internal/overlord/ifacestate/handlers.go
```diff
@@ -220,7 +221,7 @@ func workshopConns(wp *workshop.Workshop) []interfaces.ConnRef {
 	return conns
 }
 
-func (m *InterfaceManager) connectAuto(task *state.Task, wp *workshop.Workshop, info *sdk.Info, preserved map[string]schema.PreservedConn) error {
+func (m *InterfaceManager) connectAuto(ctx context.Context, task *state.Task, wp *workshop.Workshop, info *sdk.Info, preserved map[string]schema.PreservedConn) error {
```
dmitry-lyfar, line 224: I'd suggest checking confinement validity at the install stage, i.e. earlier. An unsupported plug or slot should not be added to the repository at all.

Concretely, I'm thinking it should be done in `BeforePrepareSlot` when `registerSdk` is getting called. We hard fail if there is a bad interface found at the moment, but the original behavior reported it as a warning. I haven't explored this much yet, but it would likely require extending the `SdkInfo` struct with the workshop's runtime kind that can be used by interfaces to check plugs and slots for compatibility.

jonathan-conder: I'll see how possible it is. I didn't want to leak too much about the workshop into sdk.Info because it feels out of place. But maybe we can avoid it somehow.

This will (I think) hide those plugs and slots from the CLI outputs, wouldn't that be confusing to SDK authors?

dmitry-lyfar: We can either choose to preserve the behavior and abort the change OR raise it as a warning. It's potentially possible to make it a permanent note in the SDK status too, but I would not go that way just yet.

Yes, `SdkInfo` does not have anything but the workshop's name now. It's a bit inconsistent compared to, say, `SdkInfo` vs. `PlugInfo`. The latter contains a full reference to the former. Ideally, `SdkInfo` containers a pointer to its workshop, but I don't think we want that much coupling and it's likely going to be a bigger change. So, having the awareness of where it is being installed seems to be a reasonable way.

dmitry-lyfar: Hmm, it looks like that we should check the plugs and slots provided in the workshop definition too. All that parsing and setup happens in `Workshop.SdkInfo` at the moment.

jonathan-conder: > We can either choose to preserve the behavior and abort the change OR raise it as a warning. It's potentially possible to make it a permanent note in the SDK status too, but I would not go that way just yet.

Maybe I don't understand this. Do you mean we abort the change if an SDK has an unsupported plug? I don't think that's a real option, e.g. it will block the `opencode` SDK from VMs entirely. I also don't think it's worth a warning, it will be annoying.

Maybe it's not a big deal to hide it from `workshop connections` though, as long as we do strict validation so typos don't also make the plug vanish. It makes sense that "this workshop does not have this plug," maybe we should also have an `sdk` subcommand to list all of them for a given SDK.

dmitry-lyfar: I was wrong, the original behavior just listed the bad interfaces in the task's log and skipped them.

jonathan-conder: Ah ok, yes logging makes sense

